PR #132, opened 2012-07-11
Mix RSpec with Test::Unit

mike-burns:
An edge case occurs when mixing RSpec and Test::Unit tests and also
loading both the 'rspec-rails' gem and 'shoulda-matchers' gem from the
same Gemfile group, namely [:test, :development] . Work around this by
always inserting the shoulda matchers into Test::Unit, regardless of
whether RSpec is loaded.

### DIFF
--- a/features/rails_integration.feature
+++ b/features/rails_integration.feature
@@ -86,3 +86,31 @@ Feature: integrate with Rails
     Then the output should contain "2 examples, 0 failures"
     And the output should contain "should require name to be set"
     And the output should contain "should assign @example"
+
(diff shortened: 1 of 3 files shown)


collimarco:
> An edge case occurs when mixing RSpec and Test::Unit

What do you mean by "mixing"? Can I uninstall/disable Test::Unit if I don't use it?

mcmire:
What it means is that if, for some reason, you're using both RSpec and Test::Unit (perhaps you're switching over to RSpec).

PR #141, opened 2012-08-14
Allow blank and nil on validate format

gylaz:
- Merge pull request #127
- Clean up tests

### DIFF
--- a/spec/shoulda/active_model/validate_format_of_matcher_spec.rb
+++ b/spec/shoulda/active_model/validate_format_of_matcher_spec.rb
@@ -9,35 +9,35 @@ describe Shoulda::Matchers::ActiveModel::ValidateFormatOfMatcher do
       @model = Example.new
     end
 
-    it "should be valid" do
+    it "is valid" do
       @model.should validate_format_of(:attr).with('12345')
     end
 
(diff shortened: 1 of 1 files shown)


aried3r:
Sorry for digging this up, but I haven't seen this merged elsewhere or inside the code. Any plans to do so for 3.1.0?

maurogeorge:
@aried3r the matcher used today to test the `validates_format_of` is the the [allow_value](http://matchers.shoulda.io/docs/v2.8.0/Shoulda/Matchers/ActiveModel.html#allow_value-instance_method). If you want to check that your model accept blank and nil you can do something like:

```ruby
class Company < ActiveRecord::Base
  validates :site_url, format: { with: URI.regexp }, allow_blank: true
end

# RSpec
it do
  is_expected.to allow_value(nil, '', '  ', 'http://foo.com', 'http://bar.com/baz')
    .for(:site_url)
end

it do
  is_expected.to_not allow_value('not-url', 'www')
    .for(:site_url)
end
```

aried3r:
Thank you, I'll do that. Just wanted to be sure there are no helpers, since `blank` can be so much, but your example with `nil, '', ' '` should be easy to read for others, too.

PR #377, opened 2013-11-01
Remove #for from validation matchers

mcmire:
Certain classes for validation matchers support a #for method to set the
attribute that the matcher operates on. This is bad API design, because
it means that it is possible for a developer to accidentally (or
purposefully) never call #for and yet still use the matcher (which would
result in the matcher blowing up). Because of this, the attribute should
be required in the initializer for the matcher.

Related to this, #allow_value also supports an optional attribute, which
should be required. Simply moving `.for(...)` to the first argument to
\#allow_value wouldn't read very well, so we rename it to
\#allow_value_for.

---

Obviously, this is a **backward-incompatible change**, but I wanted to get you guys' opinion on this.


mcmire:
I don't know what the ramifications of this change would be and I don't know if it's worth it to make it right now; it'd be better to fail the matcher if `for` is not provided, and make a decision on this later.

In any case, it's too late to make this change since we can't go through a deprecation cycle. Closing for now.

PR #387, opened 2013-11-26
  adds example to readme to elaborate on #with_message 

meesterdude:
small change; but was getting stuck and had to go poking about the implementation to figure out why my specs were failing and how to fix.


mcmire:
This is properly fixed in a453dccfd923b651197367283a95d338e7f01604. (There's also docs for the `:against` option I added a while back, too.)

PR #411, opened 2014-01-06
Reject associations when class is not defined

chulkilee:
In most cases associated classes will be loaded together, so it would be great if we can catch when wrong class name is used or class is missing.


chulkilee:
commit for this PR - f835cf43

PR #534, opened 2014-05-28
Add eager_load matcher

jferris:
Makes it easier to verify that a piece of code does not run N+1 queries.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
   columns allow only specific values, which is not true. If you get an error
   from `validate_uniqueness_of`, your best bet continues to be creating a record
   manually and calling `validate_uniqueness_of` on that instead.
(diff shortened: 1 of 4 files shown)


mcmire:
Revisiting this slightly. Is this something we can add to the [db-query-matchers](https://github.com/brigade/db-query-matchers) gem?

maurogeorge:
@mcmire I did not know the `db-query-matchers`, but I think this gem solves the same problem we are trying to solve here with the eager_load matcher.
If i am not missing nothing the eager_load matcher actually count the number of queries, the same behavior of the db-query-matchers.

If we want to add this same behavior, maybe we can try to expose a diffent name since we can reduce the number of queries on AR [with includes, preload and eager_load](http://blog.arkency.com/2013/12/rails4-preloading/) for example.

mcmire:
I'm going to close this feature request, but I've made a note of it.

PR #556, opened 2014-07-23
Add support for HBTM `:join_table` option

jacobsimeon:
In reference to Issue #547.

After renaming the method and adding tests for both the acceptance and rejection cases, I realized that the `AssocationMatcher` was already aware of any custom `:join_table` option that might have been passed to the association declaration.

So all I've done is add this method to `AssociationMatcher`:
``` ruby
def join_table(join_table_name)
  self
end
```

This was enough to make the tests pass, although it kinda feels like lying to the user. I see a few options:

1. Don't bother with adding and explicit `join_table` method to `AssociationMatcher` since it could potentially mislead users and the join table's existence will always be tested correctly.
2. Keep the `join_table` method since it adds parity with ActiveRecord's API and gives a signal to users (and their coworkers) that a custom join table is being used for the association.
3. Go forward with more explicit support for the `:join_table` option, similar to what is already being done for the `:foreign_key` option.

What do you think?

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -747,7 +747,7 @@ module Shoulda
       # @private
       class AssociationMatcher
         delegate :reflection, :model_class, :associated_class, :through?,
-          :join_table, :polymorphic?, to: :reflector
+          :join_table_name, :polymorphic?, to: :reflector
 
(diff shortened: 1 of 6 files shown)


mcmire:
I took care of fixing the errors and merged this as af85ac1f0e5b9a267efd4c05b30a73fe2f44d84e. Thanks for the PR, cheers!

PR #629, opened 2014-12-24
set next value using a specific method. small refactoring

lucianosousa:
uncle bob approves


mcmire:
Thanks! Merged in 9b252cca9bb28f4205e55f2baf753cb0c8fdcddf

PR #631, opened 2014-12-28
add true option for dependent matcher and test all options

### LINKED ISSUE #393: Have_many matcher should support `dependent: :restrict_with_exception`
it { should have_many(:recipients).dependent(:restrict_with_exception) }

raises:
The :dependent option must be one of [:destroy, :delete], but is :restrict_with_exception

barelyknown:
Added test to ensure that all valid `:dependent` options are permitted to fix #393. Also added a `true` option that accepts any `:dependent` option. This is useful when you're testing that all `has_many` or `has_one` relationships have a `:dependent` option set.

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matchers/dependent_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matchers/dependent_matcher.rb
@@ -34,14 +34,14 @@ module Shoulda
             @option_verifier ||= OptionVerifier.new(subject)
           end
 
-          def correct_for?(dependent=dependent)
+          def correct_for?(dependent)
             case dependent
             when true, false then option_verifier.correct_for_boolean?(:dependent, dependent)
             else option_verifier.correct_for_string?(:dependent, dependent)
             end
           end
 
-          def missing_option_for(name=name, dependent=dependent)
+          def missing_option_for(name, dependent)
             "#{name} should have #{dependent == true ? "a" : dependent} dependency"
           end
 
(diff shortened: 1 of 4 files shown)


spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
```diff
@@ -1,5 +1,16 @@
 require 'unit_spec_helper'
 
+def each_dependent_option(rails_version)
```
barelyknown: Should be all set.

lib/shoulda/matchers/active_record/association_matchers/dependent_matcher.rb
```diff
@@ -34,6 +33,18 @@ def matches?(subject)
           def option_verifier
             @option_verifier ||= OptionVerifier.new(subject)
           end
+
+          def correct_for?(dependent=dependent)
```
barelyknown: The original reason was that there is a `dependent` method and therefore the argument wouldn't be needed, but I didn't like how the code read if the argument wasn't passed. But, I just made the change that you suggested.

spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
```diff
@@ -1117,4 +1137,13 @@ def define_association_with_order(model, macro, name, order, other_options={})
     args << options
     model.__send__(macro, name, *args)
   end
+
+  def dependent_options(rails_version)
```
mcmire: What do you think about this method simply referring to `Rails.version` directly instead of taking it as an argument?

barelyknown: Yeah, that's fine. Made that change.

mcmire:
Thank you! Merged in948757b3f8af079869f2a94b628ee17db756b86d.

PR #632, opened 2014-12-28
testing against ruby v2.2

lucianosousa:


mcmire:
I think that's fine actually. That just looks like a warning, I don't actually think it's an error.

Travis is failing on a couple of jobs when it runs the build for some reason. I'll investigate this.

lucianosousa:
this is not on travis. it's in my pc. looks this line: https://gist.github.com/lucianosousa/107de6fe9d476a1e9538#file-gistfile1-txt-L639

mcmire:
If the tests aren't failing, though, then it's not an issue. I'm thinking that you're seeing a warning, not an error. In other words, you're likely (advertently or inadvertently) running RSpec with the --warnings flag on, see if you can turn this off (it might be in your .rspec or otherwise).

lucianosousa:
@mcmire the specs did not start because this mixed dependencies. it's a truly fail.

mcmire:
@lucianosousa Yeah, that's my bad, I needed to regenerate the gemfiles for that Appraisal. Can you try rebasing against master?

lucianosousa:
@mcmire done

mcmire:
@lucianosousa I think that Rails 4.0.x is technically still supposed to be compatible with Ruby 2.2. At least, it's not end-of-life yet.

mcmire:
Also can you run `bundle update sqlite3` against the 3.2 and 4.1 appraisals as well?

mcmire:
Oh I just tried to add those appraisals and got the error you were talking about. Okay. I'll look into that. Perhaps 4.0 and 4.1 aren't compatible with Ruby 2.2 just yet after all...

mcmire:
This seems to be an rspec-rails issue and was fixed very recently: https://github.com/rspec/rspec-rails/commit/999ebb7c5c178c462f944d792626d27dc48147b0. There isn't a new version out yet, but we could conceivably just point to master for right now since this is just a development dependency...

lucianosousa:
hum. didnt see this before. just wait for the new release then.

go ahead here! o/

mcmire:
Looks like RSpec 3.2 was just released, so this *should* work now once I bump the dependency.

lucianosousa:
:+1:

Looks like OK, no?

mcmire:
Yeah, I'm working on it, I'll merge this in.

mcmire:
@lucianosousa I merged your branch in #655. Thanks for the help!

PR #641, opened 2015-01-15
Fix enum attribute name verification

guilhermeFranco:
Hello guys !

I noticed today that we're making the enum definition verification based on the plural name of the attribute. The following test pass currently:

```ruby
# model
class Foo
  enum status: [:active, :inactive]
end

# spec
describe Foo do
  it { is_expected.to define_enum_for(:statuses) }
end
```

This is kind of awkward (the enum was defined for the status attribute not statuses), but still makes sense idiomatically speaking (we're indeed defining a set of statuses). This is happening because of the following verification of our matcher:

```ruby
# define_enum_for_matcher.rb
def enum_method
  attribute_name.to_s.pluralize
end

def enum_defined?
  model.class.respond_to?(enum_method)
end
```

The problem with this approach is that we're not verifying if the enum was actually defined. The `enum_defined?` method is only verifying if the class of the object under test responds to a method that has the plural name of the enum attribute present on the matcher. So, we could screw things up a little bit by doing the following (I tested it, and it passes !):

```ruby
# model
class Foo
  def self.statuses
  end
end

# Those specs passes even if no enum was defined !
describe Foo do
  it { is_expected.to define_enum_for(:status) }
  it { is_expected.to define_enum_for(:statuses) }
end
```

So, to sum up. I wrote a failing spec about the issue and fixed it, making the `enum_defined?` verification based on the actual enums defined for the object under test. I also deleted a method that is not needed anymore.

I hope it's all ok :)

### DIFF
--- a/lib/shoulda/matchers/active_record/define_enum_for_matcher.rb
+++ b/lib/shoulda/matchers/active_record/define_enum_for_matcher.rb
@@ -99,16 +99,12 @@ module Shoulda
           hashify(options[:expected_enum_values]).with_indifferent_access
         end
 
-        def enum_method
-          attribute_name.to_s.pluralize
-        end
-
         def actual_enum_values
-          model.class.send(enum_method)
+          model.class.send(attribute_name.to_s.pluralize)
         end
 
         def enum_defined?
-          model.class.respond_to?(enum_method)
+          model.defined_enums.keys.include?(attribute_name.to_s)
         end
 
         def enum_values_match?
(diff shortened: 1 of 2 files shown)


americodls:
:+1:

guilhermeFranco:
Just removed the trailing whitespaces. Sorry about that guys :disappointed:

lucianosousa:
:+1:

mcmire:
Yup, good call on this. I'll queue this for 2.8.1.

mcmire:
Sorry for the delay on this. This is merged in 714129b82f898f9f1644f43092ab21cb6f5fc2d0. Thanks!

PR #650, opened 2015-02-05
Remove small caps on section header

philtr:
Unfortunately, Brandon Grotesque doesn't have proper small caps, so it makes the large caps look much heavier than the small caps. I personally think that it looks much better without them.


courtsimas:
:+1:

mcmire:
@rachelcope What do you think about this?

mcmire:
@philtr Thanks for this! Merged.

PR #651, opened 2015-02-05
Remove debugger, byebug, and web-console from 4.2

mcmire:
debugger and byebug cannot be present in Appraisals because they only
work on specific Ruby versions, and we test against a range of Ruby
versions. Hence, they can't be present in the Rails application that
gets generated in acceptance tests, either.

Also, we don't really need web-console to be there, it's just an extra
dependency.


Appraisals
```diff
@@ -120,8 +120,6 @@ if Gem::Requirement.new('> 1.9.2').satisfied_by?(ruby_version)
     gem 'jbuilder', '~> 2.0'
     gem 'sdoc', '~> 0.4.0'
     gem 'bcrypt', '~> 3.1.7'
-    gem 'byebug'
```
seuros: we could add `platforms: [:mri_20,:mri_21]`

mcmire: Oops I missed this, sorry... 1) I didn't even know you could do this and 2) I'm not sure how I feel about additional conditionals in Gemfiles. I'd rather just remove those dependencies because I don't think they're going to make an impact on how shoulda-matchers works. But I'll keep this in mind...

PR #652, opened 2015-02-05
Test against RSpec 3 from now on

mcmire:
* Ruby 2.2 removed Minitest and Test::Unit from the standard library
  [[1]], [[2]]
* rspec-rails requires Test::Unit for Rails versions prior to 4.1 (which
  switched to Minitest)
* This doesn't work now, because we don't have Test::Unit present in the
  gem bundle
* RSpec 3.2.0 fixes this issue [[3]]

I don't really see this as a huge concern, since we were testing against
RSpec 2.99 for Rails < 4, and that has most of the changes that RSpec 3
has.

[1]: https://github.com/ruby/ruby/commit/f8c6a5dc02439bb99c8403602288c2937b9356cc
[2]: https://github.com/ruby/ruby/commit/96f552670d045c3e76c02929753f6934da51974f
[3]: https://github.com/rspec/rspec-rails/commit/999ebb7c5c178c462f944d792626d27dc48147b0#diff-08d960c572ac094640dd183fa9641393R13


lucianosousa:
:+1:

PR #653, opened 2015-02-05
Inclusion validation message with %{value}

aliismayilov:
After updating from `2.7.0` to `2.8.0` I started to have fails with the following spec:

``` ruby
class Image < ActiveRecord::Base
  validates :content_type, presence: true,
    inclusion: {
      in: %w(image/gif image/jpg image/jpeg image/pjpeg image/png image/svg+xml image/vnd.djvu),
      message: "%{value} is not a valid image mime type"
    }
end
```

``` ruby
it do
    is_expected.to validate_inclusion_of(:content_type).
      in_array(%w(image/gif image/jpg image/jpeg image/pjpeg image/png image/svg+xml image/vnd.djvu)).
      with_message('%{value} is not a valid image mime type')
end
```

``` sh
Failures:

  1) Image validations should ensure inclusion of content_type in ["image/gif", "image/jpg", "image/jpeg", "image/pjpeg", "image/png", "image/svg+xml", "image/vnd.djvu"]
     Failure/Error: is_expected.to validate_inclusion_of(:content_type).
       ["image/gif", "image/jpg", "image/jpeg", "image/pjpeg", "image/png", "image/svg+xml", "image/vnd.djvu"] doesn't match array in validation
     # ./spec/models/image_spec.rb:10:in `block (3 levels) in <top (required)>'
```

I added failing spec for `validate_inclusion_of` matcher to show the bug. I'd like to have some tips how to fix it and complete this pull request.

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_exclusion_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_exclusion_of_matcher_spec.rb
@@ -52,6 +52,11 @@ describe Shoulda::Matchers::ActiveModel::ValidateExclusionOfMatcher, type: :mode
       expect(validating_exclusion(in: 2..4, message: 'not good')).
         to validate_exclusion_of(:attr).in_range(2..4).with_message(/not good/)
     end
+
+    it 'accepts ensuring the correct range with %{value}' do
+      expect(validating_exclusion(in: 2..4, message: '%{value} is not good')).
+        to validate_exclusion_of(:attr).in_range(2..4).with_message(/ is not good/)
+    end
   end
 
   context 'an attribute with custom range validations' do
--- a/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
@@ -212,6 +212,14 @@ describe Shoulda::Matchers::ActiveModel::ValidateInclusionOfMatcher, type: :mode
         end
       end
 
+      it 'matches when validation uses given message with %{value}' do
+        builder = build_object_allowing(valid_values, message: '%{value} is not included')
+
+        expect_to_match_on_values(builder, valid_values) do |matcher|
+          matcher.with_message('%{value} is not included')
+        end
+      end
+
       it 'does not match when validation uses the default message instead of given message' do
         builder = build_object_allowing(valid_values)
 


spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```diff
@@ -212,6 +212,14 @@ def add_outside_value_to(values)
         end
       end
 
+      it 'matches when validation uses given message with %{value}' do
```
mcmire: "...with an interpolated variable", perhaps?

aliismayilov: Where to get the variable?

mcmire: I guess what I mean is, maybe we shouldn't be so specific here in the name of the test. It's not so much that `%{value}` is being used in the message. There's a standard list of variables you can use in your custom message that when ActiveRecord populates the errors hash for your record it will replace (`value`, `record`, `attribute` ... I'm probably missing one more). So it sounds like the bug is, if you use any one of these, they don't get replaced correctly. So I'm suggesting that we name the test to accurately describe the bug, that's all.

spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```diff
@@ -212,6 +212,14 @@ def add_outside_value_to(values)
...
+        builder = build_object_allowing(valid_values, message: '%{value} is not included')
+
+        expect_to_match_on_values(builder, valid_values) do |matcher|
+          matcher.with_message('%{value} is not included')
```
mcmire: I *think* this should work...

spec/unit/shoulda/matchers/active_model/validate_exclusion_of_matcher_spec.rb
```diff
@@ -52,6 +52,11 @@ def matchers
       expect(validating_exclusion(in: 2..4, message: 'not good')).
         to validate_exclusion_of(:attr).in_range(2..4).with_message(/not good/)
     end
+
+    it 'accepts ensuring the correct range with %{value}' do
+      expect(validating_exclusion(in: 2..4, message: '%{value} is not good')).
+        to validate_exclusion_of(:attr).in_range(2..4).with_message(/ is not good/)
```
mcmire: I think you can use `.with_message('%{value} is not good')` here.

mcmire:
So here's a quick rundown of what the matcher is supposed to do in this case:

1. Set the attribute to a value that's *not* in the given array of valid values
1. Call `#valid?` on the record
1. Read the error messages off of the record
1. Run the given message through i18n, expanding it if necessary
1. Assert that the record has an error message under the attribute that is equal to the expanded version of the given message

I haven't looked at the code, but I'm guessing it's the fourth step above that's not working right here. You see, when it comes to comparing error messages, we have to simulate what ActiveRecord is doing. If you say:

``` ruby
is_expected_to.validate_inclusion_of(:attr).with_message(:invalid)
```

we have to take that `:invalid` symbol and run it through I18n, because when we read the error messages off of the record, they won't be symbols, they'll be fully expanded strings (so, "is not valid" going off the example above).

If we no longer perform this step correctly that would explain why you're now getting a failure.

mcmire:
The inclusion matcher uses the `allow_value` matcher internally to check that the record allows or does not allow certain values. This is the matcher that performs the basic logic I described above. Here is the method in particular that looks for an error message on the record: https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/allow_value_matcher.rb#L280. The `expected_message` method in particular will use the message that you passed to `with_message` and conditionally transform it. If it is a Symbol, it is run through i18n, otherwise it is left alone.

I did refactor the code that performs the logic above between 2.7.0 and 2.8.0, but I didn't touch this method in particular. So I am a little confused as to why your test has all of a sudden stopped working.

mcmire:
Hey, just updating this PR -- I updated your tests and when I ran them they passed. I guess they were missing from the test suite, so I merged it in as 5d9f8da8c6611c591813a57417edff4bcaf9bc24. But I'm still not sure why you ran into the issue you mentioned. I'm going to close this since it's been so long and we haven't had any other reports of this but if you continue to see this then feel free to reopen it.

PR #662, opened 2015-02-12
Uniqueness: Support UUID columns that end in "f"

### LINKED ISSUE #402: validate_uniqueness_of matcher incorrectly calls `#next` on UUIDs, causing spec failures
I have a Rails 4/Postgres app using UUID primary keys, with a uniqueness validation on a column which is scoped to a UUID foreign key. My specs fail occasionally because shoulda-matchers is trying to insert an invalid UUID into the database.

This happens when a UUID used to test the scope of the validation happens to end in "f". The `validate_after_scope_change?` method calls `#next` on the UUID, which makes it end in "g", which is obviously invalid hex and an invalid UUID.

[See this code in ValidateUniquenessOfMatcher](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb#L157-L171):

```ruby
previous_value = all_records.map(&scope).max

# Assume the scope is a foreign key if the field is nil
previous_value ||= correct_type_for_column(@subject.class.columns_hash[scope.to_s])

next_value =
  if previous_value.respond_to?(:next)
    previous_value.next
  else
    previous_value.to_s.next
  end

@subject.send("#{scope}=", next_value)
```

The fix for the bug should be relatively straightforward: when generating the next value, first check if the column type is UUID; if so, generate a new UUID instead of calling `#next`.

However, I cannot put together a failing test, because none of the existing tests are actually using UUIDs. The test suite uses sqlite, which does not have a ":uuid" column type. This breaks the [correct_type_for_column method](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb#L187-L197):

```ruby
def correct_type_for_column(column)
  if column.type == :string
    '0'
  elsif column.type == :datetime
    DateTime.now
  elsif column.type == :uuid
    SecureRandom.uuid
  else
    0
  end
end
```

The column type for the :uuid scopes is nil, so the case statement falls through and the UUID-related tests are actually using integers for the scope values. This means that when I try to validate the UUID for these tests, it always fails.

I can put together a pull request that fixes the bug, but I don't know how to go about adding a test for the problem. What should I do?

/cc @yabawock, who implemented the existing UUID support (thanks!)

mcmire:
Fixes #402 and #587.

---

This commit fixes the uniqueness matcher so that when qualified with
`scoped_to` and one of the scopes is a UUID column that happens to end
in an "f", the matcher is able to generate a proper next value during
testing and doesn't error.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,16 @@
   * `set_session['key'].to(nil)` will no longer pass when the key in question
     has not been set yet.
 
+### Bug fixes
+
+* So far the tests for the gem have been running against only SQLite. Now they
+  run against PostgreSQL, too. As a result we were able to fix some
+  Postgres-related bugs:
+
+  * Fix `validate_uniqueness_of` + `scoped_to` so that when one of the scope
+    attributes is a UUID column that ends in an "f", the matcher is able to
+    generate a proper "next" value without erroring.
+
 # 2.8.0
 
 ### Deprecations
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -393,11 +393,14 @@ module Shoulda
         end
 
         def correct_type_for_column(column)
-          if column.type == :string
+          column = column_for(scope)
+
+          case column.type
+          when :string
             '0'
-          elsif column.type == :datetime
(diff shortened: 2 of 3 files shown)


damncabbage:
:tada: Thank you. :-)

PR #669, opened 2015-02-20
Cache Bundler on Travis

seanpdoyle:
With Container based CI, public projects can cache the results of
`bundle install`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 sudo: false
+cache: bundler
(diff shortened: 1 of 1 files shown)


mcmire:
Okay, sounds good to me. We'll see how much this speeds up the build.

seanpdoyle:
screen shot 2015-02-22 at 10 00 41 am


Looks like the bottleneck is the suite itself, not dependencies. Oh well, a few minute bump is still nice

PR #676, opened 2015-03-01
finishing drop support to ruby 1.9.x

lucianosousa:
finishing drop support to ruby 1.9.x


mcmire:
Ah, good call ;

mcmire:
Merged as d811bec14ae9362749e4dfb3f4d17917ef9c8429. Thanks!

PR #677, opened 2015-03-01
validate_after_scope_change? use compact before  retrieving max

pandamouse:
Otherwise produces error "ArgumentError: comparison of NilClass with String failed" if the resultant contains a combinations of strings and nils.


mcmire:
Thanks for the PR! I merged this as b3113117bd9bc445f297e172abebdaebb791147d.

pandamouse:
Cheers!

PR #687, opened 2015-03-26
Action filters links

ignat-zakrevsky:
Hi

It seems like links to filter methods were broken, let's fix it!


mcmire:
Thanks! Merged as 762cebdbbcb53f11253d516f8b65c6f7cebfe04c.

PR #692, opened 2015-03-30
Create the allow_values matcher

### LINKED ISSUE #562: Array version of `allow_value`
Sometimes you need to see if some values are allowed for a defined field and using the current features you would have to use many `allow_value().for()` constructs. It would be nice to have a array version, like `allow_values('foo', 'bar').for()`.

maurogeorge:
Closes #562

### DIFF
--- a/lib/shoulda/matchers/active_model/allow_value_matcher.rb
+++ b/lib/shoulda/matchers/active_model/allow_value_matcher.rb
@@ -194,6 +194,7 @@ module Shoulda
           AllowValueMatcher.new(*values)
         end
       end
+      alias_method :allow_values, :allow_value
 
       # @private
       class AllowValueMatcher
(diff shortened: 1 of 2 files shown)


mcmire:
Thanks for the PR -- I'm actually already working on this. I'd like to change `allow_value` so it only accepts one attribute and not multiple -- `allow_values` will be responsible for testing multiple values, if that's what you'd like (and that way we can move the `match?` / `does_not_match?` logic there, instead).

maurogeorge:
@mcmire cool, your implementation make more sense. I will close this PR since it add only a alias, we can continue using the `allow_value` to multiple values until the release of the 3.0.

mcmire:
I changed my mind about creating a separate `allow_values` matcher -- after playing around it a bit, I think it's an unnecessary backward incompatibility and it's not worth having to write separate tests for. 

So I went ahead and merged this: ae77c7d736664876ddea726beee92c7888090783. I'll probably end up switching it around so that `allow_values` is the "real"  matcher and `allow_value` is the alias (since it is designed to accept multiple values after all).

maurogeorge:
:+1: makes sense to me.

PR #693, opened 2015-03-30
Fix route matcher accepts a format as a symbol or string

### LINKED ISSUE #454: Route with symbol for default format cannot match
In Rails 4, the default `format` route parameter may have been set to a symbol, but because shoulda stringifies the expected parameters given, it is impossible to match a route with a default `format` set to a symbol.

```ruby
namespace :api, defaults: {format: :json} do
  resources :user
end
```

Try to match it and you will get an unhelpful error, like:

```
       <{"format"=>"json",
        "action"=>"show",
        "controller"=>"api/users"}> expected but was
       <{"format"=>:json,
        "action"=>"show",
        "controller"=>"api/users"}>
```

Even crazier, when you switch from back and forth from a symbol to a string in your tests, the error message alternates between the first and the second hash containing the symbol value.

A temporary fix is to use the RSpec `route_to` matcher, which does not stringify the values.

maurogeorge:
Closes #454

### DIFF
--- a/lib/shoulda/matchers/action_controller/route_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/route_matcher.rb
@@ -79,6 +79,10 @@ module Shoulda
       #
       #     route(:get, '/posts/1').to('posts#show', id: 1)
       #
+      # You may also specify special parameters such as `:format`:
+      #
+      #     route(:get, '/posts').to('posts#index', format: :json)
+      #
       # @return [RouteMatcher]
       #
       def route(method, path)
--- a/lib/shoulda/matchers/action_controller/route_params.rb
+++ b/lib/shoulda/matchers/action_controller/route_params.rb
@@ -3,6 +3,8 @@ module Shoulda
     module ActionController
       # @private
       class RouteParams
+        PARAMS_TO_SYMBOLIZE = %i{ format }
+
         def initialize(args)
           @args = args
         end
@@ -26,17 +28,24 @@ module Shoulda
         def extract_params_from_string
           controller, action = args[0].split('#')
           params = (args[1] || {}).merge(controller: controller, action: action)
-          stringify_values(params)
+          normalize_values(params)
         end
 
         def stringify_params
-          stringify_values(args[0])
+          normalize_values(args[0])
         end
 
-        def stringify_values(hash)
-          hash.inject({}) do |hash_copy, (key, value)|
-            hash_copy[key] = stringify(value)
-            hash_copy
+        def normalize_values(hash)
+          hash.each_with_object({}) do |(key, value), hash_copy|
+            hash_copy[key] = symbolize_or_stringify(key, value)
+          end
+        end
+
+        def symbolize_or_stringify(key, value)
+          if key.in?(PARAMS_TO_SYMBOLIZE)
+            value.to_sym
+          else
+            stringify(value)
           end
         end
 
--- a/spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
@@ -48,6 +48,26 @@ describe 'Shoulda::Matchers::ActionController::RouteMatcher', type: :controller 
             to(action: 'show', some: 'other', params: 'here')
         end
       end
+
+      context 'when route has a default format' do
+        it 'accepts' do
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to(action: 'create', format: 'json')
+        end
+
+        it 'accepts when format is specified as a symbol' do
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to(action: 'create', format: :json)
+        end
+
+        it 'rejects when format is unspecified' do
+          expect(controller_with_defined_routes).
+            not_to route(:post, "/#{controller_path}").
+            to(action: 'create')
+        end
+      end
     end
 
     context 'when controller and action are specified as a joined string' do
@@ -64,6 +84,20 @@ describe 'Shoulda::Matchers::ActionController::RouteMatcher', type: :controller 
             to("#{controller_path}#show", id: 1)
         end
       end
+
+      context 'when route has the format' do
+        it 'accepts' do
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to("#{controller_path}#create", format: 'json')
+        end
+
+        it 'rejects when format is unspecified' do
+          expect(controller_with_defined_routes).
+            not_to route(:post, "/#{controller_path}").
+            to(action: 'create')
+        end
+      end
     end
 
     def controller_with_defined_routes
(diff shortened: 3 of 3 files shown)


lib/shoulda/matchers/action_controller/route_params.rb
```diff
@@ -26,17 +28,24 @@ def controller_and_action_given_as_string?
...
-            hash_copy
+        def normalize_values(hash)
+          hash.each_with_object({}) do |(key, value), hash_copy|
+            hash_copy[key] = symbolize_or_stringify(key, value)
+          end
+        end
+
+        def symbolize_or_stringify(key, value)
+          if key.in?(VALUES_TO_SYMBOLIZE)
+            value.to_sym
```
mcmire: This works obviously, although I wonder if it makes a little more sense to say "leave this value alone if the key is :format". In other words change this line to just `value`.

lib/shoulda/matchers/action_controller/route_matcher.rb
```diff
@@ -79,6 +79,10 @@ module ActionController
       #
       #     route(:get, '/posts/1').to('posts#show', id: 1)
       #
+      # If there are a default format in your route, then specify this too:
```
mcmire: What about:

```
You may also specify special parameters such as `:format`:
```

lib/shoulda/matchers/action_controller/route_params.rb
```diff
@@ -3,6 +3,8 @@ module Matchers
     module ActionController
       # @private
       class RouteParams
+        VALUES_TO_SYMBOLIZE = %i{ format }
```
mcmire: Going along with the comment below, maybe this could be `PARAMS_TO_IGNORE`?

spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
```diff
@@ -48,6 +48,20 @@
             to(action: 'show', some: 'other', params: 'here')
         end
       end
+
+      context 'when route has the format' do
```
mcmire: s/the/a

spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
```diff
@@ -48,6 +48,20 @@
...
+        it 'accepts' do
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to(action: 'create', format: 'json')
+        end
+
+        it 'accepts as a symbol' do
```
mcmire: "accepts when specified as a symbol"?

spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
```diff
@@ -48,6 +48,20 @@
...
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to(action: 'create', format: :json)
+        end
```
mcmire: What if you leave the format out?

```
expect(controller_with_defined_routes).
  to route(:post, "/#{controller_path}").
  to(action: 'create')
```

spec/unit/shoulda/matchers/action_controller/route_matcher_spec.rb
```diff
@@ -64,6 +78,14 @@
...
         end
       end
+
+      context 'when route has the format' do
+        it 'accepts' do
+          expect(controller_with_defined_routes).
+            to route(:post, "/#{controller_path}").
+            to("#{controller_path}#create", format: 'json')
+        end
+      end
```
mcmire: Same thing here, seems we need to test that it rejects if format is not specified.

lib/shoulda/matchers/action_controller/route_params.rb
maurogeorge: We need the `value.to_sym` if we want accept the format as string or symbol.

Since we accept controller and action as string and symbols, I think is a good idea we accept the `format` as string and symbols too. What you think?

maurogeorge:
@mcmire I updated the code please take a look when you have some time

lib/shoulda/matchers/action_controller/route_params.rb
mcmire: I could go either way. This is fine.

I wish there were a better way to say this actually. The only reason we stringify is for the params I believe, the other "special" params like controller, action, format, etc. don't need to be touched. Ah well.

mcmire:
Thank you @maurogeorge! This is great. Merged as 15359eb6ec4f99acd9b970d7129edc641ad68b57.

aried3r:
I'm not sure if I'm doing something wrong here or this is a bug, but I am facing this problem with shoulda-matchers 3.0.0. Will open an issue if it is. Or go to SO, if it isn't. 

Spec:
```Ruby
it { is_expected.to route(:get, '/projects/1.pdf').to(action: :show, id: 1, format: :pdf) }
```
Output:
```diff
The recognized options <{"controller"=>"projects", "action"=>"show", "id"=>"1", "format"=>"pdf"}> did not match <{"action"=>"show", "id"=>"1", "format"=>:pdf, "controller"=>"projects"}>, difference:.
--- expected
+++ actual
@@ -1 +1 @@
-{"action"=>"show", "id"=>"1", "format"=>:pdf, "controller"=>"projects"}
+{"controller"=>"projects", "action"=>"show", "id"=>"1", "format"=>"pdf"}
```

aried3r:
I found the 'error'.

In `routes.rb` I had:
```Ruby
namespace :api, defaults: { format: 'json' } do
```
which resulted in shoulda-matchers to compare my `:json` from the specs agains the `'json'` from the routes.

I read through #454, that Rails 4 supposedly uses symbols for the `format:` parameter, but I think both is possible, even the [official guide on routing](http://guides.rubyonrails.org/routing.html#defining-defaults) still uses strings.

See also the rails source for 4.2:
https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/routing/mapper.rb#L527-L531
and
https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/routing/mapper.rb#L193-L203

If I read the source correctly, but I haven't read the routing code of rails before, so I might be wrong. However, I think #454 might have been right in the sense that mixing is wrong, but I see no indication of the `format:` being a symbol in rails.

maurogeorge:
Hi @aried3r could you please create a issue and reference this PR? Since this PR is closed this can be lost in the time.

Thanks for the report

mcmire:
I actually ran into this problem while implementing this bug fix...

It's true that `format` can be a symbol or string, but if you're defining the format as a default parameter in a route as you're done, then if I remember correctly, Rails will store the format as it is and will not convert it to a string or symbol.

As you can see in the commit I referenced above, we're converting the `format` option you pass to the matcher to a symbol, and I think that's why you're seeing a test failure. By the way, I don't think we should be doing this, and I thought I'd made a change that took this away, but I guess that didn't happen. (Neither the `route_to` matcher in rspec-rails nor the Minitest assertion methods that Rails provides (`assert_recognizes` / `assert_routing`) do this.)

So yeah -- as @maurogeorge said, please make a new issue with a test that reproduces this and I'll take a closer look.

aried3r:
Done

PR #694, opened 2015-03-31
No more throws a error when use validate_uniqueness_of with scope of boolean column

### LINKED ISSUE #457: When using `validate_uniqueness_of` with scope of boolean column, throws error
I have the following code:
(`is_rental` is a `boolean` column)
```ruby
it {should validate_uniqueness_of(:remote_reference_id).scoped_to(:scraping_source_name, :is_rental).allow_nil}
```

Backtrace:
```
# /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:157:in `each'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:157:in `max'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:157:in `block in validate_after_scope_change?'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:156:in `each'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:156:in `all?'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:156:in `validate_after_scope_change?'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/shoulda-matchers-2.5.0/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb:76:in `matches?'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-expectations-2.14.5/lib/rspec/expectations/handler.rb:24:in `handle_matcher'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/memoized_helpers.rb:68:in `should'
     # ./spec/shared_examples/models/listings/listings_with_scraping.rb:20:in `block (5 levels) in <top (required)>'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `instance_eval'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:179:in `call'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:179:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/extensions/instance_eval_with_args.rb:16:in `instance_exec'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/extensions/instance_eval_with_args.rb:16:in `instance_eval_with_args'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:247:in `instance_eval_with_args'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/hooks.rb:106:in `block (2 levels) in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/hooks.rb:108:in `call'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/hooks.rb:108:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/hooks.rb:446:in `run_hook'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:340:in `run_around_each_hooks'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:256:in `with_around_each_hooks'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:111:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:390:in `block in run_examples'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:386:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:386:in `run_examples'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:371:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `block (2 levels) in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `map'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `block in run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/reporter.rb:58:in `report'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:25:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/runner.rb:80:in `run'
     # /Users/PikachuEXE/.rvm/gems/ruby-2.0.0-p451/gems/rspec-core-2.14.8/lib/rspec/core/runner.rb:17:in `block in autorun'
```
So it should be [this line](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb#L170)

maurogeorge:
closes #457

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -257,6 +257,7 @@ module Shoulda
           @original_subject = subject
           @subject = subject.class.new
           @expected_message ||= :taken
+          @all_records = @subject.class.all
 
           set_scoped_attributes &&
             validate_everything_except_duplicate_nils_or_blanks? &&
@@ -397,12 +398,11 @@ module Shoulda
         end
 
         def validate_after_scope_change?
-          if @options[:scopes].blank?
+          if @options[:scopes].blank? || all_scopes_are_booleans?
             true
           else
-            all_records = @subject.class.all
             @options[:scopes].all? do |scope|
-              previous_value = all_records.map(&scope).compact.max
+              previous_value = @all_records.map(&scope).compact.max
 
               next_value =
                 if previous_value.blank?
@@ -447,6 +447,8 @@ module Shoulda
             DateTime.now
           when :uuid
             SecureRandom.uuid
+          when :boolean
+            true
           else
             'dummy value'
           end
@@ -474,9 +476,21 @@ module Shoulda
             previous_value.next
           elsif previous_value.respond_to?(:to_datetime)
             previous_value.to_datetime.next
+          elsif boolean_value?(previous_value)
+            !previous_value
           else
             previous_value.to_s.next
           end
+        end
+
+        def all_scopes_are_booleans?
+          @options[:scopes].all? do |scope|
+            @all_records.map(&scope).all? { |s| boolean_value?(s) }
+          end
+        end
+
+        def boolean_value?(value)
+          value.in?([true, false])
         end
 
         def defined_as_enum?(scope)
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -283,6 +283,43 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
         column_type: :string
     end
 
+    context 'when one of the scoped attributes is a boolean column' do
+      include_context 'it supports scoped attributes of a certain type',
+        column_type: :boolean
+
+      context 'when there is more than one scoped attribute and all are boolean columns' do
+        it 'accepts when all the scoped attribute are true' do
+          record = build_record_validating_uniqueness(
+            scopes: [
+              build_attribute(name: :scope1),
+              build_attribute(name: :scope2),
+            ]
+          )
+          expect(record).to validate_uniqueness.scoped_to(:scope1, :scope2)
+        end
+
+        it 'accepts when all the scoped attribute are false' do
+          record = build_record_validating_uniqueness(
+            scopes: [
+              build_attribute(name: :scope1, value: false),
+              build_attribute(name: :scope2, value: false),
+            ]
+          )
+          expect(record).to validate_uniqueness.scoped_to(:scope1, :scope2)
+        end
+
+        it 'accepts when one scoped attribute is true and other is false' do
+          record = build_record_validating_uniqueness(
+            scopes: [
+              build_attribute(name: :scope1),
+              build_attribute(name: :scope2, value: false),
+            ]
+          )
+          expect(record).to validate_uniqueness.scoped_to(:scope1, :scope2)
+        end
+      end
+    end
+
     context 'when one of the scoped attributes is an integer column' do
       include_context 'it supports scoped attributes of a certain type',
         column_type: :integer
@@ -752,6 +789,8 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
       Time.now
     when :uuid
       SecureRandom.uuid
+    when :boolean
+      true
     else
       raise ArgumentError, "Unknown type '#{attribute_type}'"
     end
@@ -764,6 +803,8 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
       SecureRandom.uuid
     elsif value.is_a?(Time)
       value + 1
+    elsif value.in?([true, false])
+      !value
     elsif value.respond_to?(:next)
       value.next
     end


lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -402,7 +402,14 @@ def validate_after_scope_change?
           else
             all_records = @subject.class.all
             @options[:scopes].all? do |scope|
-              previous_value = all_records.map(&scope).compact.max
+              previous_value =
+                if is_a_boolean_scope?(all_records, scope)
+                  value = all_records.map(&scope).last
+                  @existing_record.destroy
```
mcmire: Why do we need to destroy the record here? I would think we'd want to keep it around, that way we're properly testing the validation.

maurogeorge: When we are dealing with date, string and integers for example we do not have this problem because the `next_value` is always a unused value.

```
all_records
=> [#<Example:0x007f9a932a56b8 id: 1, attr: "dummy value", scope: 2>, #<Example:0x007f9a932a53e8 id: 2, attr: "dummy value", scope: 3>]

previous_value
=> 3

next_value
=> 4

allows_value_of(existing_value, @expected_message)
=> true
```

But when we are dealing with a boolean our `next_value` will always be equal the first one created by the matcher

```
all_records
=> [#<Example:0x007f9a96982d48 id: 1, attr: "dummy value", scope: false>, #<Example:0x007f9a969825c8 id: 2, attr: "dummy value", scope: true>]

previous_value
=> true

next_value
=> false

allows_value_of(existing_value, @expected_message)
=> true
```

So if I not delete the first record, the `allows_value_of` will always be `false` since the `scope` is already defined on the first record, and here we can define only two value, `true` and `false`.

I really don't know if this is the better approach, if you have any idea let me know. I will dig around to see if I found another way to do that.

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -283,6 +283,11 @@
         column_type: :string
     end
 
+    context 'when one of the scoped attributes is a boolean column' do
+      include_context 'it supports scoped attributes of a certain type',
+        column_type: :boolean
```
mcmire: I think the tests for these are going to be a little different, because there's a limited number of values for a boolean column.

If you have one existing record where the scope attribute is either true or false, then you can simply flip that scope attribute. If you have two existing records where one of the record's scope attribute is true and the other is false... then you can't flip the scope attribute.

Say you have a validation on an attribute against a boolean scope:

```
class Person < ActiveRecord::Base
  validates :name, scope: :snowflake
end
```

Here's what a test would look like if there's only one existing record:

```
describe Person do
  it "is valid with a unique name when snowflake is true" do
    create(:person, name: "Steve", snowflake: true)

    person = build(:person, name: "Steve", snowflake: true)
    person.valid?
    expect(person).errors[:name].not_to be_empty

    person = build(:person, name: "Joe", snowflake: true)
    person.valid?
    expect(person).errors[:name].to be_empty

    # here we can test against the snowflake attribute -
    # i.e. if name stays same but snowflake is different
    person = build(:person, name: "Steve", snowflake: false)
    person.valid?
    expect(person).errors[:name].to be_empty
  end
end
```

But now consider if there are two existing records. Here's what those tests might look like:

```
describe Person do
  it "is valid with a unique name when snowflake is true" do
    create(:person, name: "Steve", snowflake: true)
    create(:person, name: "Steve", snowflake: false)

    person = build(:person, name: "Steve", snowflake: true)
    person.valid?
    expect(person).errors[:name].not_to be_empty

    person = build(:person, name: "Joe", snowflake: true)
    person.valid?
    expect(person).errors[:name].to be_empty

    # we can't really test what happens if name stays same and snowflake is different
    # b/c all values for snowflake are filled - so let's just test what happens if name is
    # different when snowflake is false
    person = build(:person, name: "Steve", snowflake: false)
    person.valid?
    expect(person).errors[:name].not_to be_empty

    person = build(:person, name: "Joe", snowflake: false)
    person.valid?
    expect(person).errors[:name].not_to be_empty
  end
end
```

So now the question is... how do we modify the matcher to run these tests. I don't have a clear solution for this, but it seems like as soon as one of the scopes is a boolean attribute, it kind of messes with the existing logic...

lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
mcmire: Gotcha. Yeah I was thinking of this exact case, I made some comments below.

maurogeorge:
@mcmire I updated the code and skip the `when more than one record exists that has the next version of the attribute's value` if the value type is `boolean`.

My thought was, if the scope is a boolean and we define the first record as `true` the second one as `false` we do not need this test, since this expect a third valid option, and since we are dealing with booleans we can remove this test and have a test that rejects when we already have 2 records one as `true` and other as `false`.

This makes sense to you?

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -283,6 +285,27 @@
...
+          value2 = next_version_of(value1, :boolean)
+          value3 = next_version_of(value2, :boolean)
+          model = define_model_validating_uniqueness(
+            scopes: [ build_attribute(name: :scope) ]
+          )
+          create_record_from(model, scope: value2)
+          create_record_from(model, scope: value3)
+          record = build_record_from(model, scope: value1)
+
+          expect(record).not_to validate_uniqueness.scoped_to(:scope)
```
mcmire: This should actually accept, I think...

So... this is something I actually noticed a while back. We're actually missing a check here. If you're validating uniqueness of a given attribute and it's scoped, then the first thing you should check is, if I change the value of that attribute to a unique value, then it should make the record valid. But currently we just loop through all the scope attributes and assert that if we change those to unique values then it should make the record valid.

But even if we added this check then the matcher would still fail. This is because we are still trying to check that column within #validate_after_scope_change?. We need to somehow exclude the check in this case, since both true and false are taken as possible next values.

maurogeorge:
@mcmire I updated, now the `#validate_after_scope_change?` will return true when the scope is a boolean.

mcmire:
Thanks @maurogeorge, that looks better. I'll look at this in more detail later to make sure we're doing the right thing here.

mcmire:
I think this looks correct. But... (and I know you're going to hate me for this) can we add some tests for:

* when one of the scoped attributes is a boolean column but another is a different column
* when there is more than one scoped attribute and all are boolean columns

lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -474,11 +476,23 @@ def next_scalar_value_for(scope, previous_value)
             previous_value.next
           elsif previous_value.respond_to?(:to_datetime)
             previous_value.to_datetime.next
+          elsif a_boolean?(previous_value)
```
mcmire: Perhaps we can rename this to `boolean_value?`

lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -474,11 +476,23 @@ def next_scalar_value_for(scope, previous_value)
...
+            !previous_value
           else
             previous_value.to_s.next
           end
         end
 
+        def a_boolean_scope?
```
mcmire: Perhaps we can rename this to `all_scopes_are_booleans?`

maurogeorge:
@mcmire updated please take a look.

mcmire:
Thanks! Merged as 7c8e3f5081226568a20724f12e707395973b2eb5.

PR #695, opened 2015-04-01
support allow_nil with validates_presence_of

axlekb:


lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -57,6 +57,29 @@ module ActiveModel
       #
       # #### Qualifiers
       #
+      # ##### allow_nil
+      #
+      # Use `allow_nil` if you want to validate a `belongs_to` association but
```
mcmire: Does this have to be a `belongs_to`, can't it be any attribute?

I would use the same description we use for the other instances of `allow_nil`, which is:

> Use `allow_nil` to assert that the attribute allows nil.

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -111,16 +134,28 @@ def validate_presence_of(attr)
...
+          @options = {}
+        end
+
         def with_message(message)
           @expected_message = message if message
           self
         end
 
+        def allow_nil(allow_nil = true)
+          @options[:allow_nil] = allow_nil
```
mcmire: While this makes more sense, I think currently in this file we use instance variables to store options, so let's do that.

Also, this qualifier doesn't tend to take an argument (`validate_inclusion_of` seems to be the only one that does this). Let's just leave it out for now, I might go through later and add in optional arguments for boolean qualifiers.

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -111,16 +134,28 @@ def validate_presence_of(attr)
...
+          self
+        end
+
         def matches?(subject)
           super(subject)
           @expected_message ||= :blank
 
-          if secure_password_being_validated?
+          if @options[:allow_nil]
```
mcmire: Using an `elsif` here will make it so that we no longer check that the attribute is invalid with a blank value, and I think we still need that.

To make it so that we don't contradict ourselves when `allow_nil` is specified, instead of the `blank_value` method below returning `nil`, what if it returns an empty string?

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -39,6 +39,23 @@
     end
   end
 
+  context 'a belongs_to association with a presence validation and allow_nil' do
```
mcmire: Again, it is necessary for this to be a `belongs_to` association? I realize that may be what you added this for, but I believe this can just be a normal attribute.

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -39,6 +39,23 @@
...
+    it 'allows a nil value' do
+      if ::ActiveModel::VERSION::STRING.to_f >= 4.0
+        child = define_parent_and_child_belongs_to_allow_nil.new
+        expect(child).to validate_presence_of(:parent).allow_nil
+      end
+    end
+
+    def define_parent_and_child_belongs_to_allow_nil
```
mcmire: Instead of defining this method, can we put these things right in the test?

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -39,6 +39,23 @@
     end
   end
 
+  context 'a belongs_to association with a presence validation and allow_nil' do
+    it 'allows a nil value' do
+      if ::ActiveModel::VERSION::STRING.to_f >= 4.0
```
mcmire: You don't need this check here -- we dropped support for Rails 3.x and now support 4+.

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
mcmire: Actually, I'm noticing now that we don't seem to have tests for a `belongs_to` association at all. Can we two tests -- one for `belongs_to`, the other for `allow_nil` (doesn't have to be a `belongs_to`)?

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -116,12 +138,19 @@ def with_message(message)
...
+        end
+
         def matches?(subject)
           super(subject)
           @expected_message ||= :blank
 
           if secure_password_being_validated?
             disallows_and_double_checks_value_of!(blank_value, @expected_message)
+          elsif @allow_nil
+            allows_value_of(nil)
```
axlekb: @mcmire struggling a bit here to elegantly test both association AND simple attribute because `disallows_value_of(blank_value, @expected_message)` fails due to a [ActiveRecord association type mismatch](https://github.com/rails/rails/blob/v4.0.0/activerecord/lib/active_record/associations/association.rb#L213) for empty string.

Also realizing that I'm not even validating that the `belongs_to` record exists.  This seems like a big effort.

Plus I'm also getting concerned that this is in the `active_model` directory and not `active_record`.

mcmire: Okay, so this raises a question... If your `belongs_to` can be nil, but it can't be an empty string... then it can't ever be blank, can it? What are you trying to protect against?

axlekb: As I understand it, when you `validate_presence_of` a `belongs_to` association it checks to make sure that record exists.  Rails actually queries the DB to make sure that a record exists matching the `belongs_to_id`.

mcmire: Right, so `validate_presence_of` typically accesses the attribute (or association) you specify, then calls `.present?` on it and checks that this returns true; if not, then it adds a validation error.

If you pass an `:allow_nil` to `validate_presence_of`, you're telling it, don't bother with running the validation if the association is nil.

So if the association is not nil, then it will check to make sure it isn't nil (this check will always pass). If the association is nil, then no check occurs.

So there's never a case in which `.present?` returns false, which means your validation isn't really doing anything -- your record will be valid regardless of whether the association is set or not.

axlekb: Yup, looks like I was mistaken about how the `allow_nil` worked here, damn.  I thought it was allowing the associated instance to be `nil`, but ONLY IF the `belongs_to_id` was `nil`, that is, if the instance was not found, it became invalid.

Looks like I can close this PR.  Sorry for your time.

mcmire: It's ok! No worries. I didn't catch it at first.

aprescott:
I'm unable to tell if this was closed because it was merged, or for some other reason. (I can't find any comments about it either way.) Is this planned to be included in a future release?

aprescott:
Oh, I misunderstood where the comment chain at https://github.com/thoughtbot/shoulda-matchers/pull/695#discussion-diff-27579531 was going. Looks like it was closed because this implementation doesn't quite work?

mcmire:
@aprescott It's because the case for which this is presented is invalid.

`validates_presence_of` adds a restriction to an attribute such that it can't be blank. The `allow_nil` option changes the restriction so that the attribute can be nil, but not blank in any other way. There's no way that a foreign key attribute (e.g. `foo_id`) can be blank in any other way, as it's an integer column, and as such, all other blank values will be typecast to an integer.

aprescott:
@mcmire we have a case where a varchar/string column can be `nil` but shouldn't be `""` (or any other `blank?` value). We're tackling that with a regular validation:

```ruby
validates :reason, presence: { allow_nil: true }
```

Because `validates_presence_of(:reason).allow_nil` doesn't exist, we've resorted to

```ruby
should allow_value(nil).for(:reason)
should_not allow_value("").for(:reason)
should_not allow_value(" ").for(:reason)
```

which sort of covers it, but the existence of `allow_nil` would of course be clearer and mirror the validation structure.

mcmire:
Okay, I'll re-open this issue, then.

aprescott:
Thanks!

aprescott:
This is still closed. Can it be reopened?

aprescott:
Oh, oops, I missed your comment edit!

PR #697, opened 2015-04-02
Use pry-byebug instead of pry-nav

takiy33:
I think it is better to use pry-byebug beceause shoulda-matcher require Ruby 2.0 or higher.
See also [pry-nav](https://github.com/nixme/pry-nav)

Thank you.


mcmire:
Yup, good idea.

mcmire:
Thanks! Merged as 05a11ae90c9159256f1c1b7c57eb77c232a0fb44.

PR #699, opened 2015-04-10
Add more tests to the validate_numericality_of error message

### LINKED ISSUE #509: validate_numericality_of on failure may not provide the reason for failure
Given this validation

``` ruby
validates :ano, numericality: { only_integer: true, greater_than_or_equal_to: 1900 }
```

and this test:

``` ruby
it { should validate_numericality_of(:ano).is_greater_than_or_equal_to(1900) }
```

It will fail with:

```
Did not expect errors  when ano is set to 1900.000000000001, got error:
```

As you can see the failure message doesn't contain the validation error.

I believe this is because none of the submatchers for ValidateNumericalityOfMatcher set a default message.

Really, all of the submatchers should share the same message, and updating that message should propagate to all submatchers.

maurogeorge:
Closes #509

### DIFF
--- a/lib/shoulda/matchers/active_model/validator.rb
+++ b/lib/shoulda/matchers/active_model/validator.rb
@@ -49,7 +49,7 @@ module Shoulda
         end
 
         def messages_description
-          if has_messages?
+          if has_messages? && !record.valid?
             " errors:\n#{pretty_error_messages(record)}"
           else
             ' no errors'
--- a/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
@@ -512,11 +512,41 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
       end
 
       it do
+        record = build_record_validating_numericality(
+          even: true,
+          greater_than: 18
+        )
+        assertion = lambda do
+          expect(record).
+            to validate_numericality.
+            only_integer.
+            is_greater_than(18)
+        end
+        expect(&assertion).to fail_with_message_including(
+          '* "must be even" (attribute: attr, value: "19")'
+        )
+      end
+
+      it do
         record = build_record_validating_numericality(greater_than: 18)
         expect(record).
           not_to validate_numericality.
           only_integer.
           is_greater_than(18)
+      end
+
+      it do
+        skip 'Need to be fixed the code, actually do not have the right message'
+        record = build_record_validating_numericality(greater_than: 18)
+        assertion = lambda do
+          expect(record).
+            to validate_numericality.
+            only_integer.
+            is_greater_than(18)
+        end
+        expect(&assertion).to fail_with_message_including(
+          ''
+        )
       end
 
       it do
@@ -528,6 +558,22 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
           not_to validate_numericality.
           even.
           is_greater_than(18)
+      end
+
+      it do
+        record = build_record_validating_numericality(
+          even: true,
+          greater_than_or_equal_to: 18
+        )
+        assertion = lambda do
+          expect(record).
+            to validate_numericality.
+            even.
+            is_greater_than(18)
+        end
+        expect(&assertion).to fail_with_message_including(
+          %{Expected errors to include "must be greater than 18" when attr is set to 18,\ngot no errors}
+        )
       end
 
       it do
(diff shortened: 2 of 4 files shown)


lib/shoulda/matchers/active_model/validator.rb
```diff
@@ -49,7 +49,7 @@ def has_messages?
         end
 
         def messages_description
-          if has_messages?
+          if has_messages? && !record.valid?
```
mcmire: And this change was to specifically fix https://github.com/maurogeorge/shoulda-matchers/commit/84b7175ddf85dc3ad406ac1c81335d03135268cf#diff-d7c315235a16eb29151b8251207fbd67R539, correct?

mcmire:
Two things:

* Now that we have tests for the failure message I think we can remove the tests above that simply say "such and such `to_not validate_numericality`", they're now duplicates.
* What about testing the entire failure message instead of just a portion of it? It shouldn't be that much more typing and it gives a more complete picture for someone reading the test.

lib/shoulda/matchers/active_model/validator.rb
maurogeorge: Right!

maurogeorge:
@mcmire I updated to assert against the full message.

I will check the broken build, the problem is on Rails 4.2 the string value is showed as string on the `pretty_error_messages`.

```
> obj
=> #<Example:0x007ffce10121e8 id: nil, attr: "19">
> obj.attr
=> "19"
```

On the other versions, the value is showed as integer

```
> obj
=> #<Example id: nil, attr: 19>
> obj.attr
=> 19
```

If you have any idea of this, let me know.

mcmire:
Oh weird... yeah, I'll have to take a closer look at that.

maurogeorge:
@mcmire after some research I belive this is the behavior of the AR, I update the specs to handle this, please take a look.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -505,18 +505,37 @@ def default_validation_values
...
+            Expected errors to include "must be an integer" when attr is set to 0.1,
+            got errors:
+            * "must be even" (attribute: attr, value: "19")
+          MESSAGE
+        )
+        expect(&assertion).to fail_with_message(message)
       end
 
-      it do
+     it do
```
mcmire: I think this line might be off indentation-wise.

lib/shoulda/matchers/active_model/validator.rb
mcmire: This still seems odd to me, I'm investigating why we have to do this. Seems odd that the record has error messages but then it's valid at the same time.

mcmire:
So I took a look at why we are even seeing the weird "got error:" message and I think this is what's happening:

* ValidateNumericalityOfMatcher has two submatchers, OnlyIntegerMatcher and ComparisonMatcher
* Both of these submatchers change the record in question somehow and then call #valid? on that record
* ValidateNumericalityOfMatcher will run all of the submatchers and determine if they pass or fail (this is determined by whether #valid? produces any error messages on the attribute being validated)
* In this case, OnlyIntegerMatcher fails but ComparisonMatcher passes
* So ValidateNumericalityOfMatcher fails, but then when it goes to print its failure message, it accesses the record's error messages. Because ComparisonMatcher passes, the record is in a good state and doesn't have any error messages
* In other words, the error messages that Validator#messages accesses are cached at the time that #valid? is called... but the error messages that Validator#messages_description accesses are "live"
* So... either we need Validator to cache not only the error messages on the attribute being validated, but also cache all error messages on the record... or we need ValidateNumericality to *not* run all submatchers and stop at the first one that fails. So in this case OnlyIntegerMatcher would run, it would see that it fails, and it would never run ComparisonMatcher, thereby keeping the record in a "bad" state (so the cached error messages and live error messages would be the same thing).

I would probably prefer the latter solution, but it does mean that all of the tests that involve failure messages will change.

mcmire:
Basically, this is what I mean: https://github.com/thoughtbot/shoulda-matchers/compare/2dc787e6454458296f4d55c9aec465f4960b9bb3...9e0e8f7b670fee9d2e540ba70e6d5964ccfb39ae. This seem fine to you?

maurogeorge:
@mcmire thanks, I reviewed and looks good to me :shipit:

maurogeorge:
@mcmire ping :smiley: . Just to remember you to merge this, like you described here https://github.com/thoughtbot/shoulda-matchers/pull/699#issuecomment-93851307

mcmire:
Thanks! This was merged as e7087897143dfd5f59e3c512c31a8c7976ad96dd.

PR #704, opened 2015-04-21
Add doc to join_table qualifier on have_and_belong_to_many matcher

maurogeorge:
Add the docs to the new qualifier created here https://github.com/thoughtbot/shoulda-matchers/commit/af85ac1f0e5b9a267efd4c05b30a73fe2f44d84e

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -774,6 +774,29 @@ module Shoulda
       #         class_name('City')
       #     end
       #
+      # ##### join_table
+      #
+      # Use `join_table` to test usage of the `:join_table` option. This
+      # asserts that the table you're referring to actually exists.
+      #
+      #     class Person < ActiveRecord::Base
+      #       has_and_belongs_to_many :issues, join_table: 'people_tickets'
+      #     end
+      #
+      #     # RSpec
+      #     describe Person do
+      #       it do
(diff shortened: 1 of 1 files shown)


lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -774,6 +774,29 @@ def have_one(name)
...
+      #
+      #     class Person < ActiveRecord::Base
+      #       has_and_belongs_to_many :issues, join_table: 'people_tickets'
+      #     end
+      #
+      #     # RSpec
+      #     describe Person do
+      #       it do
+      #         should have_and_belong_to_many(:issues).
+      #           join_table('people_tickets')
```
mcmire: This should be `people_issues`, I think.

maurogeorge: In [this line](https://github.com/thoughtbot/shoulda-matchers/pull/704/files#diff-0a438a581d903b7f45fdb3811e6e9291R783) I defined a custom `join_table`.

My rationale is that this option is more used when you have a custom join table, I don't think we need to test join table on a default HABTM.

mcmire: Oh, no, you're right, I wasn't looking at this closely.

mcmire:
Merged as b4012404d9cea02ed03dc0890afc95ede88c44a6. Thanks!

Issue #579, opened 2014-09-04
`validate_inclusion_of` matcher throws error


glittershark:
Using the `validates_inclusion_of` matcher on 2.7.0 displays the following error:
```
Failure/Error: it { is_expected.to validate_inclusion_of(:name) }
     ArgumentError:
       Failure message is nil. Does your matcher define the appropriate failure_message[_when_negated] method to return a string?
```

mcmire:
I'm going to close this in favor of the PRs listed above ^

PR #708, opened 2015-04-29
Return a better error message when call validate_exclusion_of without a qualifier

maurogeorge:
The `validate_exclusion_of` has the same misbehavior of the `validate_inclusion_of` as described here https://github.com/thoughtbot/shoulda-matchers/issues/579

This PR adds a better error message.


mcmire:

mcmire:

PR #705, opened 2015-04-21
Return a better error message when call validate_inclusion_of without a qualifier

maurogeorge:
Closes #579

PR #709, opened 2015-04-29
Return a error message when call validate_length_of without a qualifier

maurogeorge:
The `validate_length_of` returns a false positive when called without a qualifier, example:

It is positive when the model does not define a validation.

``` ruby
expect(validating_length).to validate_length_of(:attr)
```

It is positive when the attribute does not exist in the model.

``` ruby
expect(validating_length(minimum: 4)).to validate_length_of(:bar)
```

This PR raises a error message when the matcher is called without a qualifier, follows the same logic of the https://github.com/thoughtbot/shoulda-matchers/pull/705 and https://github.com/thoughtbot/shoulda-matchers/pull/708

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
@@ -291,6 +291,14 @@ module Shoulda
         end
 
         def matches?(subject)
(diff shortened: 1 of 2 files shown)


mcmire:

mcmire:

PR #717, opened 2015-05-15
Add `:reject_if` for `accept_nested_attributes_for`

franckverrot:
This commit implements the `:reject_if` option discussed in #98.

`:reject_if` supports literals, `Proc`'s and `Symbol`s:

``` ruby
# Proc object
accepts_nested_attributes_for :mirrors, reject_if: proc { |obj| obj.count != 2 }

# Symbol
accepts_nested_attributes_for :mirrors, reject_if: :different_than_2?

# Literal
accepts_nested_attributes_for :mirrors, reject_if: false
```

### DIFF
--- a/lib/shoulda/matchers/active_record/accept_nested_attributes_for_matcher.rb
+++ b/lib/shoulda/matchers/active_record/accept_nested_attributes_for_matcher.rb
@@ -43,6 +43,36 @@ module Shoulda
       #         allow_destroy(true)
       #     end
       #
+      # Use `reject_if` to assert that the `:reject_if` option was
+      # specified, with a Proc (or lambda with an arity of 1) or a
+      # plain object.
+      #
+      #     class Car < ActiveRecord::Base
+      #       accepts_nested_attributes_for :mirrors,
+      #         reject_if: proc { |obj| obj.count != 2 }
+      #
+      #       accepts_nested_attributes_for :mirrors,
+      #         reject_if: :different_than_2?
+      #
+      #       def different_than_2?
+      #         mirrors.count != 2
+      #       end
+      #     end
+      #
+      #     # RSpec
+      #     describe Car do
+      #       it do
+      #         should accept_nested_attributes_for(:mirrors).
+      #           reject_if(:different_than_2?)
+      #       end
+      #     end
+      #
+      #     # Test::Unit
+      #     class CarTest < ActiveSupport::TestCase
+      #       should accept_nested_attributes_for(:mirrors).
+      #         reject_if(:different_than_2?)
+      #     end
+      #
       # ##### limit
       #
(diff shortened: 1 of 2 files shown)


mcmire:

aried3r:
@mcmire, I'd love to see this in the next release. Is there anything I could do to help? :)

mcmire:
@aried3r @franckverrot Forgive the necromancy, but I'm going back through stale issues and revisiting them. I came across this one and tried to revive it, and I ran into a bit of an issue. See, I'm not sure that this addition makes sense, at least in its present form. The reason for this is due to how the `:reject_if` option for `accepts_nested_attributes_for` works. I took the example from the documentation that was added here and modified it a bit:

```
class Car < ActiveRecord::Base
  accepts_nested_attributes_for :wheels,
    reject_if: :square_wheel?

  accepts_nested_attributes_for :mirrors,
    reject_if: proc { |mirror| mirror[:state] != 'broken' }

  def square_wheel?(wheel)
    wheel[:shape] == 'square'
  end
end
```

Now, what would the tests for this look like? In #98, a test for the first case would have actually looked like this:

```
describe Car do
  it { should accept_nested_attributes_for(:wheels).reject_if(:square_wheel?) }
end
```

Interestingly, in this PR, it seems the behavior of the qualifier was actually changed so that it now looks like:

```
describe Car do
  it { should accept_nested_attributes_for(:wheels).reject_if(true) }
end
```

Regardless, I am not so sure that either of these solutions makes real sense. The reason for this is that the proc/method reference that is given to the `:reject_if` option is intended to be used in the act of receiving nested attributes. If you call that proc/method on its own, what are you supposed to pass to it? You could pass a hash to it, but that hash needs to have certain keys in order to be useful, and that's impossible for the matcher to know. What you'd really have to do is do something like:

``` ruby
car = Car.new
car.wheels_attributes = [{ shape: 'square' }, { shape: 'round' }]
expect(car.wheels).to match([
  an_object_having_attributes(shape: 'round')
])
```

*(forgive any mistakes here, I haven't used `accepts_nested_attributes_for` in a while, but you get the idea)*

That would be way better, in my opinion, than simply testing that `accepts_nested_attributes_for` was defined with a certain set of options. Thing is, the `accept_nested_attributes_for` doesn't really work like this (a lot of our matchers are kind of naive in this way, in fact). There's no reason it couldn't, of course, but that seems like a separate PR.

For this reason, and considering this issue is so old, I'm going to close this. But if this feature is still desirable and you are able to take an existing test you've written and turn it into a matcher, then I think I would feel a lot better about adding this.

franckverrot:
Hey there.  I don’t have anymore use case for this so if anyone wants to take over I’ll happily let them continue.  Thanks for spending time reviewing this though.

PR #732, opened 2015-06-11
Fix the set_flash matcher to accept a symbol as a param

### LINKED ISSUE #728: set_flash does not accept a symbol as key on 3.0.0.rc1
My spec is:

```ruby
post :create, params
is_expected.to set_flash[:notice].to(/sucesso/)
```
My controller redirect is

```ruby
redirect_to root_path, notice: 'Resposta criada com sucesso!'
```

This works on 2.7 but when I update to 3.0.0.rc1 the specs fail.
If I change the the key to a string the specs pass.

```ruby
post :create, params
is_expected.to set_flash['notice'].to(/sucesso/)
```
I think the expected behavior is that we accept string and symbol as keys.

maurogeorge:
This is a regression bug fix, on the previous version the matcher
accepts a string or a symbol as a param.

This commit brings back this behavior.

closes #728

### DIFF
--- a/spec/support/unit/shared_examples/set_session_or_flash.rb
+++ b/spec/support/unit/shared_examples/set_session_or_flash.rb
@@ -56,9 +56,28 @@ shared_examples_for 'set session or flash matcher' do
 
     context 'in the positive' do
       context 'if the given key is present in the store' do
-        it 'accepts' do
-          controller = controller_with_store('the key' => 'any value')
-          expect(controller).to set_store['the key']
+        context 'and controller set the key as a string' do
+          it 'accepts the param as a string' do
+            controller = controller_with_store('the_key' => 'any value')
+            expect(controller).to set_store['the_key']
+          end
+
+          it 'accepts the param as a symbol' do
+            controller = controller_with_store('the_key' => 'any value')
+            expect(controller).to set_store[:the_key]
+          end
+        end
+
+        context 'and controller set the key as a symbol' do
+          it 'accepts the param as a string' do
+            controller = controller_with_store(the_key: 'any value')
+            expect(controller).to set_store['the_key']
+          end
+
+          it 'accepts the param as a symbol' do
+            controller = controller_with_store(the_key: 'any value')
+            expect(controller).to set_store[:the_key]
+          end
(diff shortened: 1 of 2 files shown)


spec/support/unit/shared_examples/set_session_or_flash.rb
```diff
@@ -56,9 +56,28 @@
 
     context 'in the positive' do
       context 'if the given key is present in the store' do
-        it 'accepts' do
-          controller = controller_with_store('the key' => 'any value')
-          expect(controller).to set_store['the key']
+        context 'and controller set the key as a string' do
```
mcmire: s/set/sets/

spec/support/unit/shared_examples/set_session_or_flash.rb
```diff
@@ -56,9 +56,28 @@
...
+            expect(controller).to set_store['the_key']
+          end
+
+          it 'accepts the param as a symbol' do
+            controller = controller_with_store('the_key' => 'any value')
+            expect(controller).to set_store[:the_key]
+          end
+        end
+
+        context 'and controller set the key as a symbol' do
```
mcmire: Do we need these tests? I think that `controller_with_store(the_key: 'any value')` will translate to `flash[:the_key] = 'any value'` which should work regardless of what is given to the matcher. This seems like testing Rails.

spec/support/unit/shared_examples/set_session_or_flash.rb
mcmire: (but, see comment below)

spec/support/unit/shared_examples/set_session_or_flash.rb
maurogeorge: You are right, I checked the `values_to_check` on the `FlashStore#has_key?` and the values are always a string.

I updated the code, let's wait for Travis.

mcmire:
Thanks! Merged as c6cc26d4febe62f950adaea11f4b6a1f823f5a51.

maurogeorge:
Thanks for the merge :heart:

PR #733, opened 2015-06-13
Run formatter documentation when running a single file

maurogeorge:
Today if we run a single file or spec like `callback_matcher_spec.rb` or `callback_matcher_spec.rb:12` the formatter is the default dot.

With this config when we run a single file or spec the formatter is the documentation and when we run the suite the formatter is the default dot.

I think this is good when we are working since we have the documentation output when we are doing the TDD and the regular formatter on running all suite.

### DIFF
--- a/spec/unit_spec_helper.rb
+++ b/spec/unit_spec_helper.rb
@@ -30,6 +30,10 @@ RSpec.configure do |config|
     c.syntax = :expect
   end
 
+  if config.files_to_run.one?
+    config.default_formatter = 'doc'
+  end
+
   config.mock_with :rspec
 
   if config.respond_to?(:infer_spec_type_from_file_location!)


mcmire:
This is a good idea. What about making it so that it's the default formatter regardless of how many files you're running? Also I'm wondering if it'd be good to add this to the integration_spec_helper too?

maurogeorge:
I agree on adding this on the `integration_spec_helper`.

About the doc formatter when running more than one file I personally don't like because you see a lot of output that is impossible to read, since this will be scrolling in the window.

But if you prefer I can update to this.

mcmire:
Thanks! What you have works for me.

maurogeorge:
Thanks for the merge @mcmire

PR #737, opened 2015-06-24
Allow the use of matchers with just rspec expectations gem

majioa:
Allow the use of matchers for the just rspec expectations, apart the whole RSpec.

 * It adds the specific test framework that just includes the specified
   libraries directly to RSpec::Matchers.

In configuration block

    with.test_framework :rspec_exp

should be used.

The addition is covered with an acceptance test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ complex, and error-prone.
 
 ----
 
+**NOTE: Shoulda Matchers is currently undergoing a changing of the guard. We will not be as responsive to answer questions or issues while we make this transition. Sorry about that.**
+
+----
+
 ### ActiveModel matchers
 
 * **[allow_mass_assignment_of](lib/shoulda/matchers/active_model/allow_mass_assignment_of_matcher.rb)**


maurogeorge:
Hi @majioa thanks for your PR :100: 

I have a doubt, do you have a real test case for this feature? I mean the [RSpec, Minitest and Test-unit](https://www.ruby-toolbox.com/categories/testing_frameworks) seems to be the most used unit test framework and shoulda-matchers support they.

The RSpec Expectations is to be used with [other test framework, like Minitest and Test-unit](http://www.relishapp.com/rspec/rspec-expectations/docs/test-frameworks/minitest-integration). Your test case is for example: the user that use [bacon](https://github.com/chneukirchen/bacon) with [rspec-expectations](https://github.com/rspec/rspec-expectations)?

majioa:
@maurogeorge the real test case is the cucumber + rspec-expectations + shoulda-matchers, in one of projects of mine.

maurogeorge:
@majioa thanks for the feedback.

I think the goal of the shoulda-matchers is to provide matchers to be used on unit tests, not on acceptance tests, which is the case of cucumber. Since the purpose of the tests are different.

Let's wait for a feedback from @mcmire.

mcmire:
@majioa I'm not completely opposed to this, but this is definitely a new use case I haven't seen before. I realize that RSpec::Expectations can be used outside of RSpec, but as mentioned, shoulda-matchers wasn't designed to be used outside of RSpec and Minitest (the `it { should do_whatever }` pattern works really nicely here). That doesn't mean it isn't a good idea, of course, but I haven't seen anyone try to use it before. I guess my fear, really, is that if we merged this, we'd be supporting a use case that wouldn't be utilized that often -- I can easily see this feature being forgotten about down the road. With that in mind, would it be possible for you to publish a gem that adds this instead? I can then mention this project in the README. This would let me judge how popular this would be.

majioa:
@maurogeorge `Cucumber` is just a tool, that allows your to test a behavour, in some case we may a state represent as a behavour of stasis. So there no principle position what should I test with cucumber, model, controller, or user acitivity. When I test a controller I test its actions at first, when I test a user acitivity I test the behaviour the app in whole, and when I test a model, I validate the behavour of model's, and record's activity. The mean reason to avoid the `rspec` is keep my project with minimal set of gems, include test/development environments. So I just select cucumber, in favor of rspec, but since rspec allows me to write the useful constructions `expect/to`,I just included `rspec-expectations` to me project.

@mcmire well,  the project when I used the construction `cucumber + rspec-expectations + shoulda-matchers` is here: https://github.com/znamenica/bukovina/tree/master/features

maurogeorge:
@majioa Sorry that's sound that I said you are wrong using cucumber to test your model, this was not my intention :innocent:

What I mean is usually I wrote my unit tests using RSpec or Minitest and acceptance directly on the [RSpec (like shoulda-matchers does)](https://github.com/thoughtbot/shoulda-matchers/tree/master/spec/acceptance). In a case when a project already have Cucumber for acceptance tests I probably wrote unit tests using RSpec, or if I want to reduce dependencies use the Minitest.
With the idea of built a [Test Pyramid](http://martinfowler.com/bliki/TestPyramid.html) with more unit tests and less acceptance tests.

I have the same concern as @mcmire

> I guess my fear, really, is that if we merged this, we'd be supporting a use case that wouldn't be utilized that often -- I can easily see this feature being forgotten about down the road. 

I agree with the idea of publish a gem that adds this behaviour.

majioa:
> With the idea of built a Test Pyramid with more unit tests and less acceptance tests.

Of course I'm albe to apply Test Pyramid , Test Cylinder, etc. But that is good idea to use for whole pyramid, cylinder etc, the single tool RSpec, Cucumber, etc. It seems for me strange to use many ones. 

So I prefer `cucumber + rspec-expectations + shoulda-matchers` rather than `cucumber + rspec-expectations + shoulda-matchers + minitest`.

>  I agree with the idea of publish a gem that adds this behaviour.

Which the gem do you mean?

maurogeorge:
@majioa I understand you, in the end is about choices and trade-offs.

About the gem, @mcmire [said](https://github.com/thoughtbot/shoulda-matchers/pull/737#issuecomment-115108135):

> With that in mind, would it be possible for you to publish a gem that adds this instead? I can then mention this project in the README. This would let me judge how popular this would be.

mcmire:
Okay, after taking another look at this, I'll merge this in. It seems that if you have some kind of assertion library loaded (whether Minitest or RSpec), Cucumber [automatically mixes in helpers to let you use that library](https://github.com/cucumber/cucumber-ruby/blob/8c12414e2ba9760bdc9a6e76c6b6b3afd774d552/lib/cucumber/rb_support/rb_language.rb#L123). So it's totally something you're supposed to do with Cucumber -- I didn't know this before!

I'm going to hold off on merging this until 3.1.0 as I want to get 3.0.0 out the door, so I'll keep this open until then.

majioa:
@mcmire  good to hear, thanx =)

majioa:
closed in favor of #910

PR #740, opened 2015-06-25
Use the RSpec 3 syntax on RDocs

maurogeorge:
RSpec 3 docs uses the `is_expected.to` over the `should` syntax. So I think we are good to go and update the docs to reflect the last RSpec version.

### DIFF
--- a/lib/shoulda/matchers/action_controller/callback_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/callback_matcher.rb
@@ -10,8 +10,8 @@ module Shoulda
       #
       #     # RSpec
       #     describe UsersController do
-      #       it { should use_before_filter(:authenticate_user!) }
-      #       it { should_not use_before_filter(:prevent_ssl) }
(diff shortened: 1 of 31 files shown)


mcmire:
While `should` doesn't fit *as well* into the RSpec 3 syntax, [it's still legitimate syntax](https://github.com/rspec/rspec-expectations/blob/master/Should.md#one-liners) and hasn't been deprecated. Plus, it's shorter syntax from the rather wordy `is_expected.to`. So I don't see an immediate need to change this. (And although this isn't a great reason for keeping it around, the gem is still called *should*a-matchers.)

maurogeorge:
@mcmire I usually enforce the expect syntax:

```ruby
  config.expect_with :rspec do |expectations|
    expectations.syntax = :expect
  end
```

I know that even with this, I can use the `should` one line matchers, but I prefer use the `is_expected.to` to mantain a consistence across the project.

Since you don't see an immediate need to change this, I am closing this to avoid noise on the PRs.

Issue #519, opened 2014-05-06
compatible with rspec 3.0.0 ?


zhu1230:
Shoulda::Matchers::ActiveRecord::AssociationMatcher implements a legacy RSpec matcher
       protocol. For the current protocol you should expose the failure messages
       via the `failure_message` and `failure_message_when_negated` methods.

any idea? thanks

mcmire:
@xl1004v Are you still experiencing issues with 2.8.0?

maurogeorge:
@mcmire I think we can close this.

I double checked and we have defined the alias properly. 
@matiasleidemer replied that got no warning on `2.7.0`
The last reply when this was a problem was 5 months ago.

I think we can close for now, and reopen if we got new reports on it.

mcmire:
@maurogeorge Yeah, that sounds like a good plan. Closing.

mcmire:
I actually need to go back through and remove all the aliases since we're technically just targeting RSpec 3 now.

jankeesvw:
I am seeing this error too.

mcmire:
Which RSpec version and which shoulda-matchers version?

maurogeorge:
@mcmire today what is your idea to the `shoulda-matcher` integration with RSpec 2? Is to [drop support](https://github.com/thoughtbot/shoulda-matchers/issues/519#issuecomment-90161242)? Or [keep the support](https://github.com/thoughtbot/shoulda-matchers/issues/519#issuecomment-42686507)?

I vote for keep support like the [clearence did](https://github.com/thoughtbot/clearance/pull/460).

mcmire:
@maurogeorge In 3.0, we've dropped support for RSpec 2, and no longer support it.

maurogeorge:
@mcmire thanks, I missed this on News.

PR #741, opened 2015-06-26
Drop support to RSpec 2

maurogeorge:
Since on 3.0 we [no more](https://github.com/thoughtbot/shoulda-matchers/issues/519#issuecomment-115524811) [support RSpec 2](https://github.com/thoughtbot/shoulda-matchers/blob/master/NEWS.md#backward-incompatible-changes) I think we can remove the alias that is defined to keep RSpec 2 legacy protocol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Shoulda Matchers [![Gem Version][version-badge]][rubygems] [![Build Status][travis-badge]][travis] ![Downloads][downloads-badge]
 
-Shoulda Matchers provides RSpec- and Minitest-compatible one-liners that test
+Shoulda Matchers provides RSpec 3- and Minitest-compatible one-liners that test
 common Rails functionality. These tests would otherwise be much longer, more
 complex, and error-prone.
 
@@ -97,7 +97,7 @@ interested in the README for 2.8.0 instead.][2.8.0-README]**
 
 ## Installation
 
-### RSpec
+### RSpec 3
 
 Include `shoulda-matchers` in your Gemfile:
 
--- a/lib/shoulda/matchers/action_controller/callback_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/callback_matcher.rb
@@ -169,12 +169,10 @@ module Shoulda
         def failure_message
           "Expected that #{controller_class.name} would have :#{method_name} as a #{kind}_#{callback_type}"
         end
-        alias failure_message_for_should failure_message
 
         def failure_message_when_negated
           "Expected that #{controller_class.name} would not have :#{method_name} as a #{kind}_#{callback_type}"
         end
-        alias failure_message_for_should_not failure_message_when_negated
 
         def description
           "have :#{method_name} as a #{kind}_#{callback_type}"
--- a/lib/shoulda/matchers/action_controller/filter_param_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/filter_param_matcher.rb
@@ -38,12 +38,10 @@ module Shoulda
         def failure_message
           "Expected #{@key} to be filtered; filtered keys: #{filtered_keys.join(', ')}"
         end
-        alias failure_message_for_should failure_message
 
         def failure_message_when_negated
           "Did not expect #{@key} to be filtered"
         end
-        alias failure_message_for_should_not failure_message_when_negated
 
         def description
           "filter #{@key}"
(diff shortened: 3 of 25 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Shoulda Matchers [![Gem Version][version-badge]][rubygems] [![Build Status][travis-badge]][travis] ![Downloads][downloads-badge]
 
-Shoulda Matchers provides RSpec- and Minitest-compatible one-liners that test
+Shoulda Matchers provides RSpec 3- and Minitest-compatible one-liners that test
```
mcmire: You can drop the `-` here.

mcmire:
Okay, but I've also announced:

> This doesn't mean that the gem will stop working immediately

I think my intention was to drop support now, give people time to switch over completely, and then completely drop compatibility at some point in the future.

I suppose that it seems inconsistent that 3.0 drops compatibility for Rails 3, except that Rails 5 is around the corner. Do you think that more people use RSpec 2 vs. Rails 3?

maurogeorge:
My concern is that we are following Semantic Versioning so I think this is the point to we break things. 

> This doesn't mean that the gem will stop working immediately, but we won't accept any pull requests to fix any compatibility issues, so you're encouraged to upgrade.

I imagine a case when shoulda-matchers 3.0 works on RSpec 2 and Rails 3 but shoulda-matchers 3.4 for example does not work, and since we are not accepting PRs to fix this, maybe sound like a problem.

[MAJOR is when we make incompatible API changes (drop support to RSpec 2 and Rails 3)](http://semver.org/), as a user I expect broken things on 3.0 and not on 3.4 since we are following Semantic Version.

My concern is to 3.0 and so on gave the false feeling of support and users start got some strange errors, if we are dropping the support I think it is better to make this explicit.

Makes sense to you?

mcmire:
Okay, let me think through this. So far in order for someone to start using 3.0 right away, they will have to:

* Upgrade to Ruby 2.0.0
* Upgrade to Rails 4
* Add a configuration block to their `rails_helper`
* Potentially fix any tests that involve `should_not allow_value`
* Potentially fix any tests that involve `validate_uniqueness_of`
* Use `validate_length_of` etc. instead of `ensure_length_of`
* Rename `set_the_flash` to `set_flash`
* Potentially fix any tests that involve `set_flash` or `set_session`
* Upgrade to RSpec 3?

I guess since upgrading the Rails version is a pretty big step, we probably won't see people upgrading right away. Instead, new downloads will probably come from people using the gem on a fresh project. In that case, they'd probably be using RSpec 3 and not 2.

So... alright. This is fine. If we're doing this, should we also drop support for Minitest 4?

To your point about dispelling confusion around what we do and don't support, this makes me think that we need to add a Compatibility section to the README:

> Shoulda Matchers is tested and supported against Rails 4.x, RSpec 3.x, Minitest 5, and Ruby 2.x.

maurogeorge:
I think the following has reached ended of support:

- [Rails 3.2 is not supported by the core team](http://guides.rubyonrails.org/maintenance_policy.html)
- [Ruby version 1.9.3 ended life](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/)
- [I don't find any docs saying that RSpec 2 do not receive any maintenance, but the last 2.X version was 2.99.0 - June 1, 2014](https://rubygems.org/gems/rspec/versions)
- [I don't find any docs saying that Minitest 4 do not receive any maintenance, but the last 4.X version was  4.7.5 - June 22, 2013](https://rubygems.org/gems/minitest/versions)

If I am not wrong, Rails dropped the support to Minitest 4 too, [ActiveSupport has dependency of Minitest 5](https://rubygems.org/gems/activesupport/versions/4.2.3).

So yes I think we can drop minitest 4. I see [you already have a branch that does that](https://github.com/thoughtbot/shoulda-matchers/compare/ew-remove-minitest-4).

I agree with you the Compatibility section to the README makes lot of sense.
Just to check, today we test Rails and Ruby using travis/appraisals. Our specs are using RSpec 3 and Minitest 5 is tested via acceptance tests. We cover by tests, all lib and versions that we support, right?

Makes sense add a Maintenance policy like rails  does? I mean, I think after the release of the 3.0, we will have only 3.x versions right? Not a 2.x even for bugfixes right? So I think this can be explicit.

mcmire:
Okay, this is merged as 32c0e62596b87e37a301f87bbe21cfcc77750552.

I added a Compatibility section in 6f5ee70a2c4e01ae06841dfa2b733b0ecf091295.

maurogeorge:
Thanks for the merge.

mcmire:
@maurogeorge I just opened a PR for dropping Minitest 4, what do you think? #782

PR #746, opened 2015-07-01
Require DoubleSpeak from DelegateMethodMatcher

jc00ke:
This makes it possible to require 'shoulda/matchers/independent' from
within a spec_helper file without 1) requiring all of Shoulda and 2)
also requiring DoubleSpeak, which is a dependency that should already
be handled.

### DIFF
--- a/lib/shoulda/matchers/independent/delegate_method_matcher.rb
+++ b/lib/shoulda/matchers/independent/delegate_method_matcher.rb
@@ -1,3 +1,5 @@
+require 'shoulda/matchers/doublespeak'
+
 module Shoulda
   module Matchers
     module Independent


mcmire:
I'm not sure why this is failing off the bat. It seems like a Travis-specific error or something...

mcmire:
I just attempted to write a test for this -- how are you using Shoulda::Matchers::Independent? Do you have something like this in your spec_helper?

``` ruby
require 'shoulda/matchers/independent'

RSpec.configure do |config|
  config.include(Shoulda::Matchers::Independent)
end
```

jc00ke:
Honestly, I don't remember, and I can't seem to find my work (on an old laptop). Sorry I'm not of more help at the moment.

mcmire:
Okay, no worries. I'll add a test for it anyway.

Issue #749, opened 2015-07-03
Support for allow_nil option to delegate_method matcher.


alejandrogutierrez:
Hi everyone! I would like to add support for allow_nil option to delegate_method matcher.
```ruby
it { should delegate_method(:deliver).to(:post_office).allow_nil(true) }
```
What do you think?

maurogeorge:
Hi @alejandrogutierrez I like the idea :+1: 

Just a note, I think it is better to not pass a param to `allow_nil`, but something like:

```ruby
it { should delegate_method(:deliver).to(:post_office).allow_nil }
```

Take a look [here](https://github.com/thoughtbot/shoulda-matchers/pull/722) where this was discussed.

I guess I never worked on this matcher code, but as you can see [here](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/independent/delegate_method_matcher.rb) the matcher works with different kinds of delegation([Fowardable](http://ruby-doc.org/stdlib-2.0.0/libdoc/forwardable/rdoc/Forwardable.html), [delegate](http://api.rubyonrails.org/classes/Module.html#method-i-delegate) and methods) so I think we will need to handle this across all delegation and not only on the `delegate` from Rails. Makes sense to you?

Could you please send a PR?

alejandrogutierrez:
Yes, it makes sense!
I will work on this and will send a PR as soon as possible.
Thanks!

mcmire:
Closing in favor of #751.

PR #751, opened 2015-07-03
Adding support to allow_nil option for delegate_method matcher

alejandrodevs:
Hi everyone! I would like to add support for allow_nil option to delegate_method matcher.

``` ruby
it { should delegate_method(:deliver).to(:post_office).allow_nil }
```

I update rspec-rails version in Appraisals file in order to get some new matchers.


alejandrogutierrez:
PR was updated! :dancer:

spec/unit_spec_helper.rb
```diff
@@ -55,6 +56,7 @@
   UnitTests::ColumnTypeHelpers.configure_example_group(config)
 
   config.include UnitTests::Matchers
+  config.include RSpec::Matchers::FailMatchers
```
maurogeorge: I didn't know about this [matchers](http://www.rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers/FailMatchers) :star: 

I think we can drop our [custom](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/matchers/fail_with_message_matcher.rb) [matchers](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/matchers/fail_with_message_including_matcher.rb) and use this provided by RSpec at some point.

alejandrogutierrez: Yes, I think it would be better. :+1:

Appraisals
```diff
@@ -1,7 +1,7 @@
 ruby_version = Gem::Version.new(RUBY_VERSION + '')
 
 shared_dependencies = proc do
-  gem 'rspec-rails', '>= 3.2.0', '< 4'
+  gem 'rspec-rails', '>= 3.3.0', '< 4'
```
mcmire: Mind rebasing and extracting this change as well as the changes to Gemfile and Gemfile.lock to a separate commit?

lib/shoulda/matchers/independent/delegate_method_matcher.rb
```diff
@@ -327,8 +338,21 @@ def ensure_delegate_object_has_been_specified!
           end
         end
 
+        def subject_delegates_to_nil_delegate_object_correctly?
+          return true unless delegate_object_is_nil_allowed
```
mcmire: Instead of using the modifier form of `unless`, which could be hard to recognize while reading this code, what about wrapping this entire method in `if delegate_object_is_nil_allowed`?

Appraisals
alejandrogutierrez: Sure, can you close this PR in favor of #798?

lib/shoulda/matchers/independent/delegate_method_matcher.rb
alejandrogutierrez: I need to return true when `delegate_object_is_nil_allowed ` was not set. If I do what you say I would have to add an else block returning true, it would look like:
```ruby
def subject_delegates_to_nil_delegate_object_correctly?
  if delegate_object_is_nil_allowed
    # My code here...
  else
    true
  end
end
```

I think it looks ugly, what do you think?

mcmire: I don't think so. I think it's clear to see that there is a conditional in this method.

alejandrogutierrez: Ok, I'll update this (y)

alejandrogutierrez:
Done! @mcmire PR updated! :tada: :smile:

chamnap:
when this PR could be merged :)?

maurogeorge:
@chamnap 

From [README](https://github.com/thoughtbot/shoulda-matchers#shoulda-matchers---)
> NOTE: Shoulda Matchers is currently undergoing a changing of the guard. We will not be as responsive to answer questions or issues while we make this transition. Sorry about that.

chamnap:
Thanks :)

chrishough:
Is this merge still in progress? I was looking for the same issue as https://github.com/thoughtbot/shoulda-matchers/issues/749

jamesgoodhouse:
I'd also like to know the status of this PR.

alejandrodevs:
This PR is ready to be merged! :)

lucascaton:
Any news on this? 😄

mcmire:
@alejandrodevs Sorry it took so long for me to respond to this. After cleaning up your PR a bit, I've merged it here: d49cfcae1b294e12a05e06a5612cb8ebb22a7df1. Thank you for your submission!

chrishough:
Thank you @mcmire

alejandrodevs:
Thank you @mcmire

PR #752, opened 2015-07-04
Fix `now` qualifier on `set_flash` when is called after the key and expected message is set

### LINKED ISSUE #750: set_flash[:foo].now matcher does not care if flash[:foo] really set .now
I want to assert that a `flash.now[:foo]` uses `.now`, but the test passes, even if `.now` is not set.

Code:
```Ruby
# controller.rb, simplified, authorisation, authentication etc left out
def create
  @project = Project.new(project_params)
  if @project.save
    flash[:success] = 'Project successfully created'
    redirect_to projects_path
  else
    flash[:alert] = 'Project could not be created.'
    render 'new'
  end
end

# spec
it 'sets a flash.now' do
  post :create, invalid_params
  is_expected.to set_flash[:alert].to('Project could not be created.').now
  # passes!
end
```

The reverse is not true, btw. If `flash.now[:foo]` is set in the controller but not in the spec, the test fails.

Rails 4.2.3
Ruby 2.2.2
shoulda_matchers 2.8.0

maurogeorge:
Closes  #750

When the `now` qualifier is called after the `[]` and `to`, example:

```
set_flash['to for flash.now'].to('value for flash.now').now
```

Is istantiated a new matcher(SetSessionOrFlashMatcher different from the one that is instantiated on the initialize) that does not know about the `[]` and `to` value that was set before the `now` is called.

This fixes this behavior passing the params explicitly to the matcher SetSessionOrFlashMatcher instantiated on `now`.

### DIFF
--- a/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
@@ -175,6 +175,9 @@ module Shoulda
         def now
           store = FlashStore.now
           @underlying_matcher = SetSessionOrFlashMatcher.new(store)
+          if key
+            @underlying_matcher[key]
+          end
           self
         end
 
@@ -184,6 +187,7 @@ module Shoulda
         end
 
         def [](key)
+          @key = key
           underlying_matcher[key]
           self
         end
@@ -195,7 +199,7 @@ module Shoulda
 
         protected
 
-        attr_reader :underlying_matcher
+        attr_reader :underlying_matcher, :key
       end
     end
   end
--- a/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
@@ -175,9 +175,8 @@ module Shoulda
         def now
           store = FlashStore.now
           @underlying_matcher = SetSessionOrFlashMatcher.new(store)
-          if key
-            @underlying_matcher[key]
-          end
+          @underlying_matcher[key]
+          @underlying_matcher.to(expected_value)
           self
         end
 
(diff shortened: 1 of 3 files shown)


mcmire:
I'd like to go by the same semantics that the actual flash hash uses here.

If you say these two things in a controller, do they do the same thing?

```
flash.now["foo"]
flash["foo"].now
```

If they don't do the same thing, should we even allow this behavior?

maurogeorge:
@mcmire in a controller we can do only the `flash.now["foo"]` the `flash["foo"].now` does not work, a `NoMethodError` is raised.

I don't have a strong opinion about `set_flash[:foo].now` or `set_flash.now[:foo]`, personally I prefer the first one, but only because I prefer define the key first and later all the qualifiers. But I am ok, If we go to the `set_flash.new[:foo]` only.

If we go to the `set_flash.new[:foo]` I think we need to:

- [ ] Update the news telling this is a backward-incompatible changes
- [ ] Update the docs to reflect the new only accepted syntax
- [ ] Raises a error, when this qualifier is called after the `[]`? Because people updating can be using this qualifier after the `[]`, so this will generate a false positive like the described here https://github.com/thoughtbot/shoulda-matchers/issues/750

Since change this matcher signature is a backward-incompatible changes I think we need to change this before the 3.0.0. What you think?

mcmire:
Those three tasks seem like a good plan.

And this would definitely go in 3.0. `set_flash` is already backward-incompatible, anyway (as previously the method was called `set_the_flash`), so I'm fine with it.

maurogeorge:
@mcmire I make A PR with the changes https://github.com/thoughtbot/shoulda-matchers/pull/755

Closing this in favor of the https://github.com/thoughtbot/shoulda-matchers/pull/755

PR #755, opened 2015-07-08
set_flash.now works only if called before any other qualifier

maurogeorge:
In the 3.0.0 `set_flash[:foo].now` is no longer a valid syntax, only the
`set_flash.now[:foo]` is valid [1].

Since this can create a false positive when people updating, we will raise a
exception to make the things more explicit.

[1]: https://github.com/thoughtbot/shoulda-matchers/pull/752


mcmire:
Merged in 9268b4940fb71718654376de5de7375c11c4b051. Thanks :ice_cream:

maurogeorge:
Thanks for the merge :star2:

Issue #743, opened 2015-06-29
Rails 4.2.2 compatibility


voxik:
Testing against RoR 4.2.2, I observe following test errors:

```
  1) Shoulda::Matchers::ActiveModel::DisallowValueMatcher when the value is outside of the range of the column not qualified with strict accepts, failing with the correct message
     Failure/Error: expect(&assertion).to fail_with_message(message)
       Expectation should have failed with message:
         Did not expect errors when attr is set to 100000,
         got RangeError: "100000 is out of range for ActiveRecord::Type::Integer with limit 2"
       However, the expectation did not fail.
     # ./spec/unit/shoulda/matchers/active_model/disallow_value_matcher_spec.rb:93:in `block (4 levels) in <top (required)>'

... snip ...

  9) Shoulda::Matchers::ActiveModel::AllowValueMatcher when the value is outside of the range of the column qualified with strict qualified with a message ignores any specified message
     Failure/Error: expect(&assertion).to fail_with_message(message)
       Expectation should have failed with message:
         Did not expect exception to include "some message" when attr is set to 100000,
         got RangeError: "100000 is out of range for ActiveRecord::Type::Integer with limit 2"
       However, the expectation did not fail.
     # ./spec/unit/shoulda/matchers/active_model/allow_value_matcher_spec.rb:297:in `block (5 levels) in <top (required)>'
```

It seems that AR changed again how the ranges are handled: https://github.com/rails/rails/commit/fed7d7cd7c6e71c1b78a4a2fda7d069d4aa6f468

voxik:
Please note that I am testing rspec-matchers 2.8.0 should it make any difference.

mcmire:
Okay, thanks. I'll throw a fix into the 3.0 release.

mcmire:
Closing this in favor of #762.

PR #762, opened 2015-07-23
Stop handling AR RangeError when assign a attribute

maurogeorge:
Active Record 4.2 and upper no more raises a `RangeError` on the assign, only at
database level [1].

This will remove all code related to handling of Active Record `RangeError`.

Users of Rails 4.2.0 will receive a error because Rails changed this behavior on
4.2.1 and upper. Since the migration to 4.2.0 to 4.2.X is a patch Z, that
includes no API changes, the user can update your Rails version to avoid these
errors [2].

[1]: https://github.com/rails/rails/commit/fed7d7cd7c6e71c1b78a4a2fda7d069d4aa6f468
[2]: http://guides.rubyonrails.org/maintenance_policy.html

Closes https://github.com/thoughtbot/shoulda-matchers/issues/743


mcmire:
Merged in 73cf275d104f62b32eb80b4e72577bd56ebdd730.

maurogeorge:
Thanks :100:

PR #772, opened 2015-08-18
Prefix partial name with underscore

AlexVPopov:
If a partial's is not prefixed with an underscore a test would fail with a similar error message, as:

> Failure/Error: it { should render_template(partial: 'form') }
       expecting partial \<form> but action rendered <["expenses/_form", "_form"]>.
       Expected {"expenses/_form"=>1, "_form"=>1} to include "form".

### DIFF
--- a/lib/shoulda/matchers/action_controller/render_template_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/render_template_matcher.rb
@@ -20,7 +20,7 @@ module Shoulda
       #         before { get :show }
       #
       #         it { should render_template('show') }
-      #         it { should render_template(partial: 'sidebar') }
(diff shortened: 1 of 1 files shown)


maurogeorge:
@AlexVPopov thanks for the PR.

LGTM. Following the [specs](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/action_controller/render_template_matcher_spec.rb#L26-L68) we really need to pass the `_partial` to the matcher.

mcmire:
Sweet, that makes sense to me. Thanks for the doc fix. Merged as 8868bda79af031bdbf6ec6a9509ab66bb4a1d13f.

PR #773, opened 2015-08-18
Use strings for flash keys

AlexVPopov:
If symbol keys are used with set_flash, a spec would fail with:

> Expected FooController to set flash[:bar] to "baz", but it did not

### DIFF
--- a/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/set_flash_matcher.rb
@@ -62,8 +62,8 @@ module Shoulda
       #       describe 'GET #index' do
       #         before { get :index }
       #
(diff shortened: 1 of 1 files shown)


maurogeorge:
Hi @AlexVPopov thanks for your PR.

I think this PR try to solve the same problem of the https://github.com/thoughtbot/shoulda-matchers/pull/732 I am right? I think the expected behavior is that we can send string and symbols to the matcher

AlexVPopov:
Yes, definitely. Sorry, I didn't see there was a PR for this. If you want to - close this one. Cheers.

maurogeorge:
@AlexVPopov no problem :smile: 

Let's wait @mcmire to close this.

PR #777, opened 2015-09-08
list different examples for rspec should and expect style

timoschilling:


mcmire:
Actually, the `should ...` syntax still works regardless of which style you're using: https://github.com/rspec/rspec-expectations/issues/227

Do you mind if I add a section like the following to the README instead?

> Note that in this README and throughout the documentation we're using the `should` form of RSpec's one-liner syntax over `is_expected.to`. The `should` form works regardless of how you've configured RSpec -- meaning you can still use it even when using the `expect` syntax. But if you prefer to use `is_expected.to`, you can do that too:

> ``` ruby
describe Person do
  it { is_expected.to validate_presence_of(:name) }
end
```

timoschilling:
:+1:

Issue #776, opened 2015-09-04
validate_uniqueness_of matcher failed for integer


JuanitoFatas:
Hello!

I have a model,

```ruby
class User
  # github_id is an integer.
  validates_uniqueness_of :github_id, allow_blank: true
end
```

In `spec/models/user_spec`:

```ruby
describe User do
  context "validations" do
    it { should validate_uniqueness_of(:github_id).allow_blank }
  end
end
```

This test failed because from [active_record/validate_uniqueness_of_matcher.rb#L298-L300](https://github.com/thoughtbot/shoulda-matchers/blob/v2.8.0/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L298-L300):

```ruby
def existing_record_is_blank?
  @existing_record.present? && existing_value.strip == ''
end
```

it is calling `strip` on the integer, and integer does not have `strip` method:

```
Failure/Error: it { should validate_uniqueness_of(:github_id).allow_blank }
NoMethodError:
  undefined method `strip' for 0:Fixnum
```

Should this `#existing_record_is_blank?` check `existing_value` by Rails's `blank?` method?

Thanks. I'm using shoulda-matchers v2.8.0.

mcmire:
Yeah, I think that definitely makes sense.

mcmire:
Closing this issue in favor of #781.

Issue #760, opened 2015-07-14
Tighten CouldNotSetAttributeError restriction on allow_value, but provide a way to get around it


mcmire:
Right now we restrict `allow_value` such that if you're trying to set the attribute to a non-nil value, but it ends up being nil, *or* you're trying to set the attribute to nil but it ends up being non-nil, then you'll get a `CouldNotSetAttributeError`. We need to tighten this restriction so that if the before and after values are *completely* different then you get this error. (I got just bitten by a case in which I was trying to validate numericality of an attribute that strips non-numbers anyway, thereby making the validation useless. But I didn't make the connection at first.)

**However**, we also need to recognize that this logic will sometimes fail, and people need a way to tell the gem, "look, I get it, I know that this attribute is going to change the value I give it and do something else, and that's okay". Probably this looks like another qualifier on `allow_value`, although I'm not sure what to call the qualifier yet. But it could look something like this:

``` ruby
it do
  should allow_value("foo").for(:bar).ignoring_value_changes
end
```

mcmire:
This is now done in 9d9dc4e6b9cf2c19df66a1b4ba432ad8d3e5dded.

The qualifier we're adding is called `ignoring_interference_by_writer`:

``` ruby
it { should_not allow_value([]).for(:attr).ignoring_interference_by_writer }
```

PR #781, opened 2015-09-21
Use ActiveSupport Object#blank? for checking record existing value

JuanitoFatas:
For #776.

I want to write a spec in `spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb`, but I don't really understand how to write one. Any advices?

Thanks!

--

Note for myself

Run a single spec: `bundle exec appraisal 4.2 rspec spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb`

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -346,7 +346,7 @@ module Shoulda
         end
 
         def existing_record_is_blank?
-          @existing_record.present? && existing_value.strip == ''
+          @existing_record.present? && existing_value.blank?
         end
 
         def create_record_in_database(options = {})
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -163,6 +163,17 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
       end
     end
 
+    context 'when the attribute is an integer' do
+      it 'accepts' do
+        record = build_record_validating_uniqueness(
+          attribute_type: :integer,
+          validation_options: { allow_blank: true }
+        )
+
+        expect(record).to validate_uniqueness.allow_blank
+      end
+    end
+
     context 'when the record is created beforehand' do
       context 'when the subject is a new record' do
         it 'accepts' do


JuanitoFatas:
screenshot 2015-09-22 00 06 30

:point_down: change to single quote :point_down: 

screenshot 2015-09-22 00 06 35

:confused: :grey_question:

lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -346,7 +346,8 @@ def existing_record_is_nil?
         end
 
         def existing_record_is_blank?
-          @existing_record.present? && existing_value.strip == ''
+          require 'active_support/core_ext/object/blank'
```
mcmire: Can you put this at the top of the file? I think that might be a better place for this.

Also, this should be single quotes, ignore Hound -- I'm not sure what's up, I'll have to take a closer look at that.

mcmire:
In order to figure out which spec to write, why do you feel like you need to require ActiveSupport here? What problem is this solving?

JuanitoFatas:
> why do you feel like you need to require ActiveSupport here?

Because we want to use the `blank?` method to check if existing value is blank or not.

> What problem is this solving?

Checking for integer won't fail because it does not have a `strip` method.

lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
JuanitoFatas: Thanks :+1: Changed require at top.

mcmire:
@JuanitoFatas Oh, right, I didn't catch you're fixing the issue you posted earlier, my bad.

If you look in [validate_uniqueness_of_spec](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb), you'll notice that there are lots and lots of contexts. This may seem daunting but the one we want to look for is "when the model has a uniqueness validation". So somewhere inside of this context, we need to add another context for "when the attribute is an integer". The best place for this looks like right after the first context, "when the attribute has a character limit" (which is assuming that the attribute is a string). 

As far as the test itself goes, we have a `build_record_validating_uniqueness` helper method that essentially makes a dummy model with some kind of attribute and gives that attribute a uniqueness validation. You can specify what the type of that attribute is by passing an `attribute_type` to `build_record_validating_uniqueness`. So you build a record, and then in order to test that the uniqueness validation runs against that record and passes, you can simply say `expect(record).to validate_uniqueness` (where `validate_uniqueness` is another method that wraps the `validate_uniqueness_of` matcher).

mcmire:
By the way, you shouldn't have to require ActiveSupport explicitly. Since `validate_uniqueness_of` is an ActiveRecord matcher, I think we can already assume that ActiveSupport has been fully loaded.

JuanitoFatas:
> @JuanitoFatas Oh, right, I didn't catch you're fixing the issue you posted earlier, my bad.

No worries :grinning: 

Thanks for the nice heads up on how to write a spec for this.

> By the way, you shouldn't have to require ActiveSupport explicitly. Since validate_uniqueness_of is an ActiveRecord matcher, I think we can already assume that ActiveSupport has been fully loaded.

:+1: Removed

--

The spec is still failing though :sweat:...

```
$ BUNDLE_GEMFILE=/Users/Juan/dev/shoulda-matchers/gemfiles/4.2.gemfile bundle exec rspec spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb:167

Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher
  when the model has a uniqueness validation
    when the attribute is an integer
      accepts (FAILED - 1)

Failures:

  1) Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher when the model has a uniqueness validation when the attribute is an integer accepts
     Failure/Error: expect(record).to validate_uniqueness.allow_blank
       Expected Class to be able to set attr to "", but got nil instead.
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:330:in `ensure_that_attribute_has_been_changed_to_or_from_nil!'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:313:in `set_attribute_ignoring_range_errors'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:307:in `set_attribute'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:302:in `value_matches?'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:264:in `block in matches?'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:264:in `each'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:264:in `all?'
     # ./lib/shoulda/matchers/active_model/allow_value_matcher.rb:264:in `matches?'
     # ./lib/shoulda/matchers/active_model/validation_matcher.rb:40:in `allows_value_of'
     # ./lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:318:in `allows_blank?'
     # ./lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:268:in `matches?'
     # ./spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb:173:in `block (4 levels) in <top (required)>'

Finished in 0.05759 seconds (files took 7.23 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb:167 # Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher when the model has a uniqueness validation when the attribute is an integer accepts

Randomized with seed 10504
```

My [setup & exercise](https://github.com/JuanitoFatas/shoulda-matchers/blob/patch/validate_uniqueness_of-handle-integer/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb#L168-L173) is not correct? :confused: Or we found a new bug? :heart_eyes:

mcmire:
Ah, you're trying to use the matcher with `allow_blank`, I see.

So you're getting a CouldNotSetAttributeError because when using `allow_blank`, the matcher will try to set the attribute in question to an empty string. This doesn't really make sense on an integer attribute since it doesn't allow every blank value -- it will implicitly convert an empty string to nil. So the matcher is confused because the value that the attribute was set to isn't really the value it is set to (`""` was changed to `nil`).

So I would argue that kind of test isn't necessary and we should probably have an error message that says so. I'm working on [a way to get around this](https://github.com/thoughtbot/shoulda-matchers/issues/760) in case you *really* want to write a test anyway.

PR #782, opened 2015-09-22
Drop support for Minitest 4

mcmire:
Why:

* The latest patch version of Minitest 4.x was released in June 2013.
  I think we can make a safe assumption that most people are using
  Minitest 5.x now.
* There's a precedent for upgrading, as Rails did so from Minitest 4 to
  5 in 4.1.

To satisfy the above:

* Rename the Minitest5 adapter to just Minitest.
* Stop running integration tests against Minitest 4.


maurogeorge:
LGTM let's wait for the Travis and :shipit:

mcmire:
We can't do this, actually. We still support Rails 4.0.0, and it depends on Minitest 4, so we have to continue to support Minitest 4, too.

maurogeorge:
Okay makes sense.

Issue #787, opened 2015-10-04
Case-sensitive uniqueness validation fails if default value lacks alpha characters


tovodeverett:
I've run into a fairly unique situation, but I thought I'd bring it up in case someone wants to defend against it.

If your factory for generating objects uses a string that does not have change with `swapcase` (i.e. something like `'%-1-%'`), and you're doing a `validate_uniqueness_of` test without specifying `case_insensitive`, then you will get a rather confusing error message.

Perhaps it might make sense somewhere around https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L410 to check to see that the swapcased value differs from the original value and if it doesn't to throw an error along the lines of, "if you're not doing a case_insensitive test, then you need to make sure your value here has alpha characters in it!".

mcmire:
Moving discussion to #789.

PR #789, opened 2015-10-04
Fix checking case sensitivity when value is case insensitive

albertyw:
I ran into the same issue as https://github.com/thoughtbot/shoulda-matchers/issues/787 and fixed it here.

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -450,6 +450,18 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
 
         expect(record).to validate_uniqueness
       end
+
+      context 'when the value has no case sensitivity' do
+        it 'accepts', focus: true do
+          model = define_model_validating_uniqueness(
+            attribute_type: :string,
+            validation_options: { case_sensitive: true },
+          )
+          record = create_record_from(model, attribute_name => '123')
+          error = Shoulda::Matchers::ActiveModel::CouldNotValidateCaseError
+          expect { validate_uniqueness.matches?(record) }.to raise_error error
+        end
+      end
     end
 
     context 'when qualified with case_insensitive' do
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -452,7 +452,7 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
       end
     end
 
-    context 'when case_insensitive is specified' do
+    context 'when qualified with case_insensitive' do
       it 'rejects' do
         record = build_record_validating_uniqueness(
           attribute_type: :string,
--- a/lib/shoulda/matchers/active_model/errors.rb
+++ b/lib/shoulda/matchers/active_model/errors.rb
@@ -47,6 +47,46 @@ Please use a #{record_name} with an empty `password` instead.
           model_name.humanize.downcase
         end
       end
+
+      class CouldNotValidateCaseError < Shoulda::Matchers::Error
+        def self.create(record, attribute)
+          super(record: record, attribute: attribute)
+        end
+
+        attr_accessor :record, :attribute
+
+        def message
+          <<-EOT.strip
+Your #{model} model has a uniqueness validation on :#{attribute} which is
+declared to be case-sensitive, but the value the uniqueness matcher used,
+"#{value}", doesn't contain any alpha characters, so using it to test the
+case-sensitivity part of the validation is ineffective. There are two possible
+solutions for this depending on what you're trying to do here:
+
+If you meant for the validation to be case-sensitive,
+then you need to give the uniqueness matcher a saved instance of #{model} with a
+value for :#{attribute} that contains alpha characters.
+
+If you meant for the validation to be case-insensitive,
+then you need to add `case_sensitive: false` to the validation and add
+case_insensitive to the matcher.
+
+For full examples, please see:
+
+http://github.com/thoughtbot/shoulda-matchers/wiki/error-uniques
+          EOT
+        end
+
+        private
+
+        def model
+          record.class
+        end
+
+        def value
+          record.__send__(@attribute)
+        end
+      end
     end
   end
 end
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -412,6 +412,12 @@ module Shoulda
             if @options[:case_insensitive]
               disallows_value_of(swapcased_value, @expected_message)
             else
+              if value == swapcased_value
+                raise ActiveModel::CouldNotValidateCaseError.create(
+                  existing_record,
+                  @attribute,
+                )
+              end
               allows_value_of(swapcased_value, @expected_message)
             end
           else


tovodeverett:
What if there are constraints on the value that make appending "_case" illegal?

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -450,9 +450,19 @@
 
         expect(record).to validate_uniqueness
       end
+      context 'when the value has no case sensitivity' do
```
mcmire: Maybe "when the value is the same as its swapcased version"?

Also add a line break above this line, per style.

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -450,9 +450,19 @@
...
+        it 'accepts' do
+          model = define_model_validating_uniqueness(
+            attribute_type: :string,
+            validation_options: { case_sensitive: true },
+          )
+          record = create_record_from(model, attribute_name => '123')
```
mcmire: I think `attribute_name` should be `value`, right?

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -450,9 +450,19 @@
...
+          expect(record).to validate_uniqueness
+        end
+      end
     end
 
-    context 'when case_insensitive is specified' do
+    context 'when case_sensitive is specified' do
```
mcmire: This line shouldn't be changed, "specified" is referring to the matcher. (Really this should say "when qualified with `case_insensitive`", but I'll fix it later.)

mcmire:
The current behavior is to test that the record either allows or disallows the swapcased version of the value whether `case_insensitive` is specified or not. I just read that issue and I think what we should do is not even bother making this assertion if the value is the same as its swapcased version instead of adding alpha characters onto the value (or raising an error as the aforementioned issue suggests).

So what if you changed the `validate_case_sensitivity?` method to look like this?

``` ruby
if value.respond_to?(:swapcase) && value != value.swapcase
  # allow or disallow
else
  true
end
```

tovodeverett:
From a testing standpoint, the default behavior is to verify that the uniqueness is indeed case sensitive (or to test that the uniqueness is indeed case insensitive if we declare `case_insensitive`).  If I make a mistake and supply a model that has a value that is swapcase immutable, then we're no longer testing either of those, but as a user I wouldn't necessarily be aware of that change since it was made silently.

Perhaps we need an option for `ignore_case_sensitivity`.  For instance, if I have a string where I store formatted phone numbers.  I might have validation rules that make it impossible to store alpha characters.  On the other hand, I might have a field where I want to test that it is case insensitive, but I was oblivious and set the value to '123'.  It would be nice in the latter case to be informed, "Hey, if you want the test to validate case sensitivity, you had better pass a value that lets us test that!"

mcmire:
@tovodeverett Okay, that's fair. I don't know if we need an option for it because then you'd have to remember to use that too. But that's a good warning to have.

I'm trying to understand your original use case... in your test, did you pre-create a record before using the matcher?

albertyw:
In my tests, I was not pre-creating the object (shoulda-matchers does it internally though).  

I think @tovodeverett's suggestion is best.  The current status will fail with a non-useful error and there's no way modifying the value will work because of possible additional validations.  Instead of defaulting to `true` if the value isn't case sensitive (which therefore doesn't test anything), perhaps a RuntimeError can be raised so that an extra option doesn't need to be added.

mcmire:
Okay, so you're saying that if you as the developer accidentally use a value that wouldn't create a good test by being un-swapcase-able, then you'd want the test to fail so that you're forced to replace it with something better?

albertyw:
Well, I guess to expand a bit, if there are validations to preclude changing the value to something swapcase-able (e.g. a phone number), then the test should be changed to be case insensitive.

tovodeverett:
In my case, the underlying value is case-sensitively unique.  It's a field where I store pattern matches (SQL patterns for LIKE).  I'm using factory_girl, and the pattern was along the lines of `'%-1-%'`, which is, of course, unchanging with swapcase!  I realized that since my underlying implementation was case-sensitively unique, I didn't want to indicate to the test that it should be case insensitive (since that didn't match the implementation).  So I changed the value in factory_girl to `'%-1[-a]%'` and thus ensured that I got a test that it was indeed case sensitive.

I'm 95% onboard with albertyw's suggestion - if the stored value cannot contain alpha characters, then suggest to the developer that they add `case_insensitive` to the test.  Basically, the error message would guide the developer into one of two resolutions: Either supply a test value that mutates with swapcase, or explicitly make the test `case_insensitive`.

mcmire:
Okay, just so I understand, it seems like there are two uses cases here:

1. The developer has a case-sensitive validation in the model and is using the matcher without `case_insensitive`, but the value the matcher is using isn't `swapcase`-able. To fix this, the developer should provide a record with a value that is `swapcase`-able.
2. The developer has a case-sensitive validation on an attribute that holds values which do not contain any alpha characters. The value the matcher is using also isn't `swapcase`-able, but changing the value won't work -- in this case the developer needs to change the validation (and the matcher) to be case-insensitive.

So as you say @tovodeverett we just need to explain to the developer why is this happening and guide them into one of these choices.

tovodeverett:
Yup!  Technically case 2 is a simplification, but I think it's good enough.

Technically, the attribute in question isn't really case-sensitive or case-insensitive - the concept of case sensitivity doesn't apply to an attribute which can't store alpha characters, but since calling it case-insensitive will work, I think that's a good solution (and definitely simpler than defining yet another concept).  Don't get me started on I18N, [Turkish i's](http://blog.codinghorror.com/whats-wrong-with-turkey/), and collations!

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
albertyw: I thought that too, but changing that line to `record = create_record_from(model, value: '123')` makes the test fail.

mcmire: Oh... gah... I was skimming over the code and wasn't looking closely. The key is the name of an attribute, so this is right.

albertyw:
@tovodeverett @mcmire I've updated the pull request based on your comments.

mcmire:
Cool, thanks so much. I'll take a look at this tonight.

mcmire:
@albertyw @tovodeverett Okay, I reworded the error message to explain this in more detail. What do you think about this? (the model name, attribute, and value will be filled in dynamically)

> Your Example model has a uniqueness validation on :attr which is declared to be case-sensitive, but the value the uniqueness matcher used, "123", doesn't contain any alpha characters, so the matcher could not test the case-sensitivity part of the validation effectively. There are two possible solutions for this depending on what you're trying to do here:
>
> 1. If you meant for the validation to be case-sensitive, then you need to give the uniqueness matcher a saved instance of Example with a value for :attr that contains alpha characters:
>
>   ``` ruby
>   # RSpec
>   it do
>     record = Example.create!(attr: "some value with alpha characters")
>     expect(record).to validate_uniqueness_of(:attr)
>   end
>   
>   # Minitest (Shoulda)
>   should "validate uniqueness of :attr" do
>     record = Example.create!(attr: "some value with alpha characters")
>     assert_accepts validate_uniqueness_of(:attr), record
>   end
>   ```
>
> 2. If you meant for the validation to be case-insensitive, then you need to qualify the validation, as well as the matcher, appropriately:
>
>   ``` ruby
>   # Model
>   class Example < ActiveRecord::Base
>     validates_uniqueness_of :attr, case_sensitive: false
>   end
>
>   # RSpec
>   describe Example do
>     it { should validate_uniqueness_of(:attr).case_insensitive }
>   end
>
>   # Minitest (Shoulda)
>   class ExampleTest < ActiveSupport::TestCase
>     should validate_uniqueness_of(:attr).case_insensitive
>   end
>   ```

tovodeverett:
I think that error message is insanely awesome!!!!  In fact, it's so awesome that I'm going to purposely use a few test values that don't have alpha characters just so I can smile when it prints out!  :+1:

It does a great job of explaining to the developer what the problem is, what the alternatives are, and provides a good example of how to resolve the situation under each alternative.

Be careful - if every error message I got as a developer was this helpful, Stack Overflow would need a new business model!

albertyw:
I like the error message too - it's very clear.  However, I think it's a bit long to output to console (certainly can work, but at minimum the formatting would be stripped).  Should the error message link to the text in the readme or the website?

mcmire:
@tovodeverett Thanks for the kind words! :blush: 

@albertyw Yeah... I was starting to think that. I do love great error messages, but if you saw four of these in a row then it might be a bit much. Let's try this:

> Your Example model has a uniqueness validation on :attr which is declared to be case-sensitive, but the value the uniqueness matcher used, "123", doesn't contain any alpha characters, so using it to test the case-sensitivity part of the validation is ineffective. There are two possible solutions for this depending on what you're trying to do here:
>
> 1. If you meant for the validation to be case-sensitive, then you need to give the uniqueness matcher a saved instance of Example with a value for :attr that contains alpha characters.
>
> 2. If you meant for the validation to be case-insensitive, then you need to add `case_sensitive: false` to the validation and add `case_insensitive` to the matcher.
>
> For full examples, please see:
>
> http://github.com/thoughtbot/shoulda-matchers/wiki/error-uniqcs

albertyw:
I've updated the error with your new text.  

@mcmire - I think your original text can be copied to the wiki but I can't copy your markdowni.  Can you do it?  (I fixed your link typo to http://github.com/thoughtbot/shoulda-matchers/wiki/error-uniques)

mcmire:
Cool. Thanks for the PR. I've merged your changes as ada9bd3a1b9f2bb9fa74d0dfe1f8f7080314298c. Instead of adding docs to the wiki (which I don't plan on maintaining -- really it's just user-submitted documentation), I opted to add them to the global docs. The change for that is in 80407e602ffa6f542b90cde28983855c2f673f98.

Thanks for the discussion around this!

albertyw:
Great.  Looking forward to seeing this in the next release!

mcmire:
@albertyw @tovodeverett This is now fixed in 3.0.1.

PR #798, opened 2015-10-08
Adding support to allow_nil option for delegate_method matcher

alejandrogutierrez:
Hi everyone! I would like to add support for allow_nil option to delegate_method matcher.
```ruby
it { should delegate_method(:deliver).to(:post_office).allow_nil }
```
I update rspec-rails version in Appraisals file in order to get some new matchers.


mcmire:
Why did you make a new PR? Just force push to the other PR. (Yes, you can do that :smile:)

Issue #786, opened 2015-10-03
Getting CouldNotSetAttributeError to validate_uniqueness_of for postgres in 3.0


celsoMartins:
I searched in issues and noticed that a recent commit could solve it. But no lucky here. I have a special set method to configure the name as we want. I still getting the following error:

**Error:**
![image](https://cloud.githubusercontent.com/assets/214322/10261507/ee63848a-6971-11e5-85eb-d75524eed43f.png)

**set on the model:**
![image](https://cloud.githubusercontent.com/assets/214322/10261510/fa8bb642-6971-11e5-95d5-8817a4721adf.png)

**the failing spec:**
![image](https://cloud.githubusercontent.com/assets/214322/10261512/2655da00-6972-11e5-9107-3cc450bf222b.png)

**spec_helper**
![image](https://cloud.githubusercontent.com/assets/214322/10261514/40309096-6972-11e5-88d6-2777092eb15a.png)

--EDIT--
**Gemfile**
![image](https://cloud.githubusercontent.com/assets/214322/10261518/7178eb08-6972-11e5-9c2d-55f1dc59109b.png)
--EDIT--

I'm downgrading to a previous version until we find what is happening.

Thanks

jasnow:
Try "is_expected.to validate_uniqueness_of(:name).case_insensitive" in the spec.

celsoMartins:
No lucky.

![image](https://cloud.githubusercontent.com/assets/214322/10263649/2ac6feca-69cc-11e5-8a19-420722cbe676.png)

![image](https://cloud.githubusercontent.com/assets/214322/10263652/3d0cd226-69cc-11e5-9582-b7ee9500691d.png)

BTW, I'm also getting errors on validate_numericality_of.

![image](https://cloud.githubusercontent.com/assets/214322/10263676/00c83be2-69cd-11e5-99e2-c6325d2b1d31.png)


Thanks.

jasnow:
Try "capitalize" method as in your model. The BTW example problem is documented in #784

celsoMartins:
Like this?

```it { is_expected.to validate_uniqueness_of(:nome).capitalize }```

It even was recognized.

![image](https://cloud.githubusercontent.com/assets/214322/10264626/0b3e0d3e-69e8-11e5-8703-96e55e8e8aae.png)

[...]

Issue #799, opened 2015-10-10
ignoring_interference_by_writer for validate_numericality_of


ArturT:
I'm using shoulda-matchers 3.0.0. It would be cool if `validate_numericality_of` could take `ignoring_interference_by_writer`.

I have validator class and it runs validation on ModelEntity which is created by virtus gem. Virtus does type casting to keep my field as Integer and that's why validate_numericality_of fails.

```
Failure/Error: it { should validate_numericality_of(:node_total).is_greater_than_or_equal_to(2) }
     Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError:
       Expected Class to be able to set node_total to "1", but got 1 instead.
```

mcmire:
Ah. That's a good point. If you have a plain Ruby class, there's no way for `validate_numericality_of` to know what to do.

I hesitate to add this qualifier since it means that people could blindly apply it. I'm wondering if I could add an `cast_to` qualifier that lets you specify the type of values that the attribute hold. So in this case you might say:

``` ruby
it { should validate_numericality_of(:node_total).cast_to(Integer).is_greater_than_or_equal_to(2) }
```

Essentially, what this qualifier would do is change the matcher so that if it detects a mismatch in values, it will use the class you've given to typecast the value being set and compare it to the value read back. So it'd say `Integer("1") == 1` and since that's true it would make the matcher pass.

I'll work on that next.

ArturT:
@mcmire Sounds good. Thanks!

jetaggart:
I'm seeing a similar issue with rails 4.2.4. It seems their casting logic has changed and a string gets casted to a numeric based on the column (I'm failing to find this in the changelog, so I could be mistaken). For example:

```
Failure/Error: it { is_expected.to validate_numericality_of(:order).is_greater_than_or_equal_to(0) }
     Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError:
       Expected Class to be able to set order to "abcd", but got 0 instead.
```

mcmire:
@jetaggart Sorry Jeff. Yup, you're right that it isn't in the changelog, but it's directly due to the work on all the new typecasting stuff that @sgrif did in Rails 4.2.

jetaggart:
It's a good change, even if @sgrif is making my life difficult (:heart:). I'd be happy to help get things working with the new casting. What do you think the best solution is? Just produce values for that type?

[...]

PR #800, opened 2015-10-11
Added support for ignoring_interference_by_writer to validate_uniqueness_of

tovodeverett:
I think I've finally managed to figure out how to write the tests, and so I could finally put together this pull request.  This should help resolve the sort of issues seen in #786.

### DIFF
--- a/lib/shoulda/matchers/active_model/validation_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validation_matcher.rb
@@ -8,6 +8,7 @@ module Shoulda
         def initialize(attribute)
           @attribute = attribute
           @strict = false
+          @ignoring_interference_by_writer = false
           @failure_message = nil
           @failure_message_when_negated = nil
         end
@@ -19,6 +20,11 @@ module Shoulda
 
         def strict
           @strict = true
+          self
+        end
+
+        def ignoring_interference_by_writer
+          @ignoring_interference_by_writer = true
           self
         end
 
@@ -71,6 +77,10 @@ module Shoulda
             matcher.strict
           end
 
+          if ignoring_interference_by_writer?
+            matcher.ignoring_interference_by_writer
+          end
+
           matcher
         end
 
@@ -86,11 +96,19 @@ module Shoulda
             matcher.strict
           end
 
+          if ignoring_interference_by_writer?
+            matcher.ignoring_interference_by_writer
+          end
+
           matcher
         end
 
         def strict?
           @strict
+        end
+
+        def ignoring_interference_by_writer?
+          @ignoring_interference_by_writer
         end
       end
     end
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -502,6 +502,51 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
     end
   end
 
+  context 'when the model maps all values to upcase' do
+    let(:add_attribute_upcase_normalization) do
+      lambda do |m|
+        setter_name = "#{attribute_name}=".to_sym
+        attr_name = attribute_name
+        m.class_eval do
+          define_method(setter_name) do |value|
+            upcase_value = value.respond_to?(:upcase) ? value.upcase : value
+            write_attribute(attr_name, upcase_value)
+          end
+        end
+      end
+    end
+
+    context 'when ignoring_interference_by_writer is not specified' do
+      it 'raises a CouldNotSetAttributeError' do
+        record = build_record_validating_uniqueness(
+          attribute_type: :string,
+          validation_options: { case_sensitive: false },
+          &add_attribute_upcase_normalization
+        )
+        running_matcher = lambda do
+          validate_uniqueness.case_insensitive.matches?(record)
+        end
+
+        avm_class = Shoulda::Matchers::ActiveModel::AllowValueMatcher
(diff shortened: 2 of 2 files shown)


tovodeverett:
Hound finally stopped barking at me!

mcmire:
Thanks for doing this -- I appreciate it. The thing is, I don't want to add `ignoring_interference_by_writer` to *every* matcher because I don't want people to use it blindly the first time they see an error. Yes, rescuing CouldNotSetAttributeError will make the test pass in most cases -- but the reason I added the exception is because I wanted people to pause a bit -- and in some cases the validation isn't even necessary or feasible. So I'd rather handle these cases individually and come up with specific solutions instead of giving people a shotgun solution. (I have `ignoring_interference_by_writer` on `allow_value` because it's already a lower-level matcher and I assume that people kind of know what they're doing if they drop down to that level.)

I'm actually considering adding a `cast_to` qualifier to ValidationMatcher and the inclusion matcher so that you can tell the matcher, if the model typecasts incoming values to a certain class, then everything is okay. See #799. What do you think about this?

tovodeverett:
I think the idea of passing in a lambda works, but anything else is likely to fall short.  This isn't so much serialization as normalization, and the limits of normalization are only bounded by what the developer can invent.  The developer may trim leading and trailing spaces, may eliminate long lengths of white space, may proper case things, etc.  I've written code in the past that would accept both Metric and US Customary unitized values and would store them as metric (i.e. 3.3 ft would get stored as 1.0 in a floating point field and be displayed as either 1.0 m or 3.3 ft depending upon the user's preferences).  The issue there is that a) you're type casting to some type but b) the transformation may be fairly opaque (think of code that can accept -10.5, 10.5 S, -10 30', 10 30' S, as well as optional degree symbols, decimal minutes, seconds instead decimal minutes, etc.).

I'm using https://github.com/mdeering/attribute_normalizer for normalization, and I added custom normalizers for upcase and downcase that use the logic I used in the test suite for `ignoring_interference_by_writer` (i.e. `value.respond_to?(:upcase) ? value.upcase : value`).  I do validate my normalizer in my application's test suite.

Personally, I'm happy with `ignoring_interference_by_writer` because the part of my test suite that validates the normalizer is now independent of the part of my test suite that validates the uniqueness constraints, allow_value, etc.  Furthermore, I can divide that `allow_value` testing into 4 categories:

* `to allow_value` - values that are acceptable and require no cleanup by the normalizer
* `not_to allow_value` - values that don't get cleaned up by the normalizer, but that don't pass muster
* `to allow_value ignoring_interference_by_writer` - values that wouldn't be acceptable if the normalizer didn't clean them up, but that once cleaned up are legal
* `not_to allow_value ignoring_interference_by_writer` - values that get cleaned up by the normalizer, but that still don't pass muster

But it would be easy enough for me to define a lambda in one place with `let` and then to pass it in wherever I need it, and if the logic in the normalizer changes than I can just update that lambda in one place.  I'll be the first to admit to having too much fun with `let` lambdas in RSpec - I have places in my spec where I use lambdas defined with `let` that implement a whole set of expectations (in this case, it's validating that image manipulation is doing the correct thing), and in one case a lambda to return a custom lambda for validating a given aspect of an image.

So I can go either way on this one - I agree that `ignoring_interference_by_writer` is a potentially dangerous tool in the hands of the uninformed, but it's also an easy way to say, "Hey, I do a bunch of normalization on this, so expect that."

mcmire:
> This isn't so much serialization as normalization, and the limits of normalization are only bounded by what the developer can invent.

Sigh... see... I feel like this is one of those cases where doing things the "Rails way" and stuffing all of this behavior into your ActiveRecord model goes a little too far. If it were me, I'd make a facade that was responsible for normalization and wouldn't even have to consider the possibility of typecasting or any of that jazz. I realize that not everyone shares or is going to share my opinion, though.

Anyway :) I'm a little split on this, but this seems like a bit of an edge case to me. My feeling is that `cast_to` will work in most cases, and it would be better to wait to see how badly `ignoring_interference_by_writer` would be needed. So I think I'll leave this open for now, but implement `cast_to` and release that, and then see how it goes from there.

mcmire:
After some more thought on this, I think that adding `ignoring_interference_by_writer` is the best way to handle this issue without making tests that use this qualifier confusing (see #820). I think it would be still useful, however, to track changes to values, but only bring them to the user's attention if the matcher in question ends up failing -- this way, the user (and I) will have better insight as to why the matcher is failing.

So here are some remaining tasks for this PR:

* Add tests for all of the validation matchers (I know -- maybe there's a way we can make a shared example group or something).
* Modify `allow_value` so that if the value set on the attribute and the value obtained from the attribute are different, an extra section is added to the failure message (and this is tested for every matcher -- maybe this could go in the shared example group as well)

I can begin working on this, although if you want to pitch in on the tests that's fine too.

mcmire:
@tovodeverett I took the liberty of adding `ignoring_interference_by_writer` to every matcher myself (it turned out it took a *lot* of work -- you can see it the results in 118993480604d39c73687d069f7af3726f3e3f3e and 5532f4359aa332b10de7d46f876eaffd4a95b5b6). As a bonus, it's now enabled by default, but if an attribute changes incoming values, that's still tracked, so if the matcher fails for some reason, then we can still let the user know about that changing value. I'm going to close this, but would you be willing to point to the GitHub repo for your project and let me know if that resolves the issue you were seeing?

tovodeverett:
That is beautiful!  It sounds like you totally rethought `ignoring_interference_by_writer` along the way, which wasn't trivial.  I feel a little guilty that you went to all that work, but I hope I'm not the only happy camper so you can amortize that work over a bunch of happy users!  :+1:

PR #811, opened 2015-10-20
Fix permit docs for shoulda-context

aripollak:
`should` is not a valid method inside a `should` block, so we have to use `assert_accepts` instead. This lets us use any variables from the setup block, as opposed to a bare `should` which requires us to know all parameters up front.

### DIFF
--- a/lib/shoulda/matchers/action_controller/permit_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/permit_matcher.rb
@@ -62,9 +62,10 @@ module Shoulda
       #             password: 'password'
       #           }
       #         }
-      #         should permit(:first_name, :last_name, :email, :password).
+      #         assert_accepts permit(:first_name, :last_name, :email, :password).
       #           for(:create, params: params).
-      #           on(:user)
+      #           on(:user),
+      #           subject
       #       end
       #     end
       #
@@ -132,9 +133,10 @@ module Shoulda
       #             password: 'password'
       #           }
       #         }
-      #         should permit(:first_name, :last_name, :email, :password).
+      #         assert_accepts permit(:first_name, :last_name, :email, :password).
       #           for(:update, params: params).
-      #           on(:user)
+      #           on(:user),
+      #           subject
       #       end
       #     end
       #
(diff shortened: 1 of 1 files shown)


lib/shoulda/matchers/action_controller/permit_matcher.rb
```diff
@@ -305,18 +308,10 @@ def actual_permitted_parameter_names
...
 
-        def permit_called?
-          actual_permitted_parameter_names.any?
-        end
-
         def unpermitted_parameter_names
           expected_permitted_parameter_names - actual_permitted_parameter_names
         end
 
-        def verified_permitted_parameter_names
```
aripollak: I noticed this isn't used, but it seems like it would be useful to make sure that the actual list of permitted params matches the expected list, instead of looking for the actual being a superset of the expected.

mcmire: Good catch, thanks! As far as I know this is always how the `permit` matcher has worked, so I hesitate to change the functionality now. Otherwise, that would make sense.

lib/shoulda/matchers/action_controller/permit_matcher.rb
```diff
@@ -62,9 +62,10 @@ module ActionController
       #             password: 'password'
       #           }
       #         }
-      #         should permit(:first_name, :last_name, :email, :password).
+      #         assert_accepts permit(:first_name, :last_name, :email, :password).
       #           for(:create, params: params).
-      #           on(:user)
+      #           on(:user),
+      #           subject
```
mcmire: This looks a little confusing, what do you think about re-writing it as this?

``` ruby
matcher = permit(:first_name, :last_name, :email, :password).
  for(:create, params: params).
  on(:user)
assert_accepts(matcher, subject)
```

aripollak: Sure, should I still leave out the parens? I'm not sure what the right style is here.

mcmire: I don't use Minitest, but now that I think about it, I guess it's conventional to leave out the parens for `assert_*` methods.

aripollak: Okay, fixed.

mcmire:
Thank you for this! Merged as db67d27ceda7ea68f1c94e2956f0eada0d5b9312...8e68d99217fac5dedceeeba226ea1f2d9be01e1b.

Issue #784, opened 2015-10-02
validate_numericality_of matcher trying to set a string value in 3.0


oniofchaos:
I'm trying to upgrade to version 3.0 and am running into this issue in one of my model specs. The `validate_numericality_of` matcher is trying to set a float field to a string, and erroring it out when that doesn't work. I would think that it would be expected for that assignment to fail. :confused: 

```ruby
# spec/rails_helper.rb
Shoulda::Matchers.configure do |config|
  config.integrate do |with|
    with.test_framework :rspec
    with.library :rails
  end
end

# migration file
t.float :hours_worked

# model file
validates :hours_worked, numericality: true

# spec file
it { expect(subject).to validate_numericality_of(:hours_worked) }

# error
Failure/Error: it { expect(subject).to validate_numericality_of(:hours_worked) }
Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError:
  Expected Class to be able to set hours_worked to "abcd", but got 0.0 instead.

**Comment by chrisjoywebb** [comment]
We're receiving the same error.

**Comment by brunoocasali** [comment]
:+1:

**Comment by mcmire** [comment]
This was a conscious decision to ensure that people are writing tests that are true.

`validate_numericality_of` makes the assumption that if your attribute is set to a non-numeric value, then it will cause your model to be invalid. But this isn't actually happening: since your column is numeric itself, it will typecast any value to a numeric (in this case a float). So your validation won't ever fail, because all values are now always numeric.

The exception you're getting is low-level and obviously we should make some improvements to the message, but basically, it's indicating that since the validation always passes, you can remove it and the test itself. Does this make sense?

**Comment by oniofchaos** [comment]
Yes, that makes sense now that you've explained it. Thanks! Definitely agree on seeing an improved message :+1:

**Comment by sunderwood** [comment]
I use `validate_numericality_of` to validate that the number is within a certain range of values (given that I know the field is numeric).  Using `validate_numericality_of` provides better default error messages and options than `validate_inclusion_of`. 

I can understand the desire to ensure that people are not adding superfluous validations to the model. Your test, and error are fine if someone does not include any of the options for the `validates_numericality_of`. Perhaps the test should warn that the field is already numeric instead of failing the test.

In the example above, if the validation was 
```
validates :hours_worked, numericality: {greater_than: 0}
```
I would expect
```
it { expect(subject).to validate_numericality_of(:hours_worked).is_greater_than(0)}
```
[...]

PR #812, opened 2015-10-21
Add decimal column type to numericality matcher

reacuna:
From the discussion on https://github.com/thoughtbot/shoulda-matchers/issues/784, add decimal column support.


mcmire:
Thanks! Merged as a0f12216fd12627c2545b082f5e463a712ba75e5.

PR #813, opened 2015-10-22
Run on latest point releases of Ruby

maclover7:


.travis.yml
```diff
@@ -9,8 +9,8 @@ env:
 
 rvm:
   - 2.0.0
-  - 2.1.4
-  - 2.2.0
+  - 2.1.7
+  - 2.2.3
```
tute: Should we remove the patch versions? Suggestion:

```
   - 2.0
   - 2.1
   - 2.2
``

mcmire: Oh, I didn't know you could do this. If this works, then yeah, I think this makes sense.

maclover7:
:+1:

tute:
:+1:

maclover7:
ping @mcmire

tute:
If it's ok with you @mcmire, I can rebase and merge branches in this project when we both agree (and you didn't get to before).

mcmire:
@tute Okay! That would be helpful. Thank you.

By the way this looks good. :+1:

tute:
Thank you @maclover7! :clap:

maclover7:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #818, opened 2015-10-26
Do not apply .railsrc when generating testing environment

mdeering:
Getting ready to submit a patch/pull-request for this gem and I had trouble setting up and running the testing environment on this repository. I managed to trace the issues to the fact that it the command runner was shelling out with a *rails new* and applying my ~/.railsrc options.

Simple fix here is to just add the *--no-rc* flag to the *rails new* calls to maintain consistent generation of the unit and acceptance rails test applications.


mattbrictson:
I was just bitten by this as well. Nice fix! I didn't know about the `--no-rc` option.

:+1:

PR #819, opened 2015-10-26
support for ignoring_interference_by_writer on validate_length_of

mdeering:
I agree with @mcmire on pull request #800 that ignoring_interference_by_writer should not be blanketed into to all matchers by implementing in ValidationMatcher.

I want to see if this is more *likeable* implementation that could potentially be applied on a case-by-case basis to other matchers that may require it such as in the cases of #800, #799,  #786.

I think if it is the case that ignoring_interference_by_writer ends ups needing implementation in more then a few select matchers a pattern will emerge for DRYing it up if that does become an issue. I just want to run this approach past the maintainers before attempting this on other matchers within the gem.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
@@ -302,6 +302,12 @@ module Shoulda
           self
         end
 
+        protected
+
+        def ignoring_interference_by_writer?
+          @ignoring_interference_by_writer
+        end
+
         private
 
         def translate_messages!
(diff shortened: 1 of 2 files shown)


mdeering:
@mcmire I'm pretty short on time right now. Let me know about the approach for starters then I will wrestle with those hound errors.  Build/checks all still pass. Do the hound errors matter, did my best to match the syntax I noted around where code/specs were added?

mcmire:
@mdeering This is great, but I don't think that adding `ignoring_interference_by_writer` is the right approach. I added it to `allow_value` because it's so low level that I figured if you used it you probably know what you're doing. However for the other matchers, I feel like there could very well be legitimate reasons why the exception is occurring, and so giving people a way to sweep the problem under the rug probably isn't the best solution.

I've opened up a PR, #820, and pushed the work I've been doing so far on it. I welcome any feedback on it.

mdeering:
@mcmire I agree with you, I have only looked at this issue from the context of a single matcher so far.  I can see from a quick look at #820 you have a more holistic approach already baking. I will give it a once over in the next day or two and see if there is something I can contribute to it.  Cheers.

mcmire:
@mdeering Thanks for the work on this. I ended up spending time and adding `ignoring_interference_by_writer` to every matcher myself (turned out it took a *lot* of work). I'm going to close this PR, but would you be willing to point to this repo and then re-run your tests to see if that resolved the issue?

mdeering:
Thanks, I will give it a try. Looks like you want the tests ran against master? Any chance you can sneak #818 into master also so I can avoid shuffling my railsrc out of the way again?

mcmire:
@mdeering Done!

PR #822, opened 2015-10-28
Numericality validation of virtual attributes

bdmac:
Commit 18b2859d2522a4866c398b9a32ebc3de4ec79389 broke numericality
validation of virtual attributes in ActiveRecord models. This commit
added a `column_type` method that assumes if `columns_hash` is a thing
then our attribute would be present within it. This is not true for
virtual attributes and leads to:
```
     NoMethodError:
           undefined method `type' for nil:NilClass
         # /Users/bdmac/.gem/ruby/2.2.2/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_model/validate_numericality_of_matcher.rb:442:in
         # `column_type'
```

This adds a check to `column_type` to return nil instead of throwing an
exception if a specified attribute does not exist in the `columns_hash`
because it is virtual.

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
@@ -127,6 +127,13 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
         expect(record).to validate_numericality
       end
 
+      context 'when the attribute is a virtual' do
+        it 'accepts' do
+          record = build_record_validating_numericality_of_virtual_attribute
+          expect(record).to validate_virtual_attribute_numericality
+        end
+      end
+
       context 'when the column is an integer column' do
         it 'raises an IneffectiveTestError' do
           record = build_record_validating_numericality(
@@ -224,6 +231,15 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
         expect(record).to validate_numericality.odd
       end
 
+      context 'when the attribute is a virtual' do
+        it 'accepts' do
+          record = build_record_validating_numericality_of_virtual_attribute(
+            odd: true,
+          )
+          expect(record).to validate_virtual_attribute_numericality.odd
+        end
+      end
+
       context 'when the column is an integer column' do
         it 'accepts (and does not raise an error)' do
           record = build_record_validating_numericality(
@@ -278,6 +294,15 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
         expect(record).to validate_numericality.even
       end
 
+      context 'when the attribute is a virtual' do
+        it 'accepts' do
+          record = build_record_validating_numericality_of_virtual_attribute(
+            even: true,
+          )
+          expect(record).to validate_virtual_attribute_numericality.even
+        end
+      end
+
       context 'when the column is an integer column' do
         it 'accepts (and does not raise an error)' do
           record = build_record_validating_numericality(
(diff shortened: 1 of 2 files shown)


bdmac:
I fixed the hound comments by adding hanging commas in my tests. I'm not sure why I had to since none of the other multiline method calls in that same file actually do that but cool.

mcmire:
We didn't use to have that rule in Hound, and it was added some time ago, so that's why Hound is barking about them now.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -1141,6 +1216,21 @@ def define_model_validating_numericality(options = {})
     end
   end
 
+  def define_model_validating_numericality_of_virtual_attribute(options = {})
+    virtual_attr = options.delete(:virtual_attribute) do
+      virtual_attribute_name
+    end
+
+    define_model 'Example' do |model|
```
mcmire: `define_model` creates an ActiveRecord model. We also have a `define_active_model` helper you could use that defines an ActiveModel::Model model instead. That means you could drop the next line actually. So it would look like:

``` ruby
define_active_model('Example', accessors: [virtual_attr]) do |model|
```

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -1157,7 +1247,15 @@ def validate_numericality
...
 
+  def validate_virtual_attribute_numericality
+    validate_numericality_of(virtual_attribute_name)
+  end
+
   def attribute_name
     :attr
   end
+
+  def virtual_attribute_name
```
mcmire: I think we can simply reuse `attribute_name` within `define_model_validating_numericality_of_virtual_attribute`. I suppose it's helpful conceptually, but it's not like we use both `attribute_name` and `virtual_attribute_name` in the same test. This means we could remove `validate_virtual_attribute_numericality`.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
mcmire: I realize that's not what your use case is, but the numericality matcher should work for both ActiveRecord and ActiveModel models, and we don't have tests for ActiveModel here I'm noticing. So we could kill two birds with one stone. What do you think?

bdmac: An ActiveModel model would not have tripped this problem with virtual attributes though because it would not have responded to columns_hash in the first place. That check was already there. This specifically happens with an ActiveRecord model using virtual attributes as in has_secure_password.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
bdmac: That's true I started out just adding it to the existing model builder but that did not pan out. I can remove this and reuse `atteibute_name` instead since nothing really cares what the attr is called.

bdmac:
Updated

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
mcmire: Ah... yup, that makes sense.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -1157,6 +1247,10 @@ def validate_numericality
     validate_numericality_of(attribute_name)
   end
 
+  def validate_virtual_attribute_numericality
```
mcmire: You can probably remove this too, right? Since this is now the same as `validate_numericality`.

bdmac: :+1:

mattbrictson:
@bdmac Thank you! Would love to see this merged, but in the meantime I'm using your branch and it has solved numerically validation for me.

In my case the virtual attribute is coming from `store_accessor`, e.g.

```ruby
store_accessor :options, :maximum
```

Where `options` is a PostgreSQL jsonb column.

mcmire:
Sorry for the delay on this, I was preoccupied with fixing another issue. This is now in master as 12e788f1d01be23a08237fe45eafe94313fa6214.

Issue #688, opened 2015-03-27
validates_numericality_of with symbol instead of number.


esbanarango:
Hello and thank you for this project.

I have this validation on my model:

``` ruby

    validates_numericality_of :quantity_min, greater_than: 0
    validates_numericality_of :quantity_max, greater_than: :quantity_min
```

it works as expected but when I try to test it with the _shoulda-matchers_ validators:

``` ruby

it { is_expected.to validate_numericality_of(:quantity_max).is_greater_than(:quantity_min)}

```

I got: `undefined method '+' for :quantity_min:Symbol`

mcmire:
Thanks for the report. We don't support method names for those options, but we'd gladly take a pull request that fixes this.

mcmire:
Hey folks. In an effort to lighten our load as maintainers and be able to serve you better in the future, the shoulda-matchers team is working on cleaning out the cobwebs in this repo by pruning the backlog. As there are few of us, there are a lot of items that will simply never earn our attention in a reasonable time frame, and rather than giving you an empty promise, we think it makes more sense to focus on more recent issues. That means, unfortunately, that we must close this issue.

Don't take this the wrong way: our aim is not to diminish the effort people have made or dismiss problems that have been raised. If you feel that we should reopen this issue, then please let us know so that we can reprioritize it. Thanks!

PR #823, opened 2015-10-28
Support symbol for `validates_numericality_of`

vyrak:
Have `vaidates_numericality_of` support symbols for feature request #688


totallymike:
I think this looks pretty good, and am also looking for this feature.

MDjavaheri:
@mcmire Any timetable as to when this will be resolved and merged?

mcmire:

bbugh:
@mcmire can we reopen this? `validate_numericality_of` doesn't support symbol fields which misses out on the rails feature of comparing vs a column. It looks like this was ready to go!

PR #835, opened 2015-11-06
Make README more specific for new Rspec

obduk:


soulcutter:
As an RSpec core member, I :heart: this change. ^5

obduk:
Is there anything I need to do to get this merged? Thanks.

mcmire:
Hey, sorry about the late reply on this, I've been out. I'm deciding what to do about this. This would be confusing to our fellow teammates, since we don't write RSpec tests this way. I'll have to come back to this.

obduk:
What syntax do you use? I guess the other option would be to include 2 examples?

The main reason I think this example is needed as for a new project using rspec 3, the examples do not work without the model type test, which can be confusing for new people.

obduk:
Also I think if not an explanation of why it is needed, with either this solution or the one in #924 would help people (I really don't mind either)

mcmire:
We simply say:

``` ruby
describe Foo do
  # ...
end

describe FooController do
   # ...
end
```

and let the location of the files determine what kind of test it is. Of course, this only works if `infer_spec_type_from_file_location!` is enabled.

I'll ask around and see if this syntax is something we would want to switch to internally. If not, I do agree that we should update this in the README.

soulcutter:
`config.disable_monkey_patching!` is in the default `spec_helper.rb` generated by `rspec --init`, if that has any bearing on how you want to document this. I do prefer that style, but to each their own.

I like infering spec types from file locations too, though most of the rest of the core team does not. The generators are explicit about the metadata.

If you go by the style recommended by RSpec defaults, this changeset makes sense. But you don't have to - to each their own.

obduk:
@soulcutter the only being available to model test is not caused by the `disable_monkey_patching` option but the `infer_spec_type_from_file_location` which is not in the default config or behaviour as of RSpec 3, hence the confusion. Either adding the test type model, or enabling `infer_spec_type_from_file_location` fixes it.

soulcutter:
I know, there are two things going on in the PR though - RSpec.describe vs describe and type: :model vs inferred. I was covering both in my response, sorry if I wasn't very clear.

Bradley Schaefer

obduk:
@soulcutter sorry, just realised you are the RSpec core member from earlier.

mcmire:
> `config.disable_monkey_patching!` is in the default spec_helper.rb generated by `rspec --init`, if that has any bearing on how you want to document this.

Is that true? I find that it's suggested, but commented out. Unless I'm on an older version of RSpec.

soulcutter:
@mcmire It is true. It is not commented out in 3.4.0 (I'd have to check for when that changed, but I feel like it's been a while)

mcmire:
Yeah, I just confirmed running `rails g rspec:install` on a fresh Rails app, using rspec-rails 3.4.2, and it seems that not only does `spec_helper.rb` have `config.disable_monkey_patching!` disabled by default, but `rails_helper.rb` has `config.infer_spec_type_from_file_location!` enabled by defualt.

So if that's the case, what's the official RSpec stance on this?

soulcutter:
The default generated `spec_helper.rb` is RSpec's stance on good defaults. Most of the team wants to make explicit metadata (`type: :model`) the standard, but that hasn't filtered down to the default yet apparently.

A naked `describe` produces an error with the defaults.

Missing `type: :model` will run, but potentially does not include stuff that is included by matching on that metadata.

mcmire:
So, another update: I just tested this out on a fresh app and it appears that you *can* actually use a naked `describe`, without `type: :model`, and it will just work.

But anyway, it sounds like that style is being phased out slowly, and we have a much wider audience outside of the company than inside. So, okay, I'll change this.

mcmire:
Okay, merged as a1eee11bacfd75a0276229b231ad78a19216eebd.

I also took the opportunity to go through and update all of the inline documentation, that was done in afa6a7b6661a0d4450d330922c2c930ea2f3ff68.

Cheers!

Issue #838, opened 2015-11-11
allow_value should detect if values are changed post-validation


mcmire:
A colleague at work just ran into #836, except the scenario is different. Basically, Devise adds a `before_validation` on whatever field you're using for authentication (let's say `username`) so that the value this is set to will be downcased. So this screws up the uniqueness matcher in a similar manner to what's posted in that issue.

Unfortunately, the matcher isn't detecting this and so you get a nebulous failure message from the uniqueness matcher that "A" was set on the record and an error message was found on "a". It'd probably be worth it to add some kind of check after we call `valid?` to ensure that all of the values that we've set on the model remain the same post-validation.

dgmstuart:
Perhaps I've misunderstood, but for me, whether the values pre- and post-validation are the same isn't something I care about when validating uniqueness: in some cases I'll explicitly want to change values as part of validation, so I _want_ them to be different but it may not be a property of my code that I'm interested in making assertions about.

Of course, in some situations I may want to explicitly assert this property, but it's odd to me that `shoulda-matchers` would be asserting it in an apparently unrelated check: from my perspective this feels like part of my test suite is deciding to care about properties that I don't care about, in a way that I can't control.

Or are we saying that `shoulda-matchers` is only applicable if you never change values in-between setting and getting? I guess that would be a pretty clear and coherent constraint, but it would be a little sad as well.

mcmire:
The goal I am trying to achieve is to provide more informative and helpful error messages, and to eliminate confusion when a matcher fails.

Currently, `allow_value` will raise an exception if the value that the matcher puts into the attribute under test isn't the same as the value coming out of the attribute. This behavior may be intentional, but it changes the test being run, and can lead to unintuitive test failures (which I've personally witnessed).

In #836, a similar thing occurs, except the value the attribute stores doesn't change immediately, it changes when validations are run. Here's an example of the model:

``` ruby
class User
  before_validation :downcase_username

  validates :username, uniqueness: true

  private

  def downcase_username
    self.username = username.downcase
  end
end
```

If you test this using shoulda-matchers:

``` ruby
describe User do
  it { should validate_uniqueness_of(:username) }
end
```

this test will fail with something like:

``` ruby
Expected not to have errors on :username when value was set to "A", got errors:

* "has already been taken" (attribute: :username, value: "a")
```

This is a very confusing error message. The matcher should "just work", right?

[...]

PR #839, opened 2015-11-13
838 add ignoring case sensitivity to uniqueness validation matcher

### LINKED ISSUE #836: `validates_uniqueness_of` with custom setter fails with "should require case sensitive unique value for email"
On upgrading from 2.x to 3.0.1 I get the following error:

```
User when the record has a password_digest should require should require case sensitive unique value for email
     Failure/Error: it { should validate_uniqueness_of(:email).with_message(/already been used by you/) }
       The allow_value matcher attempted to set :email on User to
       "TEST1@FOO.COM", but when the attribute was read back, it had stored
       "test1@foo.com" instead.

       This creates a problem because it means that the model is behaving in a
       way that is interfering with the test -- there's a mismatch between the
       test that was written and test that was actually run.

       There are a couple of reasons why this could be happening:

       * The writer method for :email has been overridden and contains custom
         logic to prevent certain values from being set or change which values
         are stored.
       * ActiveRecord is typecasting the incoming value.

       Regardless, the fact you're seeing this message usually indicates a
       larger problem. Please file an issue on the GitHub repo for
       shoulda-matchers, including details about your model and the test
       you've written, and we can point you in the right direction:

       https://github.com/thoughtbot/shoulda-matchers/issues
```

This seems to be because shoulda-matchers now cares about case-sensitivity when checking for uniqueness, and we have a custom setter which is downcasing values before saving:

```ruby
# models/user.rb
validates :email, uniqueness: true

def email=(string)
  return super string unless string
  super string.strip.downcase
end
```

```ruby
# models/user_spec.rb
subject(:user) { create(:user) }
it { should validate_uniqueness_of(:email) }
```

I don't know if this custom setter is a good or bad idea - happy to take advice on that. My guess is that shoulda-matchers possibly can't deal with this sort of thing because of the way it works under the hood (iirc it uses the model's getters and setters, right?).

I've tried the example with `.case_insensitive` added to the matcher, and I get basically the same message.

Would it make sense for shoulda-matchers to ignore this scenario unless `:case_insensitive` is true? Or provide an option to skip this part of the checks? I'm happy to work on a pull request if we can agree a sensible approach.

dgmstuart:
Fixes #836 
also possibly #838?

This branch is based off PR #820, but I think it could quite easily be rebased onto master with a couple of tweaks to the test suite.

The issue is that the existing functionality was either actively asserting case sensitivity or case insensitivity (depending on the `:case_sensitive` option). This commit adds an option to not assert either.

This allows handling of the scenario where the case of an attribute is
changed at some point before being assigned on the model.

Possibly some more work to be done, which may or may not belong on this PR? 
* A test cases around altering case as part of validation (as in #838)
* Perhaps it now makes sense to have an explicit `.case_sensitive` option?
* If the model has specified `case_sensitive: true` or `case_sensitive: false`, but the test suite specifies `ignoring_case_sensitivity` then it feels like there's an inconsistency - should we highlight this to the user, or treat the presence of this option as the user telling us not to?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,24 +1,49 @@
+# Contributing to shoulda-matchers
+
 We love contributions from the community! Here's a quick guide to making a pull
 request.
 
-## Overview
+1. If you haven't contributed before, please read and understand the [Code of
+   Conduct].
 
-0. If you haven't contributed before, please read and understand the [Code of
-Conduct].
+1. Ensure that you have a [working Ruby environment].
 
-1. Fork the repo.
+1. Fork the repo on GitHub, then clone it to your machine.
 
-2. Install [dependencies](#installing-dependencies).
+1. Now that you've cloned the repo, navigate to it and install dependencies by
+   running:
 
-3. Run the tests. We only take pull requests with passing tests, and it's great
-to know that you have a clean slate: `bundle && bundle exec rake`
+   ```
+   bundle install
+   ```
 
-4. If you're adding functionality or fixing a bug, add a failing test for the
-issue first.
+1. All tests should be passing, but it's a good idea to run them anyway
+   before starting any work:
+   
+   ```
+   bundle exec rake
+   ```
 
-5. Make the test pass.
+1. If you're adding functionality or fixing a bug, you'll want to add a
+   failing test for the issue first.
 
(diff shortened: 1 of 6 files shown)


dgmstuart:
Closing to rebase onto master

PR #840, opened 2015-11-13
Add ignoring case sensitivity to uniqueness validation matcher

### LINKED ISSUE #836: `validates_uniqueness_of` with custom setter fails with "should require case sensitive unique value for email"

dgmstuart:
Fixes #836 
also possibly #838?

Duplicates some of the code on the WIP PR #820 - will cause some conflicts with that, but hopefully nothing too tricky. 

The issue is that the existing functionality was either actively
asserting case sensitivity or case insensitivity. This commit adds an
option to not assert either.

This allows handling of the scenario where the case of an attribute is
changed at some point before being assigned on the model.

Perhaps it now makes sense to have an explicit `.case_sensitive` option as well?

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -726,6 +726,51 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
     end
   end
 
+  context 'when the writer method for the attribute changes incoming values' do
+    context 'and ignoring_case_sensitivity is not specified' do
+      it 'raises a CouldNotSetAttributeError' do
+        model = define_model_validating_uniqueness(
+          attribute_name: :name,
+          validation_options: { case_sensitive: false },
+        )
+
+        model.class_eval do
+          def name=(name)
+            super(name.upcase)
+          end
+        end
+
+        assertion = lambda do
+          expect(model.new).
+            to validate_uniqueness_of(:name).
+            case_insensitive
+        end
+
+        expect(&assertion).to raise_error(
+          Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError
+        )
+      end
+    end
+
+    context 'and ignoring_case_sensitivity is specified' do
+      it 'accepts (and not raise an error)' do
+        model = define_model_validating_uniqueness(
+          attribute_name: :name,
+        )
+
+        model.class_eval do
+          def name=(name)
+            super(name.upcase)
+          end
+        end
+
+        expect(model.new).
+          to validate_uniqueness_of(:name).
+          ignoring_case_sensitivity
+      end
+    end
+  end
+
   let(:model_attributes) { {} }
 
   def default_attribute
@@ -846,6 +891,7 @@ describe Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher, type: :mo
   end
 
   def define_model_validating_uniqueness(options = {}, &block)
+    attribute_name = options.fetch(:attribute_name) { self.attribute_name }
     attribute_type = options.fetch(:attribute_type, :string)
     attribute_options = options.fetch(:attribute_options, {})
     attribute = {
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -218,6 +218,7 @@ module Shoulda
         def initialize(attribute)
           super(attribute)
           @options = {}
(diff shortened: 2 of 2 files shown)


dgmstuart:
@mcmire Hi - if there's anything I can do to help get this reviewed and merged please let me know :)

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -726,6 +726,51 @@
...
+        expect(&assertion).to raise_error(
+          Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError
+        )
+      end
+    end
+
+    context 'and ignoring_case_sensitivity is specified' do
+      it 'accepts (and not raise an error)' do
+        model = define_model_validating_uniqueness(
+          attribute_name: :name,
```
mcmire: I'm noticing that in the first test, you have the validation qualified with `case_sensitive: false`, but you don't have that here.

I'm kind of forgetting what the original bug was here. Is that something we need to have here, you think? Or, should we have two sets of tests, one where the validation is  `case_sensitive: true`, one where it's  `case_sensitive: false`?

mcmire:
@dgmstuart Hey, sorry for the delay on getting back to you about this. I've finally resolved the issues with `ignoring_interference_by_writer`, so I'm turning my attention to this now. I had a quick question above, although I need to take a look at the original issues around this and review this further.

mcmire:
By the way, I think this PR can incorporate an idea I had in #838, which is to have the uniqueness matcher, upon failure, suggest to the developer that he or she try using `ignoring_case_sensitivity` in the failure message. Unfortunately, there isn't a great way to add some text to the failure message. `allow_value` is the matcher that the uniqueness matcher uses internally, so if the uniqueness matcher were to fail, then it's `allow_value` that would be producing a failure message. `allow_value` does provide a sort of extension point that lets you customize the failure message already: you can set `failure_message_preface` to a proc that returns some text ([here's](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L306) how that's used now). But as its name implies, it affects the beginning of the message, not the end. We can probably add a similar thing -- maybe call it `failure_message_suffix` -- to `allow_value`, and then use that inside of the uniqueness matcher, in order to add an additional message.

mcmire:
Also, I was going back through the issues and I think this PR should fix #786 as well.

mcmire:
Hey - so I made some tweaks to your branch, and specifically made the change I mentioned above concerning the failure message. [Here's](https://github.com/thoughtbot/shoulda-matchers/compare/add-ignoring-case-sensitivity?expand=1) those changes along with your original changes. What do you think?

mcmire:
Merged as d7d96ad207b52b01f71cfce92f9719ccdf392ad1.

dgmstuart:
Hi - great to gave this in the new release. Your changes look good to me - adding a message suggesting using the new option is nice! 

Have a great weekend.

mcmire:
Thanks!

PR #842, opened 2015-11-15
Fix join_table_matcher to work with pg views

cheshire-cat:
Change the method of table existence confirmation in `#join_table_exists?`
from `#tables.include?` to `#table_exists?`.

Why:

* In fact, the behavior of `#tables` is different for each adapter.
  For some adapters (e.g. pg_adapter) it reflects only tables.
  And `#table_exists?` always checks both tables and views.


cheshire-cat:
Typo in commit message

mcmire:
Note for the future -- if you want to change a commit, you can force-push to a branch instead of having to create a new PR :)

cheshire-cat:
Ok. thanks for an advice.

PR #847, opened 2015-11-18
allow respond_with matcher to match on lambda

TiteiKo:
In an attempt to dry up my specs, I was really missing the ability to do something like:

``` ruby
subject { -> { post :create, foo: 'bar' } }

it { is_expected.to(respond_with :created) }
it { is_expected.to change(Foobar, :count).by(1) }
```

I think it's much simple, clean and readable than the following or any other equivalent:

``` ruby
it do
  post :create, foo: 'bar'
  is_expected.to(respond_with :created)
end

it { expect{ post :create, foo: 'bar' }.to change(FooBar, :count).by(1) }
```

This patch allows the matcher to work with lambda returning an ActionController::TestResponse

Note: I'm not really sure how to test this in the current specs, but it's late here so I'll look into it tomorrow

### DIFF
--- a/lib/shoulda/matchers/action_controller/respond_with_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/respond_with_matcher.rb
@@ -94,8 +94,13 @@ module Shoulda
           @status = symbol_to_status_code(status)
         end
 
-        def matches?(controller)
-          @controller = controller
+        def matches?(subject)
+          if subject.is_a? Proc
+            @response = subject.call
+          else
+            @response = subject.response
+          end
+
           correct_status_code? || correct_status_code_range?
         end
 
@@ -123,7 +128,7 @@ module Shoulda
         end
 
         def response_code
-          @controller.response.response_code
+          @response.response_code
         end
 
         def symbol_to_status_code(potential_symbol)


mcmire:
I'm not sure I agree with this change. It doesn't seem like good design to make a matcher take something that wouldn't ordinarily take a lambda to take in order to make another test which has nothing to do with a TestResponse (or a controller object, for that matter) more palatable.

A useful way to approach this, I think, is to consider the fact that all of the matchers in this gem are just extractions from tests that would have otherwise been manually written.

The "long way" of writing your tests would be like so:

``` ruby
it "responds with 200" do
  post :create, foo: 'bar'
  expect(controller).to respond_with(:created)
end

it "creates a Foobar record" do
  old_foobar_count = Foobar.count
  post :create, foo: 'bar'
  new_foobar_count = Foobar.count
  number_of_created_foobars = new_foobar_count - old_foobar_count
  expect(number_of_created_foobars).to be 1
end
```

The `change` matcher DRYs up the second test so that you can now write:

``` ruby
it "creates a Foobar record" do
  expect { post :create, foo: 'bar' }.to change(Foobar, :count).by(1)
end
```

The reason why `change` takes a block is because `Foobar.count` has to be evaluated twice: once before the block runs, and once after. And the block could represent *any* piece of code, not merely just hitting an action (although that's the common case).

How do you DRY up the first test? A better question is, what code are you repeating that you're attempting to refactor? You could make the argument you don't want to repeat performing the action. If that's the case, then you could try extracting it to a method. Then you'd have:

``` ruby
def perform_action
  post :create, foo: 'bar'
end

it "responds with 200" do
  perform_action
  expect(controller).to respond_with(:created)
end

it "creates a Foobar record" do
  expect { perform_action }.to change(Foobar, :count).by(1)
end
```

Is there any way to DRY it up more? I don't think so -- I think you've done all you can do, it doesn't make sense to me to carry that one step further. In the first test, it's the *controller* that responds with 200, so having `respond_with` take a lambda would make the test read funnily. In the second test, it's okay because you can read it as "I expect calling `perform_action` to change `Foobar.count` by 1".

Does that all make sense?

TiteiKo:
Hi @mcmire , thanks for the detailed reply!

I had the same thoughts at first, and did the same refactoring as you just showed, and I'm sorry but (at least for now) I disagree...

First of all, let me make a small correction on what you said, the long way for the status would actually be:

```ruby
it "responds with 200" do
  post :create, foo: 'bar'
  expect(response.status).to eql(200)
  # or
  expect(response).to have_http_status(200)
end
```

As you can see, it's the response that has a status of 200, and, in an english approach, you expect the result of a request to respond with 200. 

From the rspec doc on controller spec:

>  It allows you to simulate a single http request in each example, and then
specify expected outcomes such as:
>
> * rendered templates
> * redirects
> * instance variables assigned in the controller to be shared with the view
> * cookies sent back with the response

With the upcoming changes from rails 5 and the fact that testing templates and instance variables in controllers is being dropped and already deprecated (https://github.com/rails/rails/issues/18950), the only things that should be tested in a controller spec are the responses to requests and things set in motion by them. 

It includes for example:

* redirections
* http statuses
* session store modifications
* flash store modifications
* changes in mailer/background jobs queues
* changes in objects / collections

So, in my opinion, having a subject being a lambda of the request is consistant with all of this. And having the actual subject of the test be the response is too.

As you can see in the matcher's code, the only method of the controller called is ```response```, and the reason is that the real test here is on the response.

My problem was not so much being lazy or unable to refactor my specs, but the simplicity of them.

The code reads in english:

> I'm testing the execution of a post request
> I expect it to have a response with status 200
> I expect it to add a Foo

Instead of:

> I do a post request
> I expect the last response emitted by my controller to have a status of 200
> I expect making a post request to add a Foo

mcmire:
Reading this over again, I still have the same view as before. I haven't seen this enough times in the wild to where I feel like it would be used by a lot of people. For this reason, I don't think this would be a good addition to the gem. Sorry.

Issue #805, opened 2015-10-13
validate_presence_of says "Failure message is nil" when attribute is serialized


wgc-as:
This is my model:

``` ruby
class User < ActiveRecord::Base
  serialize :settings, Hash
end
```

This is the spec:

``` ruby
require 'rails_helper'

RSpec.describe User, type: :model do
  it { is_expected.to     serialize(:settings) }
  it { is_expected.to_not validate_presence_of :settings }
end
```

It throws this error:

```
  1) User should not require settings to be set
     Failure/Error: it { is_expected.to_not validate_presence_of :settings }
     ArgumentError:
       Failure message is nil. Does your matcher define the appropriate
failure_message[_when_negated] method to return a string?
     # ./spec/models/user_spec.rb:5:in `block (2 levels) in <top (required)>'
```

In my opinion this should pass as I don't have to assign any value to `settings` to produce a valid entity. But even if it fails, it should give a better error message.

This worked before I upgraded to shoulda 3.0.0.

rails (4.2.4)
rspec-core (3.3.2)
rspec-expectations (3.3.1)
rspec-mocks (3.3.2)
rspec-rails (3.3.3)
rspec-support (3.3.0)
shoulda-matchers (3.0.0)

Mac OS X 10.9.5, Ruby 2.2.2

mcmire:
I don't know why you'd be getting this, but have you tried putting parentheses around `:settings`?

wgc-as:
It does not help. I created a repository which reproduces the issue:
https://github.com/wgc-as/shoulda_validate_presence_of

I forgot to commit the initial state after rails new separately, but the only addition is the "rspec-rails" and "shoulda-matchers" gems in the Gemfile, the migration file and the User model.

mcmire:
Thanks, I'll take a closer look at it.

wgc-as:
After some debugging, it turns out that an error is swallowed in `ValidatePresenceOfMatcher#disallows_original_or_typecast_value?` method:
```
#<Shoulda::Matchers::ActiveModel::AllowValueMatcher::CouldNotSetAttributeError:
The allow_value matcher attempted to set :settings on User to nil, but
when the attribute was read back, it had stored [] instead.

This creates a problem because it means that the model is behaving in a
way that is interfering with the test -- there's a mismatch between the
test that was written and test that was actually run.

There are a couple of reasons why this could be happening:

* The writer method for :settings has been overridden and contains
[...]

PR #851, opened 2015-11-27
Bugfix/validate presence of with serialize

j15e:
It think this should fix the issue described in #805.

I thought this would also fix a similar issue I have, but it does not... I will keep investigating. In my specific scenario I validate the presence of an attribute created by `hstore_translate` and I get the same error as described in #805.


mcmire:

PR #855, opened 2015-12-09
Match more accurately when key AND value are set for SetFlashMatcher

soulcutter:
When specifying both a flash key AND value, this ensures that the
matching value is within the given flash key. Previously it would look
for the matching value within ALL the flash keys, which could result in
false positives.

For example:

``` ruby
expect(controller).not_to set_flash['the key'].to('the value')
```

This example used to make two independent assertions - that the flash key
'the key' is set, and that some flash value includes 'the value'. The
new behavior asserts that the flash value associated with 'the key' is
set to 'the value'.

This same applies to the session matcher as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,12 +80,9 @@ curl -sL https://rpm.nodesource.com/setup | bash -
 yum install -y nodejs
 ```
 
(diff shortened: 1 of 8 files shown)


mcmire:

PR #859, opened 2015-12-10
Fix NoMethodError when using `should_not allow_value` with Minitest+shoulda-context

mattbrictson:
When using Minitest (i.e. `ActiveSupport::TestCase`) in combination with shoulda-context, a failing `should_not allow_value...` assertion would result in this error:

```
NoMethodError: undefined method `negative_failure_message'
```

This commit fixes this error by defining `failure_message_for_should_not`, which shoulda-context will use instead.

### DIFF
--- a/lib/shoulda/matchers/active_model/allow_value_matcher.rb
+++ b/lib/shoulda/matchers/active_model/allow_value_matcher.rb
@@ -492,6 +492,8 @@ module Shoulda
 
           Shoulda::Matchers.word_wrap(message)
         end
+        alias_method :failure_message_for_should_not,
+          :failure_message_when_negated
 
         def attribute_changed_value_message
           <<-MESSAGE.strip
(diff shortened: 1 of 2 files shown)


mcmire:
Ah... I had no clue.

Do you think this is something we should change in shoulda-context instead?

mattbrictson:
> Do you think this is something we should change in shoulda-context instead?

Meaning, perhaps we should change shoulda-context to use `failure_message_when_negated` instead? I am just barely familiar with the code, so I don't know whether that would be a high-risk change. My goal with this PR was to make the simplest change that could fix the bug.

If you think it's better to change shoulda-context, I can create a PR there instead.

mcmire:
Alright. Let me think about this a little more and get back to you.

mattbrictson:
I rebased against current `master` and fixed the conflicts.

I'm still curious if you are open to merge this or if should pursue changes to shoulda-context instead.

mcmire:
I'm about to go on a long trip overseas, and I'm afraid I don't have time to address this right now.

I do think this change should be made in shoulda-context, though, simply because it'll be easier. I believe in order to make this work here, you'd have to add `failure_message_for_should_not` to every single matcher, not just `allow_value`, which I'd want to avoid.

mcmire:
There, I even added a tag :)

mcmire:

PR #866, opened 2015-12-18
WIP: Add `ignoring_interference_by_writer` to everything

mcmire:
This is basically an answer to #799, #800, #819, and other issues, and also supercedes #820. 

The idea here is to add `ignoring_interference_by_writer` to all matchers (except for `allow_value`) -- but with a key twist, which is that matchers are still aware of when attributes change values, but it will only let the user know about them if the matcher in question fails. This way, the user (and I) still have a window in what is going on under the hood, but it won't disrupt what the user is trying to do normally.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,12 @@
   (formerly CouldNotSetAttributeError) when used against an attribute that is an
   enum in an ActiveRecord model.
 
+### Features
+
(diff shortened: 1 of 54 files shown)


mcmire:
The branch for this was removed too early, oops -- but this ended up being merged as 118993480604d39c73687d069f7af3726f3e3f3e and 5532f4359aa332b10de7d46f876eaffd4a95b5b6.

Issue #865, opened 2015-12-17
validate_confirmation_of does not work for integer columns


oniofchaos:
When using the confirmation validation on an integer value, the `validate_confirmation_of` matcher does not work as it appears to use string values.

mcmire:
Thanks for the report. I'm not sure when I'll get to this (a PR is encouraged), so I'll keep this open in the meantime.

kheppenstall:
Confirmed this bug. No error with strings but had the following problem with integers:

Validation:
```ruby
validates_confirmation_of :unit_price_in_cents
```
Test:
```ruby
it { is_expected.to validate_confirmation_of(:unit_price_in_cents) }
```

Error:
<img width="719" alt="screen shot 2017-04-06 at 11 40 50 am" src="https://cloud.githubusercontent.com/assets/16868275/24767929/13e62948-1abe-11e7-9a37-03077191732f.png">

mcmire:
Hey folks. In an effort to lighten our load as maintainers and be able to serve you better in the future, the shoulda-matchers team is working on cleaning out the cobwebs in this repo by pruning the backlog. As there are few of us, there are a lot of items that will simply never earn our attention in a reasonable time frame, and rather than giving you an empty promise, we think it makes more sense to focus on more recent issues. That means, unfortunately, that we must close this issue.

Don't take this the wrong way: our aim is not to diminish the effort people have made or dismiss problems that have been raised. If you feel that we should reopen this issue, then please let us know so that we can reprioritize it. Thanks!

PR #868, opened 2015-12-25
Validate confirmation of for integer values

shishir127:
Fix for #865

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
@@ -72,8 +72,8 @@ module Shoulda
       #
       # @return [ValidateConfirmationOfMatcher]
       #
-      def validate_confirmation_of(attr)
-        ValidateConfirmationOfMatcher.new(attr)
+      def validate_confirmation_of(attr, data_type = :string)
+        ValidateConfirmationOfMatcher.new(attr, data_type)
       end
 
       # @private
@@ -82,10 +82,11 @@ module Shoulda
 
         attr_reader :attribute, :confirmation_attribute
 
-        def initialize(attribute)
-          super
+        def initialize(attribute, data_type = :string)
+          super(attribute)
           @expected_message = :confirmation
           @confirmation_attribute = "#{attribute}_confirmation"
+          @data_type = data_type
         end
 
         def simple_description
@@ -94,10 +95,15 @@ module Shoulda
 
         def matches?(subject)
           super(subject)
-
-          disallows_different_value &&
-            allows_same_value &&
-            allows_missing_confirmation
+          if @data_type == :integer
+            disallows_different_integer_value &&
+              allows_same_integer_value &&
+              allows_missing_confirmation_for_integers
+          else
+            disallows_different_value &&
+              allows_same_value &&
+              allows_missing_confirmation
+          end
         end
 
         private
@@ -110,6 +116,14 @@ module Shoulda
           end
         end
 
+        def disallows_different_integer_value
+          set_confirmation(1)
+
+          disallows_value_of(0) do |matcher|
+            qualify_matcher(matcher)
+          end
+        end
+
         def allows_same_value
           set_confirmation('same value')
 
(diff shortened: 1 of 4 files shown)


spec/support/unit/record_validating_confirmation_builder.rb
```diff
@@ -6,6 +6,7 @@ class RecordValidatingConfirmationBuilder
 
     def initialize(options)
       @options = options
+      @data_type = options[:data_type].nil? ? :string : options[:data_type]
```
mcmire: Let's do this:

* I believe that I call this option `column_type` in other places, so let's rename the option to that.
* Let's move this to a method under `private`.
* Let's use `fetch` on Hash to handle the use-value-if-key-given-otherwise-use-another-value logic.

So, all together, that method would look like:

``` ruby
private

def column_type
  options.fetch(:column_type, :string)
end
```

spec/unit/shoulda/matchers/active_model/validate_confirmation_of_matcher_spec.rb
```diff
@@ -13,10 +13,13 @@
   end
 
   context 'when the model has a confirmation validation' do
-    it 'passes' do
-      builder = builder_for_record_validating_confirmation
-      expect(builder.record).
-        to validate_confirmation_of(builder.attribute_to_confirm)
+    [:string, :integer].each do |column_data_type|
+      it "passes when data type is #{column_data_type}" do
```
mcmire: Similar thing here, let's rename `column_data_type to simply `column_type`.

And then perhaps: "it passes when the column is of type #{column_type}"

spec/unit/shoulda/matchers/active_model/validate_confirmation_of_matcher_spec.rb
```diff
@@ -13,10 +13,13 @@
...
+        builder =
```
mcmire: To fit the existing style, let's write this like:

``` ruby
builder = builder_for_record_validating_confirmation(
  column_type: column_type
)
```

lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
```diff
@@ -94,7 +94,7 @@ def simple_description
 
         def matches?(subject)
           super(subject)
-
+          @data_type = @subject.class.columns_hash[@attribute.to_s].type
```
mcmire: Let's extract this to a method (called `column_type` instead) instead of making this an instance variable, I think that will ensure there's not too much clutter here.

Also, note that this assumes that the model is an ActiveRecord model, and I think we can't make that assumption, because otherwise this will fail. If you look at [other matchers](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb#L117) you'll see how we do this. Obviously, this means that we will also need a test that the confirmation matcher works with an ActiveModel model. I'm not interested in comprehensive tests right now, so a simple example should work fine ([here](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb#L67) is an example of that).

shishir127:
@mcmire I have no idea how to get rid of the HoundCI's violations. Didn't find any example at the [style guide](https://github.com/thoughtbot/guides/blob/master/style/ruby/sample.rb).

mcmire:
@shishir127 There was once an example of how to resolve those warnings in our guides, but it along the with the guideline [got removed](https://github.com/thoughtbot/guides/pull/362) a while back for some reason.

In general, here's the heuristic we tend to use:

> If you're inside of an argument list, hash, or array, indent by two spaces from the line that opened the argument list, hash, or array. (In other words, don't create a situation where you'd have to tab excessively to indent the line.)

So what that means in practice is that instead of saying:

``` ruby
this_is_a_long_method(foo,
                      bar,
                      baz)
```

you'd say:

``` ruby
this_is_a_long_method(
  foo,
  bar,
  baz
)
```

shishir127:
@mcmire Is this fix not needed anymore?

mcmire:
Sorry for the late reply on this, I've been out. I'm not sure. I'm finding it hard to remember why this fix was necessary in the first place. Can you enlighten me, please? 😬

shishir127:
It's apparently for #865, to make the validate_confirmation_of matcher work with integer values. TBH I don't remember much about this too. Do you want me to rebase and update the PR?

mcmire:
@shishir127 Yeah, if you could rebase that would definitely help.

mcmire:

PR #869, opened 2015-12-25
Typo fix

ojab:


mcmire:
@ojab Thanks for this. I imagine there is a test missing for this. Do you think you could add that?

ojab:
yep, will do tomorrow or so.
BTW can you tell me how houncl tests PRs? I assume rubocop, but not sure about settings. I'd like to add this info to the CONTRIBUTING.md.

mcmire:
@ojab Yes, Hound uses Rubocop, but we (thoughtbot) have slightly different style rules than what comes with Rubocop by default, and so this project uses those rules instead. You can see them [here](https://github.com/thoughtbot/hound/blob/master/config/style_guides/thoughtbot/ruby.yml).

ojab:
Updated PR, now there is a test.

mcmire:
Oh yeah, I forgot to mention -- some of the styles that are in Hound have changed over time, so we have a local hound.yml file in this repo to revert those style changes. That file is here: https://github.com/thoughtbot/shoulda-matchers/blob/master/.hound_config/ruby.yml. That's why you're getting a message from Hound above about single quotes. No matter -- I'll clean this up.

mcmire:
Thanks! Merged as 1dbf71bfe591f110d4a3c8e412c2c41329657ad6.

PR #876, opened 2016-01-10
add array to have_db_column.with_options

juniormesquitadandao:
Adding support for postgresql arrays data types on rails migrations.

``` ruby
  it { is_expected.to have_db_column(:phones).of_type(:string).with_options(array: true) }
```

[more](http://blog.plataformatec.com.br/2014/07/rails-4-and-postgresql-arrays/)

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
@@ -52,7 +52,7 @@ module Shoulda
       #
       # Use `with_options` to assert that a column has been defined with
       # certain options (`:precision`, `:limit`, `:default`, `:null`, `:scale`,
-      # or `:primary`, `:array`).
+      # `:primary`, or `:array`).
       #
       #     class CreatePhones < ActiveRecord::Migration
       #       def change
--- a/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
@@ -98,15 +98,17 @@ describe Shoulda::Matchers::ActiveRecord::HaveDbColumnMatcher, type: :model do
     end
   end
 
-  context 'with array option' do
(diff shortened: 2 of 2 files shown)


lib/shoulda/matchers/active_record/have_db_column_matcher.rb
```diff
@@ -52,7 +52,7 @@ module ActiveRecord
       #
       # Use `with_options` to assert that a column has been defined with
       # certain options (`:precision`, `:limit`, `:default`, `:null`, `:scale`,
-      # or `:primary`).
+      # or `:primary`, `:array`).
```
mcmire: To fit grammatically, let's change this to "`:primary`, or `:array`".

juniormesquitadandao:
Travis failed because SqlLite does not support array data type. Bypass?

mcmire:
Okay -- what if you check to see if `column` responds to `array`? If it doesn't, then have `correct_array?` return true.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
```diff
@@ -98,6 +98,18 @@
...
+      expect(with_table(:phones, :string, array: false)).
+        to have_db_column(:phones).with_options(array: false)
+    end
+
+    it 'rejects a column of wrong array' do
+      expect(with_table(:phones, :string, array: true)).
+        not_to have_db_column(:phones).with_options(array: false)
+    end
+  end
+
```
juniormesquitadandao: Good evening. I also thought about doing this, but the test will never successfully in unsupported database.

mcmire: Ah, so you'll need to wrap *this* in a conditional. There's a class-level helper, `database_supports_array_columns?` that you should be able to use.

jerryjohnjacob:
@juniormesquitadandao @mcmire Any plans to continue with this?

PR #888, opened 2016-01-19
Restoring setting attribute via __send__

### LINKED ISSUE #887: shoulda-matchers fails with private attributes
From version 3.0.1 to version 3.1.0 private attribute setting can no longer be validated via Should::Matchers. Consider the following PORO.

```ruby
class ModelWithPrivateAttribute
  include ActiveModel::Validations
  validates :name, presence: true
  def initialize(name:)
    self.name = name
  end
  attr_reader :name
  private
  attr_writer :name
end
```

The following commit broke the validation, https://github.com/thoughtbot/shoulda-matchers/commit/296211211497e624dde87adae68b385ad4cdae3a, changed the attribute tests from `__send__` to `public_send`

I believe if the following changes will resolve the issue:

* Change from `#public_send ` to `#send` on [line 83](https://github.com/thoughtbot/shoulda-matchers/blob/a4b18fcc3830d4b461ee1318e70aca20fa05a1db/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb#L83)
* Change `object.respond_to?("#{attribute_name}=")` to `object.respond_to?("#{attribute_name}=", true)` on [line 157](https://github.com/thoughtbot/shoulda-matchers/blob/a4b18fcc3830d4b461ee1318e70aca20fa05a1db/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb#L157)

jeremyf:
The following commit @296211211497e624dde87adae68b385ad4cdae3a changed
the setting of attributes from `__send__` to `public_send`. This does
not account for situations in which the setter method is private.

``` ruby
class ModelWithPrivateAttribute
  def initialize(name:)
    self.name = name
  end

  include ActiveModel::Validations
  validates :name, presence: true

  attr_accessor :name
  private :name=
end
```

Closes #887

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
@@ -37,6 +37,27 @@ Example did not properly validate that :attr cannot be empty/falsy.
     )
   end
 
+  context 'a model with a private setter method' do
+    let(:record) do
+      class ModelWithPrivateAttribute
+        include ActiveModel::Validations
+        validates :attr, presence: true
+        def initialize(options = {})
+          self.attr = options.fetch(:attr)
+        end
+        attr_accessor :attr
+        private :attr=
+      end.new(:attr => 'hello')
+    end
+    it 'will not throw an exception' do
+      assertion = lambda do
+        expect(record).to matcher
+      end
+
+      expect(&assertion).to_not raise_error
+    end
+  end
+
   context 'a model without a presence validation' do
     it 'rejects with the correct failure message' do
       record = define_model(:example, attr: :string).new
--- a/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb
+++ b/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb
@@ -80,7 +80,7 @@ module Shoulda
           end
 
           def set
-            object.public_send("#{attribute_name}=", value_written)
+            object.__send__("#{attribute_name}=", value_written)
             after_set_callback.call
 
             @result_of_checking = successful_check
@@ -154,7 +154,7 @@ module Shoulda
             if active_resource_object?
               object.known_attributes.include?(attribute_name.to_s)
             else
-              object.respond_to?("#{attribute_name}=")
+              object.respond_to?("#{attribute_name}=", true)
             end
           end
 


mcmire:
I can add this, but do you think it makes sense to validate an attribute that's private? If the name is supposed to be set, and the only way the attribute is set is by calling `.new`, then why not have a manual check in the initializer to ensure that `name` is something that is not blank?

jeremyf:
@mcmire the change from `__send__` to `public_send` is a backwards compatability breaking change and at a minimum should be part of a major version update (though I understand the challenges of tracking these changes so I'm not begruding the updates).

Regarding public setter method vs. private setter, the specific use case is a custom Ruby form object distilled to the code written below. I want to validate the user's input that are passed via parameters. So this is a stylistic choice; I'm preferring to not expose setter methods of the form object so that I'm exposing the smallest interface of that object.

```ruby
class CustomForm
  def initialize(current_user:, user_input: {})
   self.current_user = current_user
   self.name = user_input.fetch(:name, nil)
   self.role = user_input.fetch(:role) { default_role }
  end
  attr_reader :current_user, :name, :role
  private
  attr_writer :name, :current_user, :role
  include ActiveModel::Validations
  validates :name, presence: true

  def default_role
    # Some business logic
  end
end
```

mcmire:
Cool, thanks for the reply, and that use case makes sense.

jeremyf:
@mcmire And thank you for being one of the maintainers of a fantastic and useful testing library.

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -37,6 +37,27 @@
     )
   end
 
+  context 'a model with a private setter method' do
+    let(:record) do
```
mcmire: The convention in these test files is to use builder methods instead of `let`s. Namely, we often have a pair of methods: `define_model_...`, which defines and returns a model class, and `build_record_...`, which returns an instance of that model. See the bottom of this file for an example of the `create_model` / `build_record` pattern (although [the tests for the numericality matcher](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb#L1789) might be a slightly better example, since the methods are actually named consistently there).

There are a few different helper methods you can use to make models in tests, but the one that's applicable here is called `define_active_model_class`, and you can see an example of how it's used in this file on [line 328](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb#L328).

With these two methods in place, then you can write something like:

``` ruby
it 'will not throw an exception' do
  # here we are assuming you've made a method called `build_record_with_private_attribute`
  record = build_record_with_private_attribute
  assertion = -> { expect(record).to matcher }
  expect(&assertion).not_to raise_error
end
```

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -37,6 +37,27 @@
...
+        validates :attr, presence: true
+        def initialize(options = {})
+          self.attr = options.fetch(:attr)
+        end
+        attr_accessor :attr
+        private :attr=
+      end.new(:attr => 'hello')
+    end
+    it 'will not throw an exception' do
+      assertion = lambda do
```
mcmire: As indicated above, you can probably put this on one line.

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -37,6 +37,27 @@
...
+        expect(record).to matcher
+      end
+
+      expect(&assertion).to_not raise_error
```
mcmire: As indicated above, we use `not_to` instead of `to_not`.

mcmire:
Thanks :blush: 

Just as an FYI, in order to fully support this, I think we will need to add tests to every single validation matcher, since it affects all of them, not just the presence matcher. This is more difficult than it should be, as there is variation between how the tests for each matcher builds records. I've made some comments for your changes to the presence matcher tests, so you can apply those, but I'll come in after and add the remaining tests and then merge this when I'm done.

jeremyf:
@mcmire I was noting that there are no unit tests against the class that does the `#set` on the attributes. A unit test on that class would ease the pain of validators that collaborate with it.

mcmire:
@jeremyf That is true, but you'd still need integration tests to ensure that this test case works in the context of every validation matcher. (If we refactored the code in `allow_value` not to use AttributeSetter anymore for some reason, then those tests would go away.)

PR #897, opened 2016-02-08
Require ActiveSupport's `try` core extension

ramontayag:
Usage of `try` throughout shoulda-matchers is broken when this file is not required


mcmire:
Sorry for the late reply on this. Are you using shoulda-matchers in a particular environment? In other words, is there a reason why all of ActiveSupport isn't being required ahead of time?

mcmire:
I don't have enough to go on here, so I'm closing this. Feel free to respond if this is still an issue.

Issue #495, opened 2014-04-16
no implicit conversion of nil into Array using permit


shedd:
Seems to be another issue for 2.6.0

Following on from my update to #482

When my strong parameters method is defined as follows (using fetch rather than require):

```
def order_params
  # use params.fetch rather than params.require to handle cases where params are empty
  params.fetch(:order, {}).permit(:eta, :diner_id)
end
```

I run into the following failure:

```
4) Restaurant::OrdersController POST create should permit POST #create to receive parameters :eta and :diner_id
   Failure/Error: it { should permit(:eta, :diner_id).for(:create) }
   TypeError:
     no implicit conversion of nil into Array
   # /usr/local/rvm/gems/ruby-2.1.1/gems/awesome_print-1.2.0/lib/awesome_print/core_ext/array.rb:20:in `-'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/awesome_print-1.2.0/lib/awesome_print/core_ext/array.rb:20:in `call'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/awesome_print-1.2.0/lib/awesome_print/core_ext/array.rb:20:in `block (2 levels) in <class:Array>'
   # /usr/local/rvm/gems/ruby-2.1.1/bundler/gems/shoulda-matchers-8f933f80647e/lib/shoulda/matchers/action_controller/strong_parameters_matcher.rb:83:in `parameters_difference'
   # /usr/local/rvm/gems/ruby-2.1.1/bundler/gems/shoulda-matchers-8f933f80647e/lib/shoulda/matchers/action_controller/strong_parameters_matcher.rb:52:in `failure_message'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/rspec-expectations-2.14.5/lib/rspec/expectations/handler.rb:30:in `handle_matcher'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/rspec-core-2.14.8/lib/rspec/core/memoized_helpers.rb:68:in `should'
   # ./spec/controllers/restaurant/orders_controller_spec.rb:46:in `block (3 levels) in <top (required)>'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `instance_eval'
   # /usr/local/rvm/gems/ruby-2.1.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `block in run'
[...]

PR #899, opened 2016-02-08
Using #fetch with #permit isn't fully tested

aripollak:
The fix for #495 doesn't seem to work currently. There's a test case for it, but it didn't test the full behavior, so I fixed that and now it fails as expected. Unfortunately I haven't figured out what the appropriate fix is for this, so some guidance would be appreciated.

### DIFF
--- a/spec/unit/shoulda/matchers/action_controller/permit_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/action_controller/permit_matcher_spec.rb
@@ -187,7 +187,7 @@ describe Shoulda::Matchers::ActionController::PermitMatcher, type: :controller d
 
     expect(controller).
       to permit(:eta, :diner_id).
-      for(:create)
+      for(:create, params: { order: { foo: '' } })
   end
 
   it 'works with #update specifically' do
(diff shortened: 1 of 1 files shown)


aripollak:
Oops, it turns out the matcher does work as expected, but the test case was missing params to get it to work, since the regular empty hash returned by `fetch` was not instrumented. That's fixed up now.

spec/unit/shoulda/matchers/action_controller/permit_matcher_spec.rb
```diff
@@ -180,6 +180,16 @@ def params_with_conditional_require(params, *filters)
...
   end
 
+  it 'works when #fetch is used instead of #require (issue #495)' do
+    define_controller_with_strong_parameters(action: :create) do
+      params.fetch(:order, {}).permit(:eta, :diner_id)
+    end
+
+    expect(controller).
+      to permit(:eta, :diner_id).
+      for(:create, params: { order: { foo: '' } })
```
mcmire: Thanks for this -- this is definitely a more realistic way to test this. You're using `foo` as a key just so the `order` hash is non-empty, is that correct?

aripollak: Yeah.. Ideally that shouldn't need to happen for this, but I'm not sure why it's needed. I get why it's needed when using `require`, but not `fetch`.

mcmire:

PR #902, opened 2016-02-11
Fix permit matcher for multiple instances of params

aripollak:
When the permit matcher is used without `#on`, the controller does not use `params#require`, and the params object gets duplicated, the matcher did not recognize the `#permit` call inside the controller. This happened because the matcher overwrote double registries with the same parameter hash whenever ActionController::Parameters was instantiated. The duplication happens unintentionally when passing the params to a model's method that ends up calling `#assign_attributes`, like `#new`.
This is related to #899.

Here's the part of the call stack that was doing the duplication in Rails:

```
/home/ari/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activesupport-4.2.5.1/lib/active_support/hash_with_indifferent_access.rb:190:in `dup'
/home/ari/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/actionpack-4.2.5.1/lib/action_controller/metal/strong_parameters.rb:438:in `dup'
/home/ari/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activesupport-4.2.5.1/lib/active_support/hash_with_indifferent_access.rb:232:in `stringify_keys'
/home/ari/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/protected_attributes-1.1.3/lib/active_record/mass_assignment_security/attribute_assignment.rb:51:in `assign_attributes'
/home/ari/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/protected_attributes-1.1.3/lib/active_record/mass_assignment_security/persistence.rb:64:in `block in update'
```

### DIFF
--- a/lib/shoulda/matchers/action_controller/permit_matcher.rb
+++ b/lib/shoulda/matchers/action_controller/permit_matcher.rb
@@ -336,7 +336,7 @@ module Shoulda
         # @private
         class CompositeParametersDoubleRegistry
           def initialize
-            @parameters_double_registries_by_params = {}
+            @parameters_double_registries = []
           end
 
           def register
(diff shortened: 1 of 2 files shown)


mcmire:
> The duplication happens unintentionally when passing the params to a model's method that ends up calling `#assign_attributes`, like `#new`.

I know it's been a little while but can you explain this further?

aripollak:
If I'm remembering this correctly, the matcher was failing when doing something like this in the controller:
```ruby
@model = Model.new(params.fetch(:model, {}).permit(:foo))
@model.update(params.fetch(:model, {}).permit(:foo))
```

mcmire:
Okay, thanks.

The other thought I had while reading your PR was that there was a reason why I was using a hash instead of an array to store doubles for ActionController::Parameters. I can't remember why I did that off-hand, though. I'll have to take a closer look at that.

geoffharcourt:
Merging this after testing.

mcmire:
@geoffharcourt Woohoo, thanks!

PR #910, opened 2016-03-02
Allow to use the matchers with just rspec expectations gem

majioa:
Allow to use the matchers for the just rspec expectations, apart the whole

RSpec.
- It adds the specific test framework that just includes the specified
  libraries directly to RSpec::Matchers.

The addition is covered with an acceptance test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,30 @@ describe Person do
 end
 ```
 
+### RSpec-Expectations
+
+Use this instead of the RSpec ifself in case you just use the specific module gem
+'rspec-expectations' to test webpart(s).
+
+Include `shoulda-matchers` in your Gemfile:
+
+``` ruby
+group :test do
+  gem 'shoulda-matchers'
+end
+```
+
+Then, configure the gem to integrate with RSpec specifying the test framework as `:rspec_exp`:
+
+    Shoulda::Matchers.configure do |config|
+      config.integrate do |with|
+        with.test_framework :rspec_exp
+        # ...
+      end
+    end
+
+Now you can use matchers in your tests as for RSpec part.
+
 ### Minitest
 
 Shoulda Matchers was originally a component of [Shoulda][shoulda], a gem that
--- a/lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
+++ b/lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
@@ -1,0 +1,34 @@
+module Shoulda
+  module Matchers
+    module Integrations
+      module TestFrameworks
+        # @private
+        class RspecExpectations
+          Integrations.register_test_framework(self, :rspec_exp)
+
+          def validate!
+            return if defined? RSpec::Matchers
+            raise TestFrameworkNotFound, <<-EOT
+You need to include the 'rspec-expectations' gem to Gemfile, and add
+the line before requiring the shoulda matchers:
+
+require 'rspec/expectations'
+EOT
+          end
+
+          def include(*modules, **_options)
+            RSpec::Matchers.send(:include, *modules)
+          end
+
+          def n_unit?
+            false
+          end
+
+          def present?
+            true
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/acceptance/rspec_expectation_test_framework_integration_spec.rb
+++ b/spec/acceptance/rspec_expectation_test_framework_integration_spec.rb
@@ -1,0 +1,59 @@
+require 'acceptance_spec_helper'
+
+describe 'shoulda-matchers integrates libs for rspec-expectations framework' do
+  before do
+    create_rails_application
+
+    write_file 'db/migrate/1_create_users.rb', <<-FILE
+      class CreateUsers < ActiveRecord::Migration
+        def self.up
+          create_table :users do |t|
+            t.string :name
+          end
+        end
+      end
+    FILE
+
+    run_rake_tasks!(*%w(db:drop db:create db:migrate))
+
+    write_file 'app/models/user.rb', <<-FILE
+      class User < ActiveRecord::Base
+        validates_presence_of :name
+        validates_uniqueness_of :name
+      end
+    FILE
+
+    append_rake_task 'rspec_exp', 'environment', <<-CODE
+  require 'rspec/expectations'
+  require 'shoulda-matchers'
+  require_relative 'test/test_helper.rb'
+
+  def expect(value, &block)
+    ::RSpec::Expectations::ExpectationTarget.for(value, block)
+  end
+
+  def validate_presence_of(value)
+    ::ActiveSupport::TestCase.validate_presence_of(value)
+  end
+
+  it = User.create name: 'Vasja'
+  expect(User.first).to validate_presence_of(:name)
+  puts "Passed"
+    CODE
+
+    updating_bundle do
+      add_rspec_expectations_to_project!
+      add_shoulda_matchers_to_project(
+        test_frameworks: [:rspec_exp],
+        libraries: [:active_record, :active_model],
+      )
+    end
+  end
+
+  context 'when using both active_record and active_model libraries' do
+    it 'allows the use of matchers from both libraries' do
+      result = run_rake_tasks 'rspec_exp'
+      expect(result).to have_output('Passed')
+    end
+  end
+end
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -197,6 +197,30 @@ describe Person do
 end
 ```
 
+### RSpec-Expectations
+
+Use this instead of the RSpec ifself in case you just use the specific module gem
```
TheTeaNerd: I think the English can be improved here, I was a little confused by it. I suggest:

```
`shoulda-matchers` can be configured to support using _only_ the `rspec-expectations` gem instead of the full RSpec suite.

Firstly, include `shoulda-matchers` in your Gemfile:
```

...it was much harder than I expected to find a good starting point here. Could you perhaps give a one-liner for why you want to use _only_ RSpec-Expectations and not the full suite? Speed? Load-time? Fewer dependencies?

README.md
```diff
@@ -197,6 +197,30 @@ describe Person do
...
+group :test do
+  gem 'shoulda-matchers'
+end
+```
+
+Then, configure the gem to integrate with RSpec specifying the test framework as `:rspec_exp`:
+
+    Shoulda::Matchers.configure do |config|
+      config.integrate do |with|
+        with.test_framework :rspec_exp
```
TheTeaNerd: I think it would be more expressive / more obvious to use `rspec_expectations` instead of `rspec_exp`.

README.md
```diff
@@ -197,6 +197,30 @@ describe Person do
...
+
+``` ruby
+group :test do
+  gem 'shoulda-matchers'
+end
+```
+
+Then, configure the gem to integrate with RSpec specifying the test framework as `:rspec_exp`:
+
+    Shoulda::Matchers.configure do |config|
```
TheTeaNerd: Missing a code block here?

README.md
```diff
@@ -197,6 +197,30 @@ describe Person do
...
+      config.integrate do |with|
+        with.test_framework :rspec_exp
+        # ...
+      end
+    end
+
+Now you can use matchers in your tests as for RSpec part.
```
TheTeaNerd: I think this line is redundant and could be removed for simplicity.

lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
```diff
@@ -0,0 +1,34 @@
...
+  module Matchers
+    module Integrations
+      module TestFrameworks
+        # @private
+        class RspecExpectations
+          Integrations.register_test_framework(self, :rspec_exp)
+
+          def validate!
+            return if defined? RSpec::Matchers
+            raise TestFrameworkNotFound, <<-EOT
```
TheTeaNerd: Unless there's a need to explicitly support Ruby < 2.3, this would be nicer with a 'squiggly heredoc': https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

README.md
majioa: usage with only cucumber )

README.md
majioa: probably

README.md
majioa: could the `exp` statement be understood some else than `expectations` in context of rspec?

lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
majioa: ok

README.md
majioa: here is indent 4 step

README.md
```diff
@@ -197,6 +197,30 @@ describe Person do
 end
 ```
 
+### RSpec-Expectations
```
mcmire: This would be renamed to "Cucumber", and this whole section would talk about that instead of mentioning `rspec-expectations`.

README.md
mcmire: Going along with the use case, instead of `with.test_framework :rspec_exp`, this would be `with.test_framework :cucumber`.

lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
```diff
@@ -0,0 +1,34 @@
+module Shoulda
+  module Matchers
+    module Integrations
+      module TestFrameworks
+        # @private
+        class RspecExpectations
```
mcmire: Instead of RspecExpectations, this would be Cucumber, and everything in this file would talk about that rather than `rspec-expectations`.

lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
mcmire: I'm fine with this -- this makes it easier to wrap the message at 80 characters.

lib/shoulda/matchers/integrations/test_frameworks/rspec_expectations.rb
```diff
@@ -0,0 +1,34 @@
...
+            raise TestFrameworkNotFound, <<-EOT
+You need to include the 'rspec-expectations' gem to Gemfile, and add
+the line before requiring the shoulda matchers:
+
+require 'rspec/expectations'
+EOT
+          end
+
+          def include(*modules, **_options)
+            RSpec::Matchers.send(:include, *modules)
```
mcmire: So now, instead of including modules into RSpec::Matchers (which requires understanding that RSpec::Matchers is itself mixed into Object), this could be `World(*modules)` (which is a public API).

spec/acceptance/rspec_expectation_test_framework_integration_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'acceptance_spec_helper'
+
+describe 'shoulda-matchers integrates libs for rspec-expectations framework' do
```
mcmire: This file would be renamed/reframed to represent Cucumber and would set up a Cucumber project.

spec/support/acceptance/adds_shoulda_matchers_to_project.rb
```diff
@@ -118,6 +119,10 @@ def integrates_with_rspec?(test_framework)
       test_framework == :rspec
     end
 
+    def integrates_with_rspec_expectations?(test_framework)
```
mcmire: This would be `integrates_with_cucumber?`. You'd have to modify `test_helper_files_for` to do:

``` ruby
if integrates_with_cucumber?(test_framework)
  files << 'features/support/env.rb'
end
```

spec/acceptance/rspec_expectation_test_framework_integration_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    run_rake_tasks!(*%w(db:drop db:create db:migrate))
+
+    write_file 'app/models/user.rb', <<-FILE
+      class User < ActiveRecord::Base
+        validates_presence_of :name
+        validates_uniqueness_of :name
+      end
+    FILE
+
+    append_rake_task 'rspec_exp', 'environment', <<-CODE
```
mcmire: Instead of appending to a rake task like this, I would actually write this to a file. In the case of Cucumber this could be `features/cucumber_integration.feature` or whatever you want to call it. Unless that's not how you actually write your tests? I'm a bit confused here. Please enlighten me :)

majioa:
@mcmire If we suppose that the case is limited by cucumber case  that should be true, but  because the cucumber is the only explicit example, and may be someone will want to use expectations' syntax without cucumber, but on the same way, for example on test-unit replacing assert syntax. That is why I just used the common case called rspec-exp/rspec-expectations to describe the feature. Well it may be not so correct naming convention for the case, but cucumber is also not a proper naming. May be it should have name expectation-commons?

mcmire:
@majioa My sense is that making a more generic adapter would be premature abstraction. If someone else comes forward later and says that they want to use RSpec syntax with Test::Unit down the line, then I would take that into consideration, but you're literally the only person who's proposed anything like this, and I don't want to confuse the 99% of the userbase who is perfectly happy using `shoulda-matchers` with the full RSpec experience.

That said, can you provide an example of the type of test that you write? I'm not still clear on how you plan on using this.

mcmire:

majioa:
@mcmire  https://github.com/majioa/shoulda-matchers-cucumber

mcmire:
Oh very nice! I've added your extension to the README: https://github.com/thoughtbot/shoulda-matchers#extensions

PR #921, opened 2016-04-20
test against ruby 2.3

lucianosousa:


.travis.yml
```diff
@@ -11,6 +11,7 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
+  - 2.3
```
aripollak: Looks like this shortened version isn't supported on Travis, at least as of the time this got pushed.

lucianosousa: @aripollak updated. thanks for notice

guialbuk:
Superseded by #1027

PR #924, opened 2016-05-11
Include spec type inference in README

supremebeing7:
I'm adding this gem to an older project, and this trolled me for a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,14 @@ different levels where you can use these matchers:
   tagged with `type: :routing` or in files located under `spec/routing`.
 * Independent matchers are available in all example groups.
 
+**In order for rspec to infer the spec type from the spec file location,** ensure you have this in your `spec_helper.rb`:
+
+``` ruby
+RSpec.configure do |config|
+  config.infer_spec_type_from_file_location!
+end
+```
+
 **If you are using ActiveModel or ActiveRecord outside of Rails** and you want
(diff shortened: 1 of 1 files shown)


mcmire:
Sorry about this. We're not trying to troll people, honest! This setting is [on by default](https://github.com/thoughtbot/suspenders/blob/master/templates/rails_helper.rb#L18) in all our projects, so we're admittedly a little blind to the issue.

This isn't the first time this has come up. In particular #835 offers a different viewpoint, which is that we should update the README to provide the suggestion that `type: :suchandsuch` should be affixed to the end of `describe` blocks. Although this is super lame, I'm now starting to think that this is a better solution. I'll leave this open in the meantime, though.

supremebeing7:
Having it on by default has been my experience so far as well, which is probably why I ran into this trouble. I noticed that running the `rspec-rails` install generator appears to include that line in the [rails_helper](https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec/install/templates/spec/rails_helper.rb#L60), so if there weren't others having this issue, I would have almost said the README update isn't really necessary. 

Our team writes our specs the same as yours, using `describe Model do` rather than the more verbose `Rspec.describe Model, type: :model do`, and I think there's value in pointing out how to preserve that simpler syntax by ensuring the `infer_spec_type_from_file_location!` line is in the relevant helper.

Just my two cents. Thanks for considering. 😃

Issue #912, opened 2016-03-09
PR #829 breaks "define_enum_for" matcher for enums that do not actually use integer columns


begedin:
Say we want a status enum that looks like 

``` Ruby
enum status: { active: "active", pending: "pending", rejected: "rejected" }
```

With the changes made in PR #829, we can no longer spec for that enum with

``` Ruby
it { should define_enum_for(:status).with(active: "active", pending: "pending", rejected: "rejected")
```

Are active record enums with strings as underlying column types not allowed? I couldn't find any information that would explicitly make it so.

I believe the matcher should only enforce integer type if the provided expected enum value is an array, or if it's a hash with values all being integer.

guilhermearaujo:
I agree. There is absolutely no reason to enforce integers.

It started with #827, but in the author's example, he is storing an `:integer` in a column of type `:string`, and his enum maps the keys to integers. If he wanted to use integers, he could just use an array of symbols.

Had he done that, than yes, the matcher could enforce integer type in case of array, or enforce nothing otherwise.

lucasgomide:
cc: @mcmire 

can you review this ?

mcmire:
Sorry for the lack of reply on this, I've been out.

I agree that this is a bug. I'm not sure why we had the matcher assume integers except that that was the original use case.

I can't say when I'll get to this, but I've at least seen it. :)

tiii:
I don't think this is a bug. Using strings as keys is not the way to go with ActiveRecord::Enum. There is no mention about using something else than integer columns and integer values in the official documentation (4.1.8: http://apidock.com/rails/ActiveRecord/Enum | 5.0.0: http://api.rubyonrails.org/classes/ActiveRecord/Enum.html ).

If you use a string type column and non string enum values (integers or symbols) you lose the ability to retrieve the key by calling the getter method on your model (demonstrated here: https://github.com/thoughtbot/shoulda-matchers/issues/827#issuecomment-152551748 )

In my opinion rails should enforce the use of an integer column, but as long as rails does not I don't mind that should-matchers enforces it.

If we still want to allow strings as keys we should check the enum values too:
Integer column: only allow integer enum values
string column: only allow string enum values

guilhermearaujo:
(@tiii I suppose that by 'key' in your message you mean 'value', as in the value held by the key being a string, and not the key itself. Is that correct?)

[...]

PR #932, opened 2016-06-09
Fix enums with hashes

guilhermearaujo:
Because if you are using a hash instead of a list, you probably should not be using integers as values.

Fixes issue #912


brust:
Any response about this PR?

guilhermearaujo:
@mcmire any update on this?

mcmire:
This is good, but in order to complete this PR, we need to add a test to confirm that when you have a column that isn't an integer but a string, the matcher won't fail. Do you think you could add that?

PR #934, opened 2016-06-13
ActiveModel: clear errors before calling 'valid?'

samstickland:
Some form libraries, such as Reform, can use ActiveModel
validations but don't clear errors between validations (like the
ActiveModel + ActiveRecord combination does).

This change manually clears the errors before calling valid? allowing
matchers that call valid? multiple times (such as validate_length_of) to
be used with Reform and Trailblazer contracts.

### DIFF
--- a/lib/shoulda/matchers/active_model/validator.rb
+++ b/lib/shoulda/matchers/active_model/validator.rb
@@ -90,6 +90,7 @@ module Shoulda
         end
 
         def perform_validation
+          record.errors.clear
           if context
             record.valid?(context)
           else


mcmire:
If we're going to support Reform, then we'll need an integration test to cover this. Do you think you can take a crack at adding one?

samstickland:
Sure, I can do that. Although I need to tell you that the `valid?` method will be going away in future versions of Reform, leaving only `validate`, and I'm not sure what implementation will be required to fix this - see https://github.com/apotonick/reform/issues/156#issuecomment-225601149.

If setting values directly on the form object continues to be supported in Reform then I _think_  it could be as simple as calling `record.validate({})` rather than `record.valid?` in Validator. In which case, would it be acceptable to do this?

```
def perform_validation
  if record.is_a?(Reform::Form)
    record.errors.clear
    record.validate({})
  elsif context
    record.valid?(context)
  else
    record.valid?
  end
  ...
end
```

mcmire:
I don't think it'd be bad to clear errors in every case. I also don't want to explicitly hardcode Reform::Form in the code as I think that's not as future-looking as it could be. ActiveModel also has a `validate` method (that's an alias for `valid?`), so what we could do is just use that from now on.

With that in mind, what do you think about this?

``` ruby
def perform_validation
  record.errors.clear

  if record.method(:validate).arity > 0 && context
   record.validate(context)
  else
    record.validate
  end
end
```

samstickland:
Yeah, that was the bit that worried me, hardcoding Reform::Form.. However, the Reform validate method currently requires a hash (this is usually how values are provided to Reform, e.g. `form.validate(params[:artist])`). So I think the two options are:

1. Alter the Validator to accept a validation strategy... This sounds like quite extensive changes
2. Alter Reform so that it continues to present an AM compatible API after `valid?` goes away. This could be either:
a) Setting a default value of {} in `validate`
b) Re-implementing `valid?` as `def valid?(params = {}); validate(params); end;`. (Currently `valid?` in Reform is actually a slightly different method to `validate`)

Assuming the maintainer or Reform agrees I think either 2a or 2b is the way to go. What do you think?

mcmire:
Ohh, I forgot that Reform's #validate method is different. I would hesitate to monkey-patch any instances of Reform::Form to re-implement #valid?, if that's what you were thinking. So, in that new light, I think it would be acceptable to hardcode a reference to Reform::Form for now, and then perhaps we can revisit this when the change you mentioned gets into Reform.

samstickland:
Hi,

It turns out that Reform isn't designed to support multiple validations on
a single instance. Errors.clear works for now but isn't supported and will
be going away in the future.

So, for the immediate future I'm just going to proceed with a fork or
monkey patch of shoulda, but going forwards this is going to require some
more thought. Things to consider are:

1) How to reinstaniate / duplicate the Reform object between validation
tests.
2) Calling validate instead of valid?

Sam
El 14 jun. 2016 8:05 PM, "Elliot Winkler" <notifications@github.com>
escribió:

Ohh, I forgot that Reform's #validate method is different. I would hesitate
to monkey-patch any instances of Reform::Form to re-implement #valid?, if
that's what you were thinking. So, in that new light, I think it would be
acceptable to hardcode a reference to Reform::Form for now, and then
perhaps we can revisit this when the change you mentioned gets into Reform.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/thoughtbot/shoulda-matchers/pull/934#issuecomment-225983941>,
or mute the thread
<https://github.com/notifications/unsubscribe/AJfZUzjc_IaQo28P-s_KhD5ln-ikLUa8ks5qLvtggaJpZM4I0Xhv>
.

PR #941, opened 2016-06-23
Fix weird error message on should_not validate_inclusion_of(:attribute) fixes #939

### LINKED ISSUE #939: Weird error message on should_not validate_inclusion_of(:attribute)
Scenario

> Model
> - validates :attribute, inclusion: { in: 1940..Date.today.year, allow_nil: true}, on: :update

And on test file

> model_spec.rb
> it { should_not validate_inclusion_of(:birth_year).on(:create) }

After I run the test, it passes, but I get a weird output on console using doc format

> example at ./spec/models/model_spec.rb:<a line number> (Got an error when generating description from matcher: NoMethodError: undefined method `many?' for nil:NilClass -- <path to gems on rvm>/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb:365:in`simple_description')

For me, looks like a simple fix, but I couldn't find the proper place for the fix. If you could guide me to it, I can send a PR for this too.

gmmcal:
@mcmire as we talked on #939, here's my fix to that message. Test for failure is pending, but you said you could handle it.

My tests were (with my current application)

> it { should validate_inclusion_of(:atribute) } fails with message `should have range or array to validate`
> it { should_not validate_inclusion_of(:atribute) } passes with message `should have range or array to validate`

Before the change, the first case failed and the second passes with the following message

> example at ./spec/models/model_spec.rb:<a line code> (Got an error when generating description from matcher: NoMethodError: undefined method `many?' for nil:NilClass -- <gems path>/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb:365:in`simple_description') (FAILED - 1)

I thought on raising an error to that, but it doesn't make sense to raise an error on `should_not validate_inclusion_of(:atribute)` without range, since the intention is to NOT validate inclusion of something

### DIFF
--- a/lib/shoulda/matchers/active_model/errors.rb
+++ b/lib/shoulda/matchers/active_model/errors.rb
@@ -47,6 +47,23 @@ Please use a #{record_name} with an empty `password` instead.
           model_name.humanize.downcase
         end
       end
+
+      # @private
+      class NoRangeOrArrayDefinedForInclusionError < Shoulda::Matchers::Error
+        def self.create(attribute)
+          super(attribute: attribute)
+        end
+
+        attr_accessor :attribute
+
+        def message
+          <<-EOT.strip
+You are using the `validate_inclusion_of` matcher without specifying the values that the matcher should use to test the inclusion validation. You'll need to provide the matcher with either a range or an array. For instance:
+  should validate_inclusion_of(:#{attribute}).in_range(0..10)
+  should validate_inclusion_of(:#{attribute}).in_array(['one', 'two', 'three'])
+          EOT
+        end
+      end
     end
   end
 end
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -359,7 +359,7 @@ EOT
           if @range
             "validate that :#{@attribute} lies inside the range " +
               Shoulda::Matchers::Util.inspect_range(@range)
-          else
+          elsif @array
             description = "validate that :#{@attribute}"
 
             if @array.many?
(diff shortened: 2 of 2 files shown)


mcmire:
The thing is, I don't know if it makes sense to have `it { should_not validate_inclusion_of(:attribute) }` pass either. The problem here is that the matcher can't really do anything without more instructions, so whether you use `should` or `should_not` the matcher has to abort.

Instead of setting a failure message to cause it to fail, why don't we have it raise an exception instead? Some of the matchers do this already -- for instance, [`validate_uniqueness_of`](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L501).

gmmcal:
@mcmire will go on that direction then. thanks!

gmmcal:
@mcmire just pushed the changes. can you check again? :D

lib/shoulda/matchers/active_model/errors.rb
```diff
@@ -47,6 +47,23 @@ def record_name
...
+          super(attribute: attribute)
+        end
+
+        attr_accessor :attribute
+
+        def message
+          <<-EOT.strip
+You have specified that your model's #{attribute} should/should not ensure inclusion of an array or range.
+However, the test code does not provide a proper range or array to validate against.
+Please, add `in_range` or `in_array` to your test and try again.
```
mcmire: This is a good message but I think we can do a little better. What do you think about this?

```
You are using the `validate_inclusion_of` matcher without specifying the values that the matcher should use to test the inclusion validation. You'll need to provide the matcher with either a range or an array. For instance:

    should validate_inclusion_of(:#{attribute}).in_range(0..10)
    should validate_inclusion_of(:#{attribute}).in_array(['one', 'two', 'three'])
```

gmmcal:
@mcmire I was finally able to update this. sorry for the delay

PR #948, opened 2016-07-12
Fix scope existence check in uniqueness spec 

TheSmartnik:
When testing for  a  scoped attribute uniqueness, scope existence validation sometimes fails.
It seems that https://github.com/thoughtbot/shoulda-matchers/commit/4649b60a154452e1983a612326e3b9868757ea40 was aiming to fix similar problem, but have forgotten to fix scope existence check

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -548,18 +548,13 @@ module Shoulda
         end
 
         def validate_attribute_present_on_model?
-          if attribute_present_on_model?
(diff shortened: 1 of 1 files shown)


mcmire:
Ah, great catch.

Do you think you could add a test for this?

TheSmartnik:
@mcmire Hi, sorry for not answering sooner.

Unfortunately, I don't really know how to reproduce this bug as in `shoulda-matchers` specs `model.method_defined?("#{attr}=")` always returns `true`.

To be completely honest, I assumed that if 4649b60a154452e1983a612326e3b9868757ea40 had no specs, this one may have none as well.

Anyway, I would happily add test with a little help, maybe?

mcmire:

Issue #870, opened 2015-12-27
Rails 5: belongs_to :class, optional: true 


stephanngamedev:
`belongs_to` will now trigger a validation error by default if the association is not present.
We can turn this off on a per-association basis with `optional: true`.
https://github.com/rails/rails/pull/18937

We will need a qualifier for this.

mcmire:
Okay -- do you plan on using `optional: true`? We tend to add qualifiers on an as-needed basis.

stephanngamedev:
Yes. I'm developing a app where a Amiibo may or may not to belongs to Series. ex.: Sonic amiibo belongs to Smash Bros. Series, but Wolf Link amiibo belongs to none.

mcmire:
Cool. I'm not sure when I'll be able to get to this, as I am working on resolving some regressions with the 3.0 release right now, but I would happily take a PR, otherwise I'll get to it at some point down the line. I'll leave this open in the meantime since it's related to Rails 5.

thomasbiddle:
I have a use-case for this as well.

leoebrown:
I also have a use case for this. A food database where most of the foods are not created by users, but users can create foods and 'own' the foods that they create. Workaround: the admin user profile will own foods by default. Please correct me if there's a better workaround or existing solution.

stephanngamedev:
You can use ` belongs_to :user, optional: true` for specific case or just put on application.rb `config.active_record.belongs_to_required_by_default = false` to config all application

leoebrown:
@stephanngamedev I did neither of those things and it seems like I can create foods with user_id: nil no problem--am I missing something? My migration:

```
class AddUserToFoods < ActiveRecord::Migration
  def change
    add_reference :foods, :user, index: true, foreign_key: true
  end
end
```

stephanngamedev:
Are you using Rails 5?

leoebrown:
I am using Rails 4. Sorry, I see this thread is labeled Rails 5. Would your recommendations be necessary in Rails 5?

stephanngamedev:
Yes. My recomendations is to make rails 5 belongs_to behavior like rails 4 belongs_to

mcmire:
[...]

Issue #861, opened 2015-12-11
[Question] Working with belongs_to :relation, required: true


rennanoliveira:
If you have something like

``` ruby
class Group < ActiveRecord::Base
  belongs_to :manager, required: true
end
```

Currently, what I can do on my specs is:

``` ruby
it { is_expected.to belong_to(:manager) }
it { is_expected.to validate_presence_of(:manager) }
```

Which does cover it for me, but I'm curious as to ideas on having something in the likes of:

``` ruby
it { is_expected.to belong_to(:manager).with_required(true) }
```

If it's in the docs, I apologize, I couldn't find it

mcmire:
I believe this may be a newer option that we haven't added yet (I haven't heard of it).

That would be a good idea to add (although I'd have the qualifier be `required(true)` instead of `with_required(true)`).

ojab:
In rails 5 `required: true` will be default option for `belongs_to`, so it makes sense to make it default in shoulda-matchers too.
See also #870.

zavan:
I'm using Rals 5.0.0.beta3 with RSpec 3.5.0.beta1 and shoulda-matchers 3.1.1

Consider:

```ruby
class User < ApplicationRecord
  has_many :changesets, inverse_of: :user, dependent: :restrict_with_error
end

class Changeset < ApplicationRecord
  belongs_to :user, inverse_of: :changesets
end

RSpec.describe Changeset, type: :model do
  subject { create :changeset }

  it { is_expected.to belong_to(:user).inverse_of(:changesets) }
  it { is_expected.to validate_presence_of :user }
end
```

Rails changed the "required" balongs_to validation message.
The ` it { is_expected.to validate_presence_of :user }` assertion fails:

```
  1) Changeset should validate that :user cannot be empty/falsy
     Failure/Error: it { is_expected.to validate_presence_of :user }
     
       Changeset did not properly validate that :user cannot be empty/falsy.
         After setting :user to ‹nil›, the matcher expected the Changeset to be
         invalid and to produce the validation error "can't be blank" on :user.
         The record was indeed invalid, but it produced these validation errors
         instead:
     
         * user: ["must exist"]
```

murbanski:
Just FYI, try `.with_message(:required)` to shorten it a little bit:)

robbl-as:
@mcmire @stephanngamedev Isn't that issue resolved by now?

mcmire:
[...]

PR #956, opened 2016-08-19
Support qualifier(optional, required) of 'belongs_to'

riseshia:
related to #870, #861 

This is to support qualifiers(optional, required) of `belongs_to`, Please review and let me know if it seems good. Then I will write documents for this :)

btw, there is one problem, test suites require `protected_attributes` gem so that we could not run test with rails 5 environment. ;(
Hence, I could grant `required` works well(with rails 4.2) with test, but not `optional` with rails 5.x.
Because of this, I had to confirm whether these work well or not in my personal rails 5 project.
anyway, we need to support test with rails 5 environment.

Thanks for reading this. 👍

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -255,6 +255,46 @@ module Shoulda
       #
       # @return [AssociationMatcher]
       #
+      # ##### required
+      #
+      # Use `required` to assert that the `:required` option was specified.
+      #
+      #     class Person < ActiveRecord::Base
+      #       belongs_to :organization, required: true
+      #     end
+      #
+      #     # RSpec
+      #     describe Person
+      #       it { should belong_to(:organization).required }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class PersonTest < ActiveSupport::TestCase
+      #       should belong_to(:organization).required
+      #     end
+      #
+      # @return [AssociationMatcher]
+      #
+      # ##### optional
+      #
+      # Use `optional` to assert that the `:optional` option was specified.
+      #
+      #     class Person < ActiveRecord::Base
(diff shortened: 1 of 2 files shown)


mcmire:
Cool, thanks for this. This looks good, although of course you're right, we will need to make sure that this works with Rails 5.

I'm going to start working on adding Rails 5 support soon, and I'll have to come back to this when that's done. But for now I've tagged this issue appropriately to remind myself.

riseshia:
Ok, then I will update documents soon :)

JoeWoodward:
What's the status on this? Anything I might be able to help with?

mcmire:
@JoeWoodward This PR is dependent on adding support for Rails 5 (#960). So it can't be finished and merged until that is complete.

I really haven't had time to work on this gem, by the way, so if you're looking to help out, that's the number item to finish right now.

edwardmp:
@mcmire I'll try to look into this. By the way, I use this gem with Rails 5 and besides this haven't experienced any issues.

By the way, would it make sense to support that `belongs_to :assocation_name, [required: true]` can also be tested by just putting `should validate_presence_of(:assocation_name)` explicitly in a test?

mcmire:
@edwardmp Yes, that would be the manual way to test it. So the `belongs_to` matcher somehow needs to automatically run that matcher as well. Perhaps this is the point at which composable matchers would be useful. If both `belongs_to` and `validate_presence_of` were composable, then the test could effectively say:

``` ruby
should belong_to(:association_name).and.validate_presence_of(:association_name)
```

mcmire:
@riseshia I realize it's been a long time, but thank you for this PR! I've taken your changes and put them into a new PR with additional tweaks: #1058. Instead of merely checking that the association has a `required` or `optional` option, we actually verify the given association is or is not allowed to be `nil`. I've given you credit in the commit message and of course the NEWS file will link back to this PR.

Issue #954, opened 2016-08-09
Support env named parameter to import SERVER_PORT header.


zw963:
Rails 5 could not get @request object in ActionDispatch::IntegrationTest,
But it support a env name args to support import SERVER_PORT header.

``` cl
get api_v1_test_hello_url,
      params: { name: 'new_name' },
      env: { 'SERVER_PORT' => '8629' }
```

How to do this in should-matcher ? 

Thanks.

zw963:
I have test like this, it told me no should method exist.

```rb
class RoutesTest < ActionController::TestCase
  include Shoulda::Matchers::ActionController
  setup { request.env['SERVER_PORT'] = '8629' }
  test 'test case' do
    should route(:get, 'v1/test/hello').to('api/v1/test#hello')
  end
end
```

If change to this:
```rb
class RoutesTest < ActionController::TestCase
  include Shoulda::Matchers::ActionController
  setup { request.env['SERVER_PORT'] = '8629' }
  should route(:get, 'v1/test/hello').to('api/v1/test#hello')
end
```

should exist, but `request.env` is not valid.

mcmire:
It looks like the `env` argument that `get`, `post`, etc. takes refers to the Rack environment hash. Setting `SERVER_PORT` doesn't actually change the Rack env, however -- [it gets used to build the URL](https://github.com/rails/rails/blob/3e45dd988cad539fe2cbc79b505fde99a44472e8/actionpack/lib/action_dispatch/testing/integration.rb#L396).

That said -- `route` doesn't actually hit a URL, it only checks to make sure that a combination of request method and corresponds to a controller/action pair. So I don't see why you need to set `SERVER_PORT` here.

zw963:
because i need test route which use constraint to limit only special port can be access。

mcmire:
Oh, interesting. I didn't know you could do that.

[Under the hood]](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/action_controller/route_matcher.rb#L131) the `route` matcher actually uses Rails' [`assert_routing`](http://api.rubyonrails.org/classes/ActionDispatch/Assertions/RoutingAssertions.html#method-i-assert_routing) test helper under the hood. There doesn't seem to be a way to specify a port with that, either. One would think it would be possible to specify a full URL (`assert_routing("http://foo.com:1234/my/path", ...)`), but I'm not sure if that works or not.

[...]

Issue #949, opened 2016-07-13
rails 5 depreciation warning validate_uniqueness_of for multiple scopes


elbartostrikesagain:
rails: 4.2.6
should-matchers: 3.1.1
(postgres if that matters)

I'm trying to use validate_uniqueness_of for something that is scoped to 2 things(note `default` is a boolean that defaults to false if that matters):

```
validates :default, uniqueness: { scope: [:object_type, :object_id], message: "an addresses object can only have one default" }
```

the test I'm using is:

```
it { should validate_uniqueness_of(:default).scoped_to(:object_type, :object_id).with_message('an addresses object can only have one default')}
```

The test passes but raises a depreciation warning for rails 5 😟 :

```
DEPRECATION WARNING: You attempted to assign a value which is not explicitly `true` or `false` ("an arbitrary value") to a boolean column. Currently this value casts to `false`. This will change to match Ruby's semantics, and will cast to `true` in Rails 5. If you would like to maintain the current behavior, you should explicitly handle the values you would like cast to `false`.)
```

s2t2:
I'm getting a slightly different error:

    it { should validate_uniqueness_of(:first_attribute).scoped_to([:poly_type, :poly_id]) }
     
     NoMethodError:
       undefined method `superclass' for nil:NilClass

Wondering if there is a temporary work-around?

mcmire:
@s2t2 That does seem like a separate issue. Care to post a backtrace?

s2t2:
@mcmire sure, see below:

```` rb
Failures:

  1) Notification validations uniqueness should validate that : first_attribute is case-sensitively unique within the scope of :poly_type and :poly_id
     Failure/Error: it { should validate_uniqueness_of(: first_attribute).scoped_to(:poly_type, :poly_id) }
     
     NoMethodError:
       undefined method `superclass' for nil:NilClass
     # /usr/local/var/rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activerecord-5.0.3/lib/active_record/validations/uniqueness.rb:47:in `find_finder_class_for'
     # /usr/local/var/rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activerecord-5.0.3/lib/active_record/validations/uniqueness.rb:14:in `validate_each'
[...]

Issue #945, opened 2016-07-05
have_db_index: normalize_columns_to_array is always an array


dniman:
In activerecord-5.0.0\lib\active_record\connection_adapter\postgresql\schema_statements.rb **columns** variable(in code below)  can be string or [].

```
def indexes
.....
    if indkey.include?(0) || opclass > 0
      columns = expression
    else
      columns = Hash[query(<<-SQL.strip_heredoc, "SCHEMA")].values_at(*indkey).compact
        SELECT a.attnum, a.attname
        FROM pg_attribute a
        WHERE a.attrelid = #{oid}
          AND a.attnum IN (#{indkey.join(",")})
      SQL

      # add info on sort order for columns (only desc order is explicitly specified, asc is the default)
      orders = Hash[
        expressions.scan(/(\w+) DESC/).flatten.map { |order_column| [order_column, :desc] }
       ]
     end 
 ..........
 end
```

In shoulda-matchers-3.1.1\lib\shoulda\matchers\active_record\have_db_index_matchar.rb 
   @columns is always an [] after normalize_columns_to_array method. 

```
.....
def initialize(columns)
  @columns = normalize_columns_to_array(columns)
  @options = {}
end
.....
def matched_index
  indexes.detect {|each| each.columns == @columns}
end
.....

def indexes
  ::ActiveRecord::Base.connection.indexes(table_name)
end
.....
def normalize_columns_to_array(columns)
  Array.wrap(columns).map(&:to_s)
end
```

Therefore we cannot use functional indexes like lower((code)::text) with have_db_index.
In my_model_spec.rb:

```
it { should have_db_index("lower((code)::text)").unique(:true)}
```

and after running spec:

```
Failure/Error: it { should have_db_index("lower((code)::text)").unique(:true)}
  Expected MyModel to have a unique index on columns lower((code)::text) ()
```

I changed the matched_index method locally in shoulda-matchers-3.1.1\lib\shoulda\matchers\active_record\have_db_index_matcher.rb
to fix this from

```
 def matched_index
   indexes.detect { |each| each.columns == @columns }      
 end
```

to

```
def matched_index
  indexes.detect { |each| each.columns == (each.columns.is_a?(String)? @columns.join('') : @columns) }
end
```

tejasbubane:
Could you use markdown code highlighting with proper indentation? 
Refer: https://guides.github.com/features/mastering-markdown/ At the moment, it is difficult to make out what your code is doing.

[...]

Issue #944, opened 2016-07-02
have_db_column of_type can't compare types with activerecord-jdbc-adapter/rails 5


wintersolutions:
When I run the following spec in MRI everything works:

`it { is_expected.to have_db_column(:example).of_type(:string) }`

When I run it on JRuby 9.1.2 and with activerecord-jdbc-adapter (github: 'jruby/activerecord-jdbc-adapter', branch: 'rails-5') I get:

> Expected Keyword to have db column named language_code of type string of limit 2 (FooBar has a db column named example of type #ActiveModel::Type::String:0x4c082a7d, not string.)

This seems to be the case for all types of attributes, not only strings. A possible workaround, i.e. comparing to `ActiveModel::Type::String` on JRuby, is not possible since were getting a instance instead of class when we call `to_s` i.e. ActiveModel::Type::String:**0x4c082a7d**. 

That means `shoulda-matchers` will not be compatible with rails 5 on JRuby.

Relevant line:
https://github.com/thoughtbot/shoulda-matchers/blob/v3.1.1/lib/shoulda/matchers/active_record/have_db_column_matcher.rb#L152

wintersolutions:
Found similar issues with `have_db_column(:example).with_options(limit: 42)` and `enum_for` matchers. I wrote the following monkeypatch to get my specs running again:

```
if RUBY_PLATFORM == 'java'
  class Shoulda::Matchers::ActiveRecord::HaveDbColumnMatcher

    protected

    def correct_column_type?
      return true unless @options.key?(:column_type)

      if matched_column.type.class.to_s =~ /#{@options[:column_type].to_s}/i
        true
      else
        @missing = "#{model_class} has a db column named #{@column} " <<
            "of type #{matched_column.type.class.to_s} which did not match #{@options[:column_type]}."
        false
      end
    end

    def correct_limit?
      return true unless @options.key?(:limit)

      if matched_column.type.limit.to_s == @options[:limit].to_s
        true
      else
        @missing = "#{model_class} has a db column named #{@column} " <<
            "of limit #{matched_column.type.limit}, " <<
            "not #{@options[:limit]}."
        false
      end
    end
  end

  class Shoulda::Matchers::ActiveRecord::DefineEnumForMatcher

    def column_type_is_integer?
      column.type.class.to_s =~ /integer/i
    end
  end
end
```

mcmire:
[...]

PR #960, opened 2016-09-04
WIP: Add support for Rails 5

mcmire:
- [ ] Fix `validate_inclusion_of` so that it only checks if `minimum` is 0 if the column is a numeric column
- [ ] Fix places where we are using controller test helpers (`get`, `post`, etc.) so that they use keyword arguments (#867, fixed by #964)
- [ ] Fix `serialize` so that it does not refer to `cast_type` (#913)
- [x] Instead of referring to just `connection.tables`, use `connection.data_sources` (#933, fixed by #943)
- [ ] Fix compatibility with JRuby / JDBC (#944)
- [ ] Fix `have_db_index` so it works with expressions as well as columns (#945)
- [ ] Fix `validate_uniqueness_of` when used with a boolean so it doesn't produce a warning (#949)
- [ ] Modify `route` to support a `port` option (#954)
- [ ] Fix `validate_presence_of` as the validation message changed ([#861](https://github.com/thoughtbot/shoulda-matchers/issues/861#issuecomment-189563425))


stevebissett:
I came across this from:
http://stackoverflow.com/questions/39611716/how-do-i-test-associations-in-models-with-rspec-in-rails-5?noredirect=1#comment66531045_39611716

@mcmire - do you have any idea on when you might be done with this? Appreciate the work BTW.

mcmire:
Hey @stevebissett,

Sorry for the wait on this. Unfortunately I don't really have a timeline for when this will be complete. I tend to work on it whenever I have time and I haven't had a whole lot of that lately.

aried3r:
Hey @mcmire, I saw you sadly no longer have the time to maintain this project. I wanna start off by saying Thanks! for all your efforts so far! 👍 

But wrt to this PR, would you be open to drop the conditionals for ActiveModel 3 and maybe some older Rails 4.x versions in favor of supporting Rails 5.0 and the upcoming 5.1 which will remove a lot of deprecated functions. It would require a new major release, but also means that there would be a working version of shoulda-matchers (3.x) for older Rails versions and a new, lighter version (because of dropping support) of it?

I even see `unless active_model_3_2? || active_model_4_0?` in this PR (although just moved).

mcmire:
@aried3r Thanks for the support, it means a lot. :) Yeah, I am totally open to that. In fact for shoulda I have already removed support for Rails < 4.2: https://github.com/thoughtbot/shoulda/commit/b9353e29cdfbc484bfd808df55226bb26d0fc4a5, so I think we could definitely do the same thing here and release a new major version.

em77:
Hey @mcmire I just wanted to check in and see if there's any update on the timeline for this. I appreciate all the work you've put into this project and would like to see it continue to be compatible with Rails versions going into the future!

guialbuk:
Hi, @em77 .
The Rails 5 support is in progress but it can be used in production (I'm using in two Rails 5.1 projects)
Point you Gemfile to ```rails-5``` branch and you are good to go :)
https://github.com/thoughtbot/shoulda-matchers#getting-started

em77:
@guialbuk Hey thanks for the encouraging reply. So that branch is considered safe for rails 5.1? Is there any documentation of what methods or functionality are _not_ working properly on rails 5.0 and 5.1?

guialbuk:
@em77 It is safe to use on Rails 5.0 and 5.1 The list on the top of this PR (```rails5``` branch, not ```rails-5```) is not up to date and many issues have been solved. Even Ruby 2.4 is supported in the ```rails-5```branch.
The ```rails-5```branch will soon be merged into ```master``` and become ```4.0.beta1``` version.

mcmire:
@guialbuk I wonder if it would be helpful to keep this list up to date while the `rails-5` branch is WIP (this way everyone knows, and it helps me triage things). I know you've merged the `tables` PR (#943) -- is there anything else that your branch has fixed, or is that it for now?

mcmire:
Actually I think what I'll do is close this and move this to an issue so it's not confusing. But I'd still like to keep a list :)

PR #971, opened 2016-10-27
Add MRI Ruby 2.3 & 2.4 to the build process

thedrow:
Supersedes #921


lucianosousa:
it will not work. you should put explicitly `2.3.1`

oliverklee:
In the meantime, we should use 2.3.3. And 2.4.0 has also been released.

guialbuk:
Fixed by #1027

PR #992, opened 2017-01-11
pass correct object type to association reflection

### LINKED ISSUE #952: Conditions qualifier, when association owner is part of the condition
How do you validate the conditions for a `has_one`/`has_many` association, if the association owner is part of the condition? Example:

``` ruby
class Foo
  has_many :bars, -> (foo) { where(baz_id: foo.default_baz_id }
  belongs_to :default_baz, class_name: 'Baz'
end

class Bar
  belongs_to :foo
  belongs_to :baz
end
```

I tried both `conditions(where(baz_id: foo.default_baz_id)` and `conditions(-> (foo) { where(baz_id: foo.default_baz_id })`, but I get:

``` ruby
NoMethodError:
  undefined method `default_baz' for #<Class:0x007fc2d6ca7dd0>

```

formigarafa:
This is to make the association relation receive the expected parameter
(a AR instance, not its class). fix #952

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/association_matchers/model_reflection_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/association_matchers/model_reflection_spec.rb
@@ -95,11 +95,27 @@ describe Shoulda::Matchers::ActiveRecord::AssociationMatchers::ModelReflection d
             belongs_to :country, -> { where(mood: 'nice') }
           end
           delegate_reflection = person_model.reflect_on_association(:country)
+          person = person_model.new
           reflection = described_class.new(delegate_reflection)
 
-          actual_sql = reflection.association_relation.to_sql
+          actual_sql = reflection.association_relation(person).to_sql
           expected_sql = Country.where(mood: 'nice').to_sql
           expect(actual_sql).to eq expected_sql
+        end
+
+        it 'passes the object that has the association to the block' do
+          spy = spy('spy')
+          define_model(:country)
+          person_model = define_model(:person, country_id: :integer) do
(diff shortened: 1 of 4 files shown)


spec/unit/shoulda/matchers/active_record/association_matchers/model_reflection_spec.rb
```diff
@@ -92,13 +92,15 @@
         context 'when the scope is a block' do
           it 'executes the block in the context of an empty scope' do
             define_model(:country, mood: :string)
-            person_model = define_model(:person, country_id: :integer) do
-              belongs_to :country, -> { where(mood: 'nice') }
+            columns = { country_id: :integer, spirit: :string }
+            person_model = define_model(:person, columns) do
+              belongs_to :country, ->(person) { where(mood: person.spirit) }
```
mcmire: Thanks for this PR. Instead of modifying this test (which is a known use case) what do you think about making a new test?

formigarafa: I also thought the same, previously I committed as such, but it looked so much as a duplication that I removed. I will put it back.

formigarafa:
I wrote another test before but it was duplicating most of the previous code, so I removed it.
Now I have fixed it to only test the change. It should be better, now.

mcmire:
Sorry for the delay in getting to this. I took the liberty of adding a missing test to this and cleaning up a couple things around the existing tests. This is merged now!

formigarafa:
@mcmire , I don't mind at all and I really appreciate that. Thanks.

PR #995, opened 2017-01-20
uniqueness: don't use a private Rails API

alyssais:
`ActiveRecord::Validations#_validations` is a private API.

Fortunately, Rails provides an API that does exactly what's needed here, and it's been in Rails since at least 3.0.0, so let's use that instead.

***

I ran into an issue using shoulda-matchers with [schema_validations](https://github.com/SchemaPlus/schema_validations). It overrides the public `ActiveModel::Validations` methods to dynamically add validations based on the database schema. Since shoulda-matchers accesses validations using a private API, it incorrectly causes test failures.

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -396,7 +396,7 @@ module Shoulda
         end
 
         def validations
-          model._validators[@attribute].select do |validator|
+          model.validators_on(@attribute).select do |validator|
             validator.is_a?(::ActiveRecord::Validations::UniquenessValidator)
           end
         end


lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -381,7 +381,7 @@ def description_for_case_sensitive_qualifier
         end
 
         def validations
-          model._validators[@attribute].select do |validator|
+          model.validators_on(@attribute).select do |validator|
```
TheTeaNerd: :+1:, https://apidock.com/rails/ActiveModel/Validations/ClassMethods/validators_on does seem more appropriate.

mcmire:

alyssais:
> Don't take this the wrong way: our aim is not to diminish the effort
> people have made or dismiss problems that have been raised. If you
> feel that we should reopen this PR, then please let us know so that we
> can reprioritize it. Thanks!

This appears to still be relevant, and it's a trivial change, so I think
this PR should be reopened.

mcmire:
Ok!

mcmire:
I apologize it took me so long to merge this — this is in master now.

PR #1008, opened 2017-04-09
Fix validate number comparison with strict error

sunkai612:
Thanks first for this project, it saves so much time 😄 

I've encountered an `ArgumentError` with below context:
```Ruby
# app/models/foo.rb
validates :positive_integer_column,
  numericality: {
    only_integer: true,
    greater_than: 0
  },
  strict: true

# spec/models/foo.rb
it do
  should validate_numericality_of(:positive_integer_column)
           .only_integer
           .is_greater_than(0)
           .strict
end
```

This PR contains the fix for the combined usage of comparison and strict and add tests for it.

Thanks for the review in advance 😃 👍

### DIFF
--- a/lib/shoulda/matchers/active_model/validation_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validation_matcher.rb
@@ -24,8 +24,8 @@ module Shoulda
           self
         end
 
-        def strict
-          @expects_strict = true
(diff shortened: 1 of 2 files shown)


yeuem1vannam:
Any plan to merge this PR?

mcmire:

PR #1012, opened 2017-04-19
Validate confirmation of with integers

kheppenstall:
In response to issue #865 to allow `validate_confirmation_of` with integers, instead of just strings.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
@@ -109,8 +109,18 @@ module Shoulda
         end
 
         def allows_same_value
+          allows_same_value_as_int || allows_same_value_as_string
+        end
+
+        def allows_same_value_as_string
           allows_value_of('same value') do |matcher|
             qualify_matcher(matcher, 'same value')
+          end
+        end
+
+        def allows_same_value_as_int
+          allows_value_of(2) do |matcher|
+            qualify_matcher(matcher, 2)
           end
         end
 
--- a/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
@@ -94,38 +94,48 @@ module Shoulda
 
         def matches?(subject)
           super(subject)
-
-          disallows_different_value &&
-            allows_same_value &&
-            allows_missing_confirmation
+          type = find_type(subject)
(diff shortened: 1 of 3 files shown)


kheppenstall:
I'm new to contributing so any feedback is appreciated and I'm of course happy to rewrite however!

lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
```diff
@@ -94,28 +94,48 @@ def simple_description
...
         private
 
-        def disallows_different_value
-          disallows_value_of('different value') do |matcher|
-            qualify_matcher(matcher, 'some value')
+        def find_type(subject)
+          subject.class.columns_hash['attribute_to_confirm'].type rescue :string
+        end
+
+        def different_value
```
mcmire: What do you think about defining these as constants rather than methods? Also, what do you think about using the plural (`DIFFERENT_VALUES`) rather than the singular?

spec/unit/shoulda/matchers/active_model/validate_confirmation_of_matcher_spec.rb
```diff
@@ -146,4 +146,36 @@
...
+  context 'when attribute to confirm is type integer' do
+    context 'when the model has a confirmation validation' do
+      it 'passes' do
+         builder = builder_for_record_validating_confirmation(type: :integer)
+      expect(builder.record).
+        to validate_confirmation_of(builder.attribute_to_confirm)
+      end
+    end
+
+    context 'when the model has all attributes, but does not have the confirmation'do
```
mcmire: Let's tweak this wording a bit. What do you think about just using:

* "when the model has a confirmation validation on the given attribute"
* "when the model does not have a confirmation validation on the given attribute"

lib/shoulda/matchers/active_model/validate_confirmation_of_matcher.rb
```diff
@@ -94,28 +94,48 @@ def simple_description
 
         def matches?(subject)
           super(subject)
-
-          disallows_different_value &&
-            allows_same_value &&
-            allows_missing_confirmation
+          type = find_type(subject)
```
mcmire: If `find_type` no longer takes a `subject`, then you don't really need to assign a variable to the method -- you can use it as though it were a variable, and that means you don't need to pass it to every place that needs it. What do you think about renaming `find_type` to simply `attribute_type`?

mcmire:

Issue #1007, opened 2017-04-07
validate_length_of does not work when used with associations


LukasBarry:
I have been attempting to write a test on a validation for length, and I keep getting the same error. I created a small sample app, as simple as possible, to show the error: [Sample Project](https://github.com/LukasBarry/shoulda_failure/tree/master/ShouldaFailure)

Basically, I have a has_many association that must have at least 1, but when I run the test
`should validate_length_of(:books).is_at_least(1)` I get the following error:

```
NoMethodError:
       undefined method `each' for "":String
```

My sample project current models:
```
class Author < ApplicationRecord
  has_many :books

  validates :books, length: { minimum: 1 }
end
```
```
class Book < ApplicationRecord
  belongs_to :author, dependent: :destroy
end
```
And my current test file:
```
require 'rails_helper'

describe Author do
  it 'validates length of books' do
    should validate_length_of(:books).is_at_least(1)
  end
end
```

I believe this is because Shoulda is using a string to test length, but I am attempting to test the length of a collection and not a string. However, while attempting to trace it back through the gem, I found nothing that I could use to fix my issue. 

Any help would be greatly appreciated. Thanks!

mcmire:
Hey @LukasBarry. When using `is_at_least`, the matcher will build a string with length of `<length> - 1` (which in this case is 0), set the attribute (`books`) to that string, and assert that this causes a failure. You can see that happening here: https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb#L340

Regardless of how you are using `validate_length_of`, however, the matcher does make the assumption that the column you're testing is a string column; it's not designed to work with anything else. I've never used the validation with associations before, but after reading the Rails source code, it does seem like it's a valid use case, so this is new to me. We'll need a PR to fix this. Can you submit one?

LukasBarry:
Absolutely, I will get one together and submit it. Thanks!

ysyyork:
Hi, I just came across the same issue and was planning to make changes. But have you already started the change? And is there any update?

mcmire:
Hey @ysyyork! As you can see here there is a PR here that was submitted. While it's a good start it does need some tests. Unfortunately this issue is low on our priority list at the moment but if you would like to pick it up and supply the tests (submitting a new PR of course) then that would be much appreciated!

ysyyork:
@mcmire Sure, I will give it a look if I have the time.

[...]

PR #1016, opened 2017-05-01
Changes string_of_length to handle arrays

LukasBarry:
# Changes `string_of_length` to handle arrays

I recently discovered an error on a valid use case of the `validates_length_of` in the `shoulda-matchers` gem. I was attempting to test the length of an associated array, but kept receiving the same error:
```
NoMethodError:
```
[Here is the issue I submitted to Thoughtbot](https://github.com/thoughtbot/shoulda-matchers/issues/1007)

Here is my suggested fix. I am new to contributing, so any feedback would be greatly appreciated!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,20 +9,20 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    appraisal (2.1.0)
+    appraisal (2.2.0)
       bundler
       rake
       thor (>= 0.14.0)
     byebug (9.0.6)
     coderay (1.1.1)
-    diff-lcs (1.2.5)
+    diff-lcs (1.3)
     fssm (0.2.10)
     method_source (0.8.2)
     multi_json (1.12.1)
     pry-byebug (3.4.2)
       byebug (~> 9.0)
       pry (~> 0.10)
-    pygments.rb (1.1.1)
+    pygments.rb (1.1.2)
       multi_json (>= 1.0.0)
     rake (10.5.0)
     redcarpet (3.4.0)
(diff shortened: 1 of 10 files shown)


lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
```diff
@@ -368,15 +368,19 @@ def allows_maximum_length?
...
 
         def disallows_length_of?(length, message)
-          disallows_value_of(string_of_length(length), message)
+          disallows_value_of(string_or_array_of_length(length), message)
         end
 
-        def string_of_length(length)
-          'x' * length
+        def string_or_array_of_length(length)
+          if @attribute.is_a? String
```
mcmire: Hmm... I'm not sure that this actually works like you think. I believe that `@attribute` here refers to the _name_ of the attribute, not an attribute as a whole.

In order to properly check this, you have to look at the column type. Unfortunately, you can't make the assumption that the model being tested is an ActiveRecord model -- it could be an ActiveModel model, in which case there is no column and hence no type.

There are other matchers in which we do this properly; take a look at https://github.com/thoughtbot/shoulda-matchers/blob/ac9f4822ca25732ccf4d0081d28cac7f19f3b9aa/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L517 for inspiration. (Although it shouldn't be that complicated in this case -- you don't need to map `:integer`, for instance, to `:fixnum` or anything like that.)

Gemfile.lock
```diff
@@ -9,20 +9,20 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    appraisal (2.1.0)
+    appraisal (2.2.0)
```
mcmire: Do you mind breaking these changes to the gemfiles off into a separate commit? It looks like the crux of these changes is upgrading Appraisal from 2.1.0 to 2.2.0.

matsales28:
The idea behind this PR was resolved on https://github.com/thoughtbot/shoulda-matchers/pull/1560. I appreciate your contribution and time @LukasBarry. I'll be closing this PR.

PR #1025, opened 2017-06-29
Add matcher for has_secure_token

fny:
This pull request adds a matcher for Rails 5's has_secure_token.

This matcher will check for the following:

 - has_secure_token has been called on the ActiveRecord model
 - the appropriate columns and indices exist in the databse

The supports the default :token and custom tokens (e.g. `has_secure :auth_token`)


lib/shoulda/matchers/active_record.rb
```diff
@@ -13,6 +13,7 @@
 require "shoulda/matchers/active_record/have_db_column_matcher"
 require "shoulda/matchers/active_record/have_db_index_matcher"
 require "shoulda/matchers/active_record/have_readonly_attribute_matcher"
+require "shoulda/matchers/active_record/have_secure_token_matcher"
```
fny: I'm just following the existing styles in this file! 😄

lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        def failure_message
+          "Expected #{@subject} to #{description} but the following issues were found: #{@issues.join(', ')}"
+        end
+
+        def failure_message_when_negated
+          "Did not expect #{@issues.join(', ')}"
+        end
+
+        def matches?(subject)
```
fny: This method only has 9 lines?

lib/shoulda/matchers/active_record.rb
mcmire: Yeah, don't worry about this :)

lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
```diff
@@ -0,0 +1,105 @@
...
+          @subject = subject
+          @issues = []
+          add_issue_unless has_expected_instance_methods?(subject),
```
fny: This would drive up the method length tremendously.  Any ideas?

mcmire: I'm not worried necessarily about the method length. Right now the indentation makes it a little readable (at least compared to the style that's being used) and it's hard to make out what the arguments are.

I suppose, though, that it does feel a little procedural in this method. I also don't see much of a need for `add_issue_unless`, to be honest. But I like the validation pattern you've established here. So what do you think about this (suggestions for a method name are welcome)?

``` ruby
def matches?(subject)
  @subject = subject
  @errors = run_checks
  @errors.empty?
end

private

def run_checks
  errors = []

  if !has_expected_instance_methods?
    errors << "..."
  end

  if !has_expected_db_column?
    errors << "..."
  end
  
  if !has_expected_db_index?
    errors << "..."
  end

  errors
end
```

fny:
Hah! This is nearly identical to what I had in my first iteration. I only
extracted everything into `add_issue_unless` to make Rubocop shut up about
being one line over the limit... Ah, the dark side of linting...

I think you'll be pleased with the next push!

2017-06-29 19:51 GMT-04:00 Elliot Winkler <notifications@github.com>:

fny:
We should be good to go here.

carlosramireziii:
Anything I can do to help this PR get merged?

erlandsona:
Would love to see this merged :) 👍

lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
```diff
@@ -0,0 +1,111 @@
...
+            @subject.respond_to?("regenerate_#{token_attribute}") &&
+            @subject.class.respond_to?(:generate_unique_secure_token)
+        end
+
+        def has_expected_db_column?
+          matcher = HaveDbColumnMatcher.new(token_attribute).of_type(:string)
+          matcher.matches?(@subject)
+        end
+
+        def has_expected_db_index?
```
kuldeepaggarwal: AFAIK, This is not the condition of `has_secure_token :column_name` method call

fny: Nor is it a condition that column of `has_secure_token` that the column exists, but in my opinion, it's more superficial test without those checks. In what case is it appropriate to have a unique token without a corresponding database column? In what case is it appropriate to have a unique token without a unique index in the db?

mockdeep:
Looks like your PR needs a rebase @fny. Thanks for doing this!

mcmire:
Sorry, yeah, master has gone through several changes since this PR was released. A rebase would be lovely although we will certainly get to this when we get some time, just a little busy at the moment. Sorry for the patience everyone. :)

guialbuk:
Thanks, @mcmire . I've closed the PR by accident.
Now merging and moving towards 4.0 :)

PR #1032, opened 2017-07-11
Drop support for deprecated Ruby (2.0 and 2.1) and Rails (4.0 and 4.1) versions

guialbuk:
Support of Ruby 2.1 has ended on April, 1st 2017
https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

Rails 4.1 has reached end of life in 2016

This is an important step to keep shoulda-matchers up to date with current ruby and rails versions and to keep the codebase clean and easier to contribute.


mcmire:
👏 👏 👏 

Great! Keep in mind that this means we'll need to bump to 4.0. I would be fine with that, but it's something to keep in mind.

By the way -- I've been keeping a log of changes in NEWS.md. You may want to add this there (it's easier to do it as you go along rather than before a release).

guialbuk:
Thanks, @mcmire . I will keep NEWS.md updated.

PR #1037, opened 2017-07-25
Rails 5: Skip array-column-related tests for uniqueness matcher

mcmire:
For some reason, in Rails 5, the uniqueness validation has stopped
working for array columns. What this means is that if you try to use an
array column for a scope, then the query that the validation performs
blows up because the value in the query for the scope is not an array
but a regular scalar value. This does not seem to be documented
anywhere, so perhaps this never worked to begin with, or it accidentally
worked in Rails 4. Until we solve this mystery, skip the tests that have
to do with array columns.

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -808,7 +808,11 @@ within the scope of :scope1.
       end
     end
 
-    if database_supports_array_columns? && active_record_supports_array_columns?
+    if (
+      database_supports_array_columns? &&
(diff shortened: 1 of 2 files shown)


mcmire:
@guialbuk Pinging you on this. Wanted to make sure you saw this and had a chance to review it. The build fails right now because this is dependent on #1038.

spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
```diff
@@ -808,7 +808,11 @@
       end
     end
 
-    if database_supports_array_columns? && active_record_supports_array_columns?
+    if (
```
mcmire: This is unfortunate but there's not a better way to do this.

PR #1038, opened 2017-07-25
Doublespeak: Define #respond_to_missing? on ObjectDouble

### LINKED ISSUE #1006: delegate_method produces `forwarding to private method` warning on MRI-2.4
Forwarding methods to private methods is deprecated and produces a warning in MRI-2.4 (see [MRI bug #12782](https://bugs.ruby-lang.org/issues/12782)
```ruby
require 'forwardable'

class PostOffice
  def deliver
  end
end

class Courier
  extend Forwardable

  attr_reader :post_office

  def_delegators :post_office, :deliver

  def initialize
    @post_office = PostOffice.new
  end
end


describe Courier do
  it { should delegate_method(:deliver).to(:post_office) }
end
```
produces
```
Courier
/Users/ojab/.rvm/gems/ruby-2.4.0/bundler/gems/shoulda-matchers-d3145e911902/lib/shoulda/matchers/independent/delegate_method_matcher.rb:335:in `public_send': Courier#deliver at /Users/ojab/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/forwardable.rb:156 forwarding to private method #deliver
  should delegate #deliver to #post_office object

Finished in 0.00153 seconds (files took 0.90469 seconds to load)
1 example, 0 failures
```

Doublespeak uses private `method_missing` for catching calls, so we have an issue.
I've filled [MRI bug 13325](https://bugs.ruby-lang.org/issues/13326) for clarification about delegate+method_missing.

mcmire:
The `delegate_method` matcher uses Doublespeak internally to do its job.
The delegate object is completely stubbed using an ObjectDouble,
available from Doublespeak, which is shoulda-matchers's own test double
library. ObjectDouble is a class that responds to every
method by implementing `method_missing`.

Say you are using Ruby 2.4 and you are testing a class that uses the
Forwardable module to do some delegation, and you are using
`delegate_method` in your test. When you run your test you will a
warning that looks something like:

    Courier#deliver at ~/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/forwardable.rb:156 forwarding to private method #deliver

Why is this happening? When the code in your class gets exercised,
Forwardable will delegate the delegate method in question to
ObjectDouble, and the method will get intercepted by its
`method_missing` method. The fact that this is actually a private method
is what Forwardable is complaining about.

To fix this, all we need to do is add `respond_to_missing?` in addition
to `method_missing?`. This is explained here:

https://bugs.ruby-lang.org/issues/13326

---

This fixes #1006.

### DIFF
--- a/lib/shoulda/matchers/doublespeak/object_double.rb
+++ b/lib/shoulda/matchers/doublespeak/object_double.rb
@@ -14,11 +14,11 @@ module Shoulda
           @calls_by_method_name[method_name] || []
         end
 
-        def respond_to?(name, include_private = nil)
+        def respond_to?(_name, _include_private = nil)
           true
         end
 
-        def respond_to_missing?(name, include_all)
+        def respond_to_missing?(_name, _include_all)
           true
         end
 
--- a/lib/shoulda/matchers/doublespeak/object_double.rb
+++ b/lib/shoulda/matchers/doublespeak/object_double.rb
@@ -18,6 +18,10 @@ module Shoulda
           true
         end
 
+        def respond_to_missing?(name, include_all)
+          true
+        end
+
         def method_missing(method_name, *args, &block)
           call = MethodCall.new(
             method_name: method_name,


mcmire:
@guialbuk Pinging you on this. Wanted to make sure you saw this and had a chance to review it. The build fails right now because this is dependent on #1037.

mcmire:
Oh whoops! Looks like you did, never mind :)

PR #1051, opened 2017-08-03
Use model class' DB connection in have_db_index matcher

jeffwatts:
**Why**
If a rails project connects to mutliple distinct databases, then it will
only be able to connect to the primary application DB if
ActiveRecord::Base.connection is used to find indexes. By using the
model class under test's connection, we ensure that we get the
appropriate set of indexes for our test.

**What**
Uses `model_class.connection.indexes` to get indexes for a table instead
of `ActiveRecord::Base.connection.indexes`.

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
@@ -136,7 +136,7 @@ module Shoulda
         end
 
         def indexes
-          ::ActiveRecord::Base.connection.indexes(table_name)
+          model_class.connection.indexes(table_name)
         end
 
         def expectation


mcmire:
@jeffwatts Thanks for the PR! I'm wondering if we can add a test for this. How are you using multiple databases in production?

jeffwatts:
Yeah, I briefly looked into how I might add a test for this, but it didn't look super easy in your existing framework. In our production code, we connect to DB's that are not our primary app DB by essentially calling [`establish_connection`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-establish_connection) with the appropriate connection params when we load models that do not connect to the primary app DB. Does that make sense? It looks something like this:

```ruby
class NonAppDbModel < ActiveRecord::Base
  self.abstract_class = true

  def self.init_connection
    config = get_non_app_db_connection_info
    establish_connection(config)
  end

  init_connection
end

class ConcreteNonAppDbModel < NonAppDbModel
  # use like a typical AR model
end
```

jeffwatts:
Do you have any suggestions for how I can add tests within your codebase for this? I looked into it and it seemed quite difficult given the way things are set up now, especially relative to the nature of the change here.

jooeycheng:
Hi @mcmire, I facing similar issues with @jeffwatts, and would like for this fix be merged.

Could you advise how to test this, and what steps are required to "revive" this PR.
I think the fix is pretty straightforward, and existing tests should cover.

jooeycheng:
Related: https://github.com/thoughtbot/shoulda-matchers/commit/41c763d27656fd76ed6be567ce9cb063f6ab6696

jooeycheng:
Hey @mcmire, just checking in on the status of this. Is there anything I can do to help?

mcmire:
@jooeycheng If the original use case still applies here, then we need tests that reflect this use case. The problem is that our test suite is not set up to connect to multiple databases. We would have to investigate how to do that in order to properly merge this in.

When running the unit tests, a Rails application is generated on the fly at the beginning and it is booted so that ActiveRecord is initialized accordingly based on the settings inside of that Rails app. [This](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/unit/rails_application.rb) is the file that sets up the Rails app and [here](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/unit/rails_application.rb#L112-L114) is where we write the `database.yml` file, and the settings for that file are determined [here](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/tests/database_configuration.rb) and [here](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/tests/database.rb). Effectively there are three environments set up in `database.yml` but they all use the same database name. What we could do instead is set up two environments, "development" and "production", and we could add a suffix to the database name accordingly so that we end up with `shoulda-matchers-test_development` and `shoulda-matchers-test_production`.

Once we have that, then we can write a new test. If you look at the bottom of the [tests for `have_db_index`](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/unit/shoulda/matchers/active_record/have_db_index_matcher_spec.rb) you will see that we have a helper method that creates a table and an associated model. We can write a new failing test that uses the same `create_table` and `define_model_class` methods to replicate the scenario that causes this bug. (You can look at [ModelBuilder](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/unit/helpers/model_builder.rb#L37-L53) and [ClassBuilder](https://github.com/thoughtbot/shoulda-matchers/blob/e5948d586f08b0ce6846b334d114fb818a231467/spec/support/unit/helpers/class_builder.rb#L48-L74) for more on how these work.)

jooeycheng:
@mcmire 

I have opened #1200. Let me know if it makes sense.

> I think there are three just to replicate how a typical Rails app works, but that isn't necessary.) What we could do instead is set up two environments, "development" and "production" 

I'll retain all 3 environments bcos I received an error after removing `test`:

```
RuntimeError:
  Command "bundle exec rake db:drop:all db:create:all db:migrate" exited with status 1.
  Output:
  ---START----------------------------------------------------------------
  rake aborted!
  ActiveRecord::AdapterNotSpecified: 'test' database is not configured. Available: ["development", "production"]
  /home/travis/.rvm/gems/ruby-2.4.4/gems/activerecord-4.2.11.1/lib/active_record/connection_adapters/connection_specification.rb:248:in `resolve_symbol_connection'
  /home/travis/.rvm/gems/ruby-2.4.4/gems/activerecord-4.2.11.1/lib/active_record/connection_adapters/connection_specification.rb:211:in `resolve_connection'
  /home/travis/.rvm/gems/ruby-2.4.4/gems/activerecord-4.2.11.1/lib/active_record/connection_adapters/connection_specification.rb:139:in `resolve'
```

However, we'll only be using the `development` and `production` databases.

PR #1058, opened 2017-10-03
Add required/optional qual's to belongs_to/has_one

mcmire:
Rails 5 made two changes to `belongs_to` associations:

* `required` and `optional` were added as options (which add and remove
  a presence validation on the association, respectively)
* `required` was made the default

In addition, a `required` option was also added to `has_one`.

These new qualifiers allow us to test these options appropriately.

Credit: Shia <rise.shia@gmail.com>

---

Fixes #956, #870, #861.


mcmire:
@guialbuk Would you mind reviewing this? :)

PR #1082, opened 2018-02-25
Fix validates_uniqueness_matcher for integer attributes

### LINKED ISSUE #1081: validates_uniqueness_of doesn't work correctly on integer columns
Hi,

I love this gem and recently needed something that isn't released yet, so I felt a bit adventurous and tried out what's in the master branch.

I encounter one specific issue. If you have a `validates_uniqueness_of` matcher on an Integer-backed column I receive an `undefined method length for 0:Integer` error. 

Diving into the code a bit, I notice that in the [`arbitrary_non_blank_value`](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L520) method the value of `non_blank value` would equal `0` for a Integer column. An integer column typically has a limit of 4 (bytes), but this code somehow assumes that if there is a limit, `non_blank_value` responds to the `length` method as well, which of course is not the case for an Integer value like `0`.

I believe this issue may have been introduced by this commit: https://github.com/thoughtbot/shoulda-matchers/commit/9103ead2147b012dabf7fdf9c68b9995d24697ae. Before this change `non_blank value` always defaulted to a hardcoded string, but after the change it can be other types than a string.

A fix could check if `non_blank` responds to `length`, but that probably isn't an proper fix.
I guess that basically the code in the if statement is only relevant for `String` type columns, thus we should probably check if `non_blank_value` is a string or not. If someone can confirm if that is what's intended then I can submit a PR fixing this.

Example fix:
```ruby
if non_blank_value.is_a?(String) && limit && limit < non_blank_value.length
    'x' * limit
else
    non_blank_value
end
```

edwardmp:
This fixes #1081. As discussed, I first added a failing test that confirms the issue is present, then added the fix and checked that the test passed.

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -521,7 +521,8 @@ module Shoulda
           non_blank_value = dummy_value_for(@attribute)
           limit = column_limit_for(@attribute)
 
-          if limit && limit < non_blank_value.length
+          is_string_value = non_blank_value.is_a?(String)
+          if is_string_value && limit && limit < non_blank_value.length
             'x' * limit
           else
             non_blank_value
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -356,6 +356,17 @@ Example did not properly validate that :attr is case-sensitively unique.
       end
     end
 
+    context 'when the attribute is of integer type' do
+      it 'accepts' do
+        record = build_record_validating_uniqueness(
+          attribute_type: :integer,
+          attribute_options: { limit: 4 },
+        )
+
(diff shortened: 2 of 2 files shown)


lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
```diff
@@ -520,8 +520,8 @@ def update_existing_record!(value)
         def arbitrary_non_blank_value
           non_blank_value = dummy_value_for(@attribute)
           limit = column_limit_for(@attribute)
-
-          if limit && limit < non_blank_value.length
+          if non_blank_value.is_a?(String) && limit &&
```
mcmire: One thing and I'll merge this -- what do you think about extracting this conditional to another method (or at the very least a variable)? The indentation mars the readability a little bit here.

edwardmp: @mcmire Agreed, it does hurt legibility. Just amended the commit to put the first conditional in a variable.

mcmire:
Thanks so much!!

PR #1090, opened 2018-03-13
Update MIT-License

attritionorg:
standardizing between products. wasn't sure which one to use!


mcmire:
Haha thank you - while you're at it can you update the year? :)

attritionorg:
Generally the 'rule' (I think) is you don't update the year unless the code has been updated during that year, which I did not look at.

PR #1091, opened 2018-03-17
Change FactoryGirl by FactoryBot

Ana06:
The name of FactoryGirl was changed to FactoryBot (making all of as using it really "happy" :angry:), so it should be updated it in the documentation.


guialbuk:
Thanks, @Ana06 !

Issue #683, opened 2015-03-19
validate_presence_of(...) with allow_nil for association name


axlekb:
Was thinking of the use case to validate the presence of a `belongs_to` record, but not make it required.

An example would be: Document belongs to Signer, but the Signer is not added until after the Document is created.

Here's how I would do the validation:

``` ruby
validates :signer, :presence => true, :allow_nil => true
```

Here's how I'd expect to validate it:

``` ruby
it { is_expected.to validate_presence_of(:signer).allow_nil }
```

but `allow_nil` is not a chainable option for `validate_presence_of`, should it be?

mcmire:
Yup! (`allow_blank` should also be a chainable option too.)

axlekb:
This can be closed too!

mcmire:
:sunglasses:

aprescott:
It looks like #695 (which was going to implement this) was closed. Is this feature implemented? I'm looking to use it!

mockdeep:
Just came across this. Would love to have this matcher. We want to have values be either `nil` or anything that's not an empty string.

mcmire:
I'm going to reopen this for now. If someone could re-create the PR but provide a better use case than combining`validate_presence_of` with `belongs_to`, then that would be a good start.

aprescott:
I've encountered this a few times, for cases where I have an attribute that should either be `nil` or a non-`blank?` string:

```ruby
validates :reason, presence: { allow_nil: true }
```

The work-around has been to manually cover a few cases:

```ruby
it { should allow_value(nil).for(:reason) }
it { should_not allow_value("").for(:reason) }
it { should_not allow_value(" ").for(:reason) }
```

I imagine, like the opening description, that this could instead be something like `validate_presence_of(:reason).allow_nil`. It could test the right cases without having to copy-paste this for any attribute that wants to validate in this way.

mockdeep:
Being that we are receiving user input from client-side, which comes back as empty string rather than `null`, we ended up opting for a setter method instead of the validation:

```ruby
def thing=(thing)
  super(thing.presence)
end
```

mcmire:
Sorry for the super long delay on this. This is now available in 4.1.0.

Issue #904, opened 2016-02-20
undefined method `attribute_setter' for nil:NilClass


Draiken:
Hello!

I'm receiving an error when a validation fails with the `validates_inclusion_of` matcher. If the validation is present, everything is fine, but without it when building the failure message you get this error:

```
NoMethodError:
       undefined method `attribute_setter' for nil:NilClass
     # ./.gem/ruby/2.3.0/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/allow_value_matcher.rb:405:in `failure_message'
     # ./.gem/ruby/2.3.0/gems/activesupport-4.2.5.1/lib/active_support/core_ext/object/try.rb:77:in `public_send'
     # ./.gem/ruby/2.3.0/gems/activesupport-4.2.5.1/lib/active_support/core_ext/object/try.rb:77:in `try!'
     # ./.gem/ruby/2.3.0/gems/activesupport-4.2.5.1/lib/active_support/core_ext/object/try.rb:63:in `try'
     # ./.gem/ruby/2.3.0/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validation_matcher.rb:129:in `failure_reason'
     # ./.gem/ruby/2.3.0/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validation_matcher.rb:56:in `block in failure_message'
     # ./.gem/ruby/2.3.0/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validation_matcher.rb:55:in `tap'
     # ./.gem/ruby/2.3.0/gems/shoulda-matchers-3.1.1/lib/shoulda/matchers/active_model/validation_matcher.rb:55:in `failure_message'
     # ./spec/lib/models/address_spec.rb:22:in `block (2 levels) in <module:Models>'
```

My class and spec look close to this:

```
class Thing
  include Virtus.model
  include ActiveModel::Model

  ARRAY = ["value"] 

  attribute :thing, String
end

RSpec.describe Thing do
  it { is_expected.to validate_inclusion_of(:thing).in_array(Thing::ARRAY) }
end
```

I am using active model outside of rails, configuring `Shoulda` to integrate with `RSpec` and `ActiveModel`. Ruby version is 2.3.0 and ActiveModel version is 4.2.5.1

Maybe this is an incompatibility with Virtus that I'm not aware of?

Thanks

halilim:
[...]

PR #1100, opened 2018-04-17
Resolves #1097: Add `allow_nil` qualifier to `validate_presence_of`

### LINKED ISSUE #1097: Add `allow_nil` qualifier to `validate_presence_of`
Rails supports `validates :some_attribute, presence: { allow_nil: true }` on models, but currently the `validate_presence_of` matcher doesn't have the `allow_nil` qualifier.

It will be much simpler to write specs with this added.
Please let me know if it is not added for some reasons.

cjmao:
Please refer to [#1097][].
Also related: [#683][].

[#1097]: /thoughtbot/shoulda-matchers/issues/1097
[#683]:  /thoughtbot/shoulda-matchers/issues/683

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -57,6 +57,27 @@ module Shoulda
       #
       # #### Qualifiers
       #
+      # ##### allow_nil
+      #
+      # Use `allow_nil` if your model has an optional attribute.
+      #
+      #   class Robot
+      #     include ActiveModel::Model
+      #     attr_accessor :nickname
+      #
+      #     validates_presence_of :nickname, allow_nil: true
+      #   end
+      #
+      #   # RSpec
+      #   RSpec.describe Robot, type: :model do
+      #     it { should validate_presence_of(:nickname).allow_nil }
+      #   end
+      #
+      #   # Minitest (Shoulda)
+      #   class RobotTest < ActiveSupport::TestCase
+      #     should validate_presence_of(:nickname).allow_nil
+      #   end
+      #
       # ##### on
       #
       # Use `on` if your validation applies only under a certain context.
@@ -114,6 +135,16 @@ module Shoulda
         def initialize(attribute)
           super
           @expected_message = :blank
+          @expects_to_allow_nil = false
+        end
+
+        def allow_nil
+          @expects_to_allow_nil = true
+          self
+        end
+
+        def expects_to_allow_nil?
+          @expects_to_allow_nil
         end
 
         def matches?(subject)
@@ -122,6 +153,9 @@ module Shoulda
           if secure_password_being_validated?
             ignore_interference_by_writer.default_to(when: :blank?)
             disallows_and_double_checks_value_of!(blank_value, @expected_message)
+          elsif expects_to_allow_nil?
+            allows_value_of(nil) &&
+              disallows_original_or_typecast_value?(blank_value, @expected_message)
           else
             disallows_original_or_typecast_value?(blank_value, @expected_message)
           end
@@ -146,12 +180,14 @@ module Shoulda
         end
 
         def disallows_original_or_typecast_value?(value, message)
-          disallows_value_of(blank_value, @expected_message)
+          disallows_value_of(value, message)
         end
 
         def blank_value
           if collection?
             []
+          elsif expects_to_allow_nil?
+            ''
           else
             nil
           end
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -152,12 +152,18 @@ module Shoulda
 
           if secure_password_being_validated?
             ignore_interference_by_writer.default_to(when: :blank?)
-            disallows_and_double_checks_value_of!(blank_value, @expected_message)
+            check_values(disallowed_values) do |value|
+              disallows_and_double_checks_value_of!(value, @expected_message)
+            end
           elsif expects_to_allow_nil?
             allows_value_of(nil) &&
-              disallows_original_or_typecast_value?(blank_value, @expected_message)
+              check_values(disallowed_values) do |value|
+                disallows_original_or_typecast_value?(value, @expected_message)
+              end
           else
-            disallows_original_or_typecast_value?(blank_value, @expected_message)
+            check_values(disallowed_values) do |value|
+              disallows_original_or_typecast_value?(value, @expected_message)
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -146,12 +180,14 @@ def disallows_and_double_checks_value_of!(value, message)
         end
 
         def disallows_original_or_typecast_value?(value, message)
-          disallows_value_of(blank_value, @expected_message)
```
cjmao: This seems to be a misuse.

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -146,12 +180,14 @@ def disallows_and_double_checks_value_of!(value, message)
...
+          disallows_value_of(value, message)
         end
 
         def blank_value
           if collection?
             []
+          elsif expects_to_allow_nil?
+            ''
```
cjmao: Maybe need a better implementation?

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
mcmire: Yup, you're right!

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
mcmire: Now that I think about it, perhaps the matcher should work like this?

* If the matcher is qualified with `allow_nil`, then it should test that empty-string and nil are disallowed.
* However, if it _is_ qualified with `allow_nil`, then it should only test that empty-string is disallowed (because it tests that nil is allowed).

What do you think about that?

cjmao: IMO, both a blank string and a `nil` value return `true` to `#blank?`,
so as long as the `ActiveModel::Validations::PresenceValidator` uses `#blank?` to check the validity of an attribute, both a true blank string and a `nil` can be considered a blank string (when the attribute's type is either `String` or `Nil`).

mcmire: > Perhaps what you mean is, when the matcher is not qualified with allow_nil it should disallow both `''` and `nil`?

Yes, that's what I meant :)

> IMO, both a blank string and a nil value return true to #blank?

Right, exactly. As it stands, the presence matcher only uses `nil` in making its assertion, and in your changes here you conditionally flip that to an empty string, but really, in the default case where `allow_nil` is not specified, the presence matcher needs to make use of both, since PresenceValidator is not just testing that the attribute is nil but is testing that it is blank.

cjmao: What do you think about this:

```diff
 def matches?(subject)
   super(subject)

   if secure_password_being_validated?
     ignore_interference_by_writer.default_to(when: :blank?)
     disallows_and_double_checks_value_of!(blank_value, @expected_message)
-  elsif expects_to_allow_nil?
-    allows_value_of(nil) &&
-      disallows_original_or_typecast_value?(blank_value, @expected_message)
   else
-    disallows_original_or_typecast_value?(blank_value, @expected_message)
+    disallowed_values
+      .map { |value| disallows_original_or_typecast_value?(value, @expected_message) }
+      .reduce { |prev, curr| prev && curr }
   end
 end
```

```diff
-def blank_value
+def disallowed_values
+  disallowed = []
+
   if collection?
-    []
+    disallowed << []
   elsif expects_to_allow_nil?
-    ''
+    disallowed << ''
   else
-    nil
+    disallowed << nil
+    disallowed << ''
   end
+
+  disallowed
 end
```

mcmire: That's good! Perhaps instead of `reduce` you could use `all?` And I don't think you have to write any tests to cover this -- we are changing how the matcher works internally, and as long as these changes still make the current tests pass then you should be good.

cjmao: I have tried to use the code similar to the previous example to see if it still passes the tests.
However, after I ran the specs, there is one failure. And I think the cause might be testing agains an empty string when not qualified with `allow_nil`.
So I checked out `c0960bd7` (the commit right before this PR), and made the following modification only:
```diff
 def blank_value
   if collection?
     []
   else
-    nil
+     ''
   end
 end
```
After I ran the specs again, there are several failed specs expecting an assertion to fail with a message but the assertion did not fail. 
So I think maybe we should not explicitly disallow an empty string when not using `allow_nil`.

cjmao: Also, there are several places that have the following example:
```ruby
it_supports(
  'ignoring_interference_by_writer',
  tests: {
    accept_if_qualified_but_changing_value_does_not_interfere: {
      changing_values_with: :nil_to_blank
    },
    reject_if_qualified_but_changing_value_interferes: {
      model_name: 'Example',
      attribute_name: :attr,
      changing_values_with: :never_falsy,
      expected_message: <a message>
    }
  }
)
```
Though I don't fully understand how this works, I think this might already have covered the case where setting an attribute to an empty string: `changing_values_with: :nil_to_blank `.

mcmire: Can you go ahead and push up your modifications so that I can see the test failures? That way I can debug them for you.

Also in response to your last comment, that test is making sure that if you have an attribute that changes nil to an empty string when it is assigned, that this won't cause the presence validation to fail. I guess it's true that this is implicitly testing the empty string case, but I think it would be more clear to have an explicit test case / context so that it would be easier to find if we have to do so later.

cjmao: Sure, I've just pushed my code.

mcmire:
This looks great. The only issue with this is that it looks like we are running into #904 on one of the tests. This isn't your fault, but it does mean that I would have to get a fix for that into master before merging this so that CI can run cleanly. Sorry :(

cjmao:
Yes of course.

Also, I've noticed these lines:

```ruby
  ...
  elsif expects_to_allow_nil?
    allows_value_of(nil) &&
      check_values(disallowed_values) do |value|
        disallows_original_or_typecast_value?(value, @expected_message)
      end
  else
  ...
```
Do we need to add a message to `allows_value_of(nil)`?

mcmire:
#904 will be fixed soon, so this is just a note that in in order to merge this, we are going to need these changes to show up in `does_not_match?` as well as per #1136.

mcmire:
Sorry it took me so long to merge this, but I took your ideas here and ran with them: 834d8d0356573b9f47e63a1b910cfa8f3d815e51. Thanks so much!

PR #1103, opened 2018-05-10
validate_length_of now works with numeric columns

### LINKED ISSUE #1094: validate_length_of doesn't work with numeric columns
Hi, i have this validations
```
  validates :long_description, length: { minimum: 50 }
  validates :general_sum, length: { maximum: 6 }
```
And i have this rspec test
```
describe '#validations' do
    it { should validate_length_of(:long_description).is_at_least(50) }
    it { should validate_length_of(:general_sum).is_at_most(6) }
  end
```
test for minimum, works correct but on maximum i got this error:

```
    should validate that the length of :long_description is at least 50
    should validate that the length of :general_sum is at most 6 (FAILED - 1)

Failures:

  1) Initiative#validations should validate that the length of :general_sum is at most 6
     Failure/Error: it { should validate_length_of(:general_sum).is_at_most(6) }
     
       Initiative did not properly validate that the length of :general_sum is
       at most 6.
         After setting :general_sum to ‹"xxxxxxx"› -- which was read back as
         ‹0› -- the matcher expected the Initiative to be invalid, but it was
         valid instead.
     
         As indicated in the message above, :general_sum seems to be changing
         certain values as they are set, and this could have something to do
         with why this test is failing. If you've overridden the writer method
         for this attribute, then you may need to change it to make this test
         pass, or do something else entirely.
     # ./spec/models/initiative_spec.rb:38:in `block (3 levels) in <top (required)>'

```

What am i doing wrong? Thanks

novotarq:
fixes #1094

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
@@ -3,23 +3,31 @@ require 'unit_spec_helper'
 describe Shoulda::Matchers::ActiveModel::ValidateLengthOfMatcher, type: :model do
   context 'an attribute with a non-zero minimum length validation' do
     it 'accepts ensuring the correct minimum length' do
-      expect(validating_length(minimum: 4)).
-        to validate_length_of(:attr).is_at_least(4)
+      attribute_types(minimum: 4).each do |options|
+        expect(validating_length(options)).
+          to validate_length_of(:attr).is_at_least(4)
+      end
     end
 
     it 'rejects ensuring a lower minimum length with any message' do
-      expect(validating_length(minimum: 4)).
-        not_to validate_length_of(:attr).is_at_least(3).with_short_message(/.*/)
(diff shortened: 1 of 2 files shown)


spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
```diff
@@ -3,23 +3,31 @@
 describe Shoulda::Matchers::ActiveModel::ValidateLengthOfMatcher, type: :model do
   context 'an attribute with a non-zero minimum length validation' do
     it 'accepts ensuring the correct minimum length' do
-      expect(validating_length(minimum: 4)).
-        to validate_length_of(:attr).is_at_least(4)
+      attribute_types(minimum: 4).each do |options|
```
mcmire: Just a note here - instead of embedding loops inside of tests, we typically place the loops outside of the tests. See the tests for the numericality matcher for how to achieve this. (We can do this when we merge this in as it's a bit complicated.)

mcmire:

PR #1109, opened 2018-06-12
Handle uuid columns in validate_absence_of

ketiko:
There is not an explicit value set for uuid columns in the
validate_absence_of matcher.  The default string value is used which
will get set to nil instead when the type cast happens.  This causes the
matcher to fail when testing that the model does indeed validate for
absence of that attribute.


guialbuk:
Thanks @ketiko 👍

PR #1117, opened 2018-07-20
Fix has_secure_password checks for Rails 6

leemachin:
If you're trying out Rails 6 and using shoulda-matchers, then this PR in particular will cause problems: https://github.com/rails/rails/commit/08dde0f355e0450bc2cb7e0b4b95a0ec18d1a870#diff-a6f9af64a6d420eb1103e16c0a0a8f04

`ActiveModel::SecurePassword:: InstanceMethodsOnActivation` is no longer included.

This PR simply makes the test for that class more specific, so it still works for stable versions of Rails.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -139,7 +139,7 @@ module Shoulda
 
             @attribute == :password
           else
-            auth_attr = @subject.instance_methods.find {|meth| meth.to_s =~ /authenticate_[\w_]+/ }
+            auth_attr = @subject.class.instance_methods.find {|meth| meth.to_s =~ /authenticate_[\w_]+/ }
 
             return false unless auth_attr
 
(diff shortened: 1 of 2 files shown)


lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -134,9 +134,10 @@ def simple_description
         private
 
         def secure_password_being_validated?
-          defined?(::ActiveModel::SecurePassword) &&
-            @subject.class.ancestors.include?(::ActiveModel::SecurePassword::InstanceMethodsOnActivation) &&
-            @attribute == :password
+          return false unless defined?(::ActiveModel::SecurePassword::InstanceMethodsOnActivation)
```
eugeneius: This will cause `secure_password_being_validated?` to always return false on Rails 6 - but presumably this method exists for a reason, and it should continue to return true when the attribute being validated is a secure password.

mcmire:
Hey @leemachin, thanks for making this PR. We're about to release 4.0.0, so I'm not sure we can include this just yet, but considering that Rails 6 is around the corner, I'll make a note of this so we can include this soon.

connorshea:
Now that Rails 6 betas are coming out, can this be merged? :)

mcmire:
We can't merge this quite yet as we don't want to stuff 4.0 with too many things, but we'll make sure to include this in the next release.

mcmire:
@leemachin Thanks for this PR! I massaged it a bit and merged it in as #1193 above. ^

PR #1124, opened 2018-09-03
Allow validates_length_of to work on association - Fix #1007

### LINKED ISSUE #1007: validate_length_of does not work when used with associations

prashantjois:
if the length check is on an association then create an array to assign to the association instead of a string

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
@@ -398,11 +398,33 @@ module Shoulda
         end
 
         def allows_length_of?(length, message)
-          allows_value_of(string_of_length(length), message)
+          allows_value_of(attribute_of_length(length), message)
         end
 
         def disallows_length_of?(length, message)
-          disallows_value_of(string_of_length(length), message)
+          disallows_value_of(attribute_of_length(length), message)
+        end
+
+        def attribute_of_length(length)
+          attribute_is_association? ? association_of_length(length) : string_of_length(length)
+        end
+
+        def attribute_is_association?
+          !!attribute_association_reflection
+        end
+
+        def attribute_association_reflection
+          if @subject.class.respond_to?(:reflect_on_association)
+            @subject.class.reflect_on_association(@attribute)
+          end
+        end
+
+        def association_of_length(length)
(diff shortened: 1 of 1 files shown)


StragaSevera:
It would be great to have this feature in the main gem!

KapilSachdev:
@mcmire is there anything pending for this PR?

mcmire:
@KapilSachdev Yes. In order to merge this addition, there need to be tests. Essentially, all of the scenarios represented within `validate_length_of_spec` and all of the tests for the qualifiers need to be updated to also account for associations. This includes:

* Validation option `minimum` + qualifier `is_at_least`
* Validation option `maximum` + qualifier `is_at_most`
* Validation option `is` + qualifier `is_equal_to`
* `allow_nil`
* `allow_blank`
* message `with_message`
* `with_short_message`
* `with_long_message`

KapilSachdev:
My bad, i missed the `PR: Needs Tests` label.
As the OP hasn't responded and in case anyone else does not wants to work, I will move this PR forward asap, .

mcmire:
@KapilSachdev No worries, I should have said something initially when adding that tag. Thank you, that would be much appreciated!

matsales28:
Closing this PR as this was implemented on #1569.

PR #1128, opened 2018-09-05
Enable chain after allow_nil in validates_inclusion_of matcher

willnet:
78a482f995db719e8fad6d9a5dae4a3fcc727c18 make `allow_nil` disable to chain because it forgot `return self`


formigarafa:
This PR is I am waiting for this PR.
meanwhile `.tap(&:allow_nil)` is doing the job.

mcmire:

PR #1133, opened 2018-09-11
Don't require bootsnap in gen'ed Rails 5.2+ apps

mcmire:
[Bootsnap][1] is a new gem that speeds up load times for Rails apps and
is included by default in Rails 5.2+ apps. However, this somehow messes
with Zeus, which we use often to run tests locally. We don't need it in
the Rails apps that the tests generate, so make sure it's not required
in those apps.

[1]: https://github.com/Shopify/bootsnap

### DIFF
--- a/spec/support/unit/rails_application.rb
+++ b/spec/support/unit/rails_application.rb
@@ -75,6 +75,7 @@ module UnitTests
     def generate
       rails_new
       fix_available_locales_warning
+      remove_bootsnap
       write_database_configuration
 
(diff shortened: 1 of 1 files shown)


guialbuk:
No problem 👍

voxik:
Wouldn't be better to do something like:

~~~
sed -i '/rails new/ s/)/ --skip-bootsnap)/' spec/support/unit/rails_application.rb
~~~

PR #1135, opened 2018-09-12
Fix inclusion to correctly disallow outside values

mcmire:
The inclusion matcher, when qualified with `in_array`, was using
AllowValueMatcher to check that values outside the array were disallowed
by the model (and then inverting its result). However, it should have
been using DisallowValueMatcher all this time. This commit fixes that.

Without this fix, the following error is raised when using the inclusion
matcher against a model which does not have the proper inclusion
validation on it:

    undefined method `attribute_setter' for nil:NilClass

This happens because the inclusion matcher is a complex matcher, i.e.,
it runs a series of submatchers internally and the result of those
submatchers contributes to whether or not the matcher matches. If one of
those submatchers fails, the inclusion matcher immediately fails and
spits out the failure message associated with that submatcher.

However, there is a fundamental difference between AllowValueMatcher and
DisallowValueMatcher as it relates to how they function:

* AllowValueMatcher sets an attribute to a value on a record and expects
  the record not to fail validation.
* DisallowValueMatcher sets an attribute to a value on a record, but
  expects the record *to* fail validation.

The issue in this case is that, because AllowValueMatcher was used
instead of DisallowValueMatcher, the inclusion matcher thought that the
AllowValueMatcher failed, when in fact it passed (this result was just
inverted). So it tried to generate a failure message for a matcher that
didn't fail in the first place. By using DisallowValueMatcher, we set
the proper expectations.

---

Related issue: #904

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,6 +70,11 @@ is now:
   * *PR: [#1073]*
   * *Original issue: [#949]*
 
+* Fix `validate_inclusion_of` so that if it fails, it will no longer blow up
+  with the error "undefined method \`attribute_setter' for nil:NilClass".
+
+  * *Original issue: [#904]*
+
 ### Features
 
 * Add `required` and `optional` qualifiers to `belong_to` and `have_one`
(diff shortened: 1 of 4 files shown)


guialbuk:
Fixed by #1136

mcmire:
Oop, okay, I was going to merge this after I rebased the other branch, as it contained some other non-related stuff. Just give me a thumbs-up next time and I can make sure to sort them out :)

PR #1137, opened 2018-09-13
Simplify Rubocop configuration

mcmire:
The original Rubocop configuration was copied from other thoughtbot
projects. Of course, Rubocop comes with its own set of default settings.
Therefore, we don't need to re-specify settings that it already has --
we only need to specify the ones that we're overriding.

Besides this, Rubocop isn't even in the Gemfile, so even though Hound
runs on GitHub automatically, it'd be nice to spot style issues as you
are editing files.

Given this, this commit simplifies the Rubocop configuration and also
adds Rubocop to the Gemfile so that you can configure your editor to
lint files as you are editing them.


Gemfile
```diff
@@ -6,10 +6,11 @@ gem 'pry'
 gem 'pry-byebug'
 gem 'rake', '12.3.0'
 gem 'rspec', '~> 3.6'
+gem 'rubocop'
```
guialbuk: It's recommended to use `require: false` for Robocop in Gemfile
https://github.com/rubocop-hq/rubocop#installation

mcmire:
@guialbuk Should be fixed now, and I also rebased. Look good now?

guialbuk:
Looks good 👍

Just a quick question. Do you think we should add `rubocop required: false` to Appraisals file as a shared dependency for all generated Gemfiles?
https://github.com/thoughtbot/shoulda-matchers/blob/a465e796e88c8e73be67b9b739473e67707674ef/Appraisals#L14-L17

mcmire:
@guialbuk Oops, sorry, looks like I never responded to this. The goal of adding Rubocop to the Gemfile is to give you the `rubocop` executable that you can make use of in your editor. The tests don't need it to run, so I don't know we would want to put it inside the Appraisals. Does that make sense?

guialbuk:
Yeah, makes sense 😊
So it would be better to remove them from `gemfiles/*.gemfile`
Whay do you think?

mcmire:
@guialbuk Oh, I think those lines are getting added automatically to the gemfiles, anyway. I think what Appraisal does is take your Gemfile and combine with the Appraisal gem set to create the appraisal-specific gemfiles. So I could remove them from those gemfiles, but they will just get added back when you run `appraisal install`.

guialbuk:
I see. So it shouldn't be a problem then. 👍

PR #1144, opened 2018-10-02
Add with_failure_message qualifier and deprecate with_message

### LINKED ISSUE #1022: Wording `allow_value` .. `with_message` does not read well
The following usage of `allow_value` feels very strange to read:
```
it { is_expected.to allow_value('Suzuki').for(:brand).with_message('is out of business!') }
```
Everything makes sense up until `.with_message`. That method's argument is not used as some accompanying input to the value ('Suzuki'), instead it represents an expected response.

Perhaps we can alias `#with_message` with a clearer name? Maybe something like `#and_return`?
ie.
```
it { is_expected.to allow_value('Suzuki').for(:brand).and_return('is out of business!') }
```

It is important that one-liners be as readable as possible, and I feel like `allow_value` falls a little short in this example.

indocomsoft:
Fixes #1022

### DIFF
--- a/lib/shoulda/matchers/active_model/allow_value_matcher.rb
+++ b/lib/shoulda/matchers/active_model/allow_value_matcher.rb
@@ -225,8 +225,8 @@ module Shoulda
       #         with_failure_message(/open or closed/)
       #     end
       #
-      # Use `with_failure_message` with the `:against` option if the attribute the
-      # validation message is stored under is different from the attribute
+      # Use `with_failure_message` with the `:against` option if the attribute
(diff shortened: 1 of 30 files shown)


mcmire:
Hey @indocomsoft. This looks good, although just a heads-up that we're about to release a new major version and we've frozen the codebase barring any issues with that. So I'm not sure that it's an opportune time to add this now, but we'll consider this for the next version.

indocomsoft:
Yup, I understand -- very eager to start using v4.0.0

Just thought I should contribute back to a gem that has helped me so much.

mcmire:

PR #1148, opened 2018-10-07
Set messages to :inclusion only for range and not array

### LINKED ISSUE #1146: 4.0.0 rc1 breaks inclusion_of matchers with custom message
Here's the test which breaks

``` ruby
it do
  is_expected.to validate_inclusion_of(:gender).
    in_array(Forms::Detainee::GENDERS)
end
```

and here's the form code:

``` ruby
module Forms
  class Detainee < Forms::Base
    GENDERS = %w[male female unknown indeterminate].freeze
    validates :gender, inclusion: { in: GENDERS }
  end
end
```

And heres the test failure:

```
1) Forms::Detainee#validate should validate that :gender is either ‹"male"›, ‹"female"›, ‹"unknown"›, or ‹"indeterminate"›
     Failure/Error:
       is_expected.to validate_inclusion_of(:gender).
         in_array(Forms::Detainee::GENDERS) #.

       Expected detainee to validate that :gender is either ‹"male"›,
       ‹"female"›, ‹"unknown"›, or ‹"indeterminate"›, but this could not be
       proved.
         After setting :gender to ‹"shoulda-matchers test string"›, the matcher
         expected the detainee to be invalid and to produce the validation
         error "^Answer Gender" on :gender. The record was indeed invalid, but
         it produced these validation errors instead:

         * surname: ["^Enter a last name", "^Enter a last name", "^Enter a last
           name", "^Enter a last name", "^Enter a last name"]
         * forenames: ["^Enter first name(s)", "^Enter first name(s)", "^Enter
           first name(s)", "^Enter first name(s)", "^Enter first name(s)"]
         * date_of_birth: ["^Enter a valid date of birth", "^Enter a valid date
           of birth", "^Enter a valid date of birth", "^Enter a valid date of
           birth", "^Enter a valid date of birth"]
         * gender: ["^Choose a gender"]
     # ./spec/forms/detainee_spec.rb:40:in `block (3 levels) in <top (required)>'
```

importantly it is looking for the error message `^Answer Gender`
But we have a localised error message `^Choose a gender`

So I believe the issue is that it no longer asserts against the correct localised error message.

Relevant bits of our locale file:

``` yaml
en:
 errors: &errors
    messages:
      inclusion: "^Answer %{attribute}"
    models:
      detainee:
        attributes:
          gender:
            inclusion: "^Choose a gender"
```

I can fix the test by explicitly setting the expected message

``` ruby
it do
  is_expected.to validate_inclusion_of(:gender).
    in_array(Forms::Detainee::GENDERS).
    with_message(I18n.t('errors.models.detainee.attributes.gender.inclusion'))
end
```

however this seems wrong, as it worked before.

I did a (manual) git bisect,
and the error seems to be introduced in or around this commit

https://github.com/thoughtbot/shoulda-matchers/commit/507dc570047d671459b7d4b5e714972baffb2b3e

Thanks.

indocomsoft:
Fixes #1146 

The problem is, I suspect, that before 507dc57, `@low_message ||= :inclusion` and `@high_message ||= :inclusion` are set only for range and not for array. However, on 507dc57, both `@low_message` and `@high_message` are set to `:inclusion` regardless of whether it's array or range. This PR should restore the old behaviour.

Confirmed working by @matthewrudy

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -308,8 +308,6 @@ EOT
           @range = range
           @minimum = range.first
           @maximum = range.max
-          @low_message ||= :inclusion
-          @high_message ||= :inclusion
           self
         end
 
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -295,8 +295,8 @@ EOT
           @range = nil
           @minimum = nil
           @maximum = nil
-          @low_message = :inclusion
-          @high_message = :inclusion
+          @low_message = nil
+          @high_message = nil
         end
 
         def in_array(array)
@@ -308,6 +308,8 @@ EOT
           @range = range
           @minimum = range.first
           @maximum = range.max
+          @low_message ||= :inclusion
+          @high_message ||= :inclusion
           self
         end
 
--- a/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
@@ -432,7 +432,10 @@ describe Shoulda::Matchers::ActiveModel::ValidateInclusionOfMatcher, type: :mode
 
     context 'given nil' do
       it 'is as if with_message had never been called' do
-        builder = build_object_allowing(valid_values)
+        builder = build_object_allowing(
+          valid_values,
+          validation_options: { message: 'this message should not matter' }
+        )
 
         expect_to_match_on_values(builder, valid_values) do |matcher|
           matcher.with_message(nil)


matthewrudy:
works for me
:+1:

mcmire:
@indocomsoft Would you mind seeing if you can add a test for this?

indocomsoft:
Hi, sorry I've been quite busy. I have just added the test.

I am also removing the code that sets `@low_message` and `@high_message` to `:inclusion` as I don't find anywhere else in the code that checks for this. So what I did is I just changed the part of the `with_message(nil)` test in `validate_inclusion_of` to ensure that any message is ignored. Do let me know if I need to make any other change.

mcmire:
@indocomsoft Thanks for the PR. I'm a little perplexed at this change, however, given the original issue that was reported. The reason I made the change you noticed to begin with is that I felt like `:inclusion` _should_ be the default message for both array and range (after all, if you look at the [source code for the inclusion validation](https://github.com/rails/rails/blob/d684778a000cdcfcac53f6390c565c6a91a3391e/activemodel/lib/active_model/validations/inclusion.rb#L12) you will find that it always uses the same error message regardless of what `:in` is). Now, speaking on #1146, what was reported, of course, was that the matcher seemed to be looking for a different validation message, but I don't think this was due to whether `in_array` or `in_range` was used, but rather, the issue was the default message for `:inclusion` was overridden in `en.yml` and the matcher did not pick up on this. So I think the problem with the inclusion matcher is its interpretation of what the `:inclusion` message is, and I think that's what needs to be fixed here. Furthermore, because the inclusion matcher inherits from ValidationMatcher, my sense is that this may not be just a problem with inclusion, but with a bunch of other matchers, too.

So I appreciate your change, but it seems a little suspicious to me that clearing out `@low_message` and `@high_message` would magically fix the problem. I think I need to look into this further. I'll leave this PR open in the meantime until I've gathered more information about this.

mcmire:
I appreciate the fix for this, but I'm closing this on account of the fact that I can't reproduce this, as I mentioned here: https://github.com/thoughtbot/shoulda-matchers/issues/1146#issuecomment-459192103. Perhaps more insight will appear later.

PR #1162, opened 2018-12-21
Make have_db_index_matcher respect the connection of individual model

haopingfan:
Hi Sir, 

Problem
-
I established the connections of a group of models to another database.
However, I ran into problem when I try to use `have_db_index_matcher` on those models.

`ActiveRecord::StatementInvalid, Table doesn't exist`

Solution
-
The `have_db_index_matcher` should respect the connection of individual model instead of using the connection of `ActiveRecord::Base` directly.


mcmire:

PR #1163, opened 2019-01-03
ValidateInclusionOfMatcher#allow_nil should return `self`

danfrenette:
It seems that this method was returning `self` explicitly until [this change](https://github.com/thoughtbot/shoulda-matchers/commit/97e25692537777cf33ea35ad1d4bd18b7de9a777#diff-3706e19f08c07fb311fbf18e8fce5629), but overwriting the method seems to resolve the failure I've been getting.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -322,6 +322,7 @@ EOT
 
         def allow_nil
           @options[:allow_nil] = true
+          self
         end
 
         def expects_to_allow_nil?
--- a/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
@@ -2,25 +2,6 @@ require 'unit_spec_helper'
 
 describe Shoulda::Matchers::ActiveModel::ValidateInclusionOfMatcher, type: :model do
   shared_context 'for a generic attribute' do
-    def self.testing_values_of_option(option_name, &block)
-      [nil, true, false].each do |option_value|
-        context_name = "+ #{option_name}"
-        option_args = []
-        matches_or_not = ['matches', 'does not match']
-        to_or_not_to = [:to, :not_to]
-
-        unless option_value == nil
-          context_name << "(#{option_value})"
-          option_args = [option_value]
-        end
-
-        if option_value == false
-          matches_or_not.reverse!
-          to_or_not_to.reverse!
-        end
-      end
-    end
-
     context 'against an integer attribute' do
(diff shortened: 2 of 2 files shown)


trevorrjohn:
Any status objections to this being merged?

mcmire:
It looks like there's some issues with CI not fully running the tests. I just restarted the build so maybe that will magically make it work.

guialbuk:
It seems TravisCI is using bundler `2.0.1` which conflicts with some Gemfile.lock files

```
rails (= 4.2.10) was resolved to 4.2.10, which depends on
             bundler (< 2.0, >= 1.3.0)
     
         Current Bundler version:
           bundler (2.0.1)
````

And more issues are happening with Ruby 2.2, which I think we could drop support for Ruby 2.2, as its official support ended  six months ago: https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

What do you think?

trevorrjohn:
> It seems TravisCI is using bundler `2.0.1` which conflicts with some Gemfile.lock files
> 
> ```
> rails (= 4.2.10) was resolved to 4.2.10, which depends on
>              bundler (< 2.0, >= 1.3.0)
>      
>          Current Bundler version:
>            bundler (2.0.1)
> ```
> And more issues are happening with Ruby 2.2, which I think we could drop support for Ruby 2.2, as its official support ended six months ago: https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/
> 
> What do you think?

Definitely think dropping support makes sense, probably should be in another issue/PR.

guialbuk:
Yeah. I'll open a PR to drop Ruby 2.2 support and another one to fix the Rails 4.2 + Bundle 2.0  issue. 👍

mcmire:
Yeah that sounds good @guialbuk.

mlodato517:
Is this a duplicate of #1128 ?

mlodato517:
This looks like a duplicate of #1128

mrleeio:
Any way I can help get this fix into master?  Or with the 4.0 release in general?

Working on a Ruby 2.4.x upgrade and shoulda-matcher is the last blocker.

Reminder that Ruby 2.3 loses security patches soon, something I need to care about at my organization.

> Ruby 2.3
status: security maintenance
release date: 2015-12-25
EOL date: scheduled for 2019-03-31

mcmire:
@mrleeio If you could test 4.0.0.rc1 and let me know if you see any other Ruby 2.4 or Rails 5 issues that would be great! I'm probably going to release 4.0 in the next week or so.

mrleeio:
@mcmire this issue seems to be the only one impacting our codebase on Rails 4.x, happy to toss more updates your way once I start looking at Rails 5.

Thank you for all the hard work, looking forward to that release.

danfrenette:
🎉 Thanks for the merge @mcmire, super grateful to have contributed to something my team and I use on multiple projects every day!

PR #1164, opened 2019-01-09
Remove Ruby 2.2 support

guialbuk:
Official support for Ruby 2.2 ended in June 2018, so we are dropping support for it as well.

Ref: https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ rvm:
   - 2.5.1
   - 2.4.4
   - 2.3.7
-  - 2.2.8
 
 gemfile:
   - gemfiles/4.2.gemfile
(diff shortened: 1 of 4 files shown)


mcmire:
Nice! Can you update NEWS.md as well? Also, it looks like the Ruby 2.3.7 / Rails 4.2 tests are failing. We might need to update the corresponding appraisal to lock to a specific version of Bundler, I'm not sure.

mcmire:
I'm going to work on getting this merged in if you don't mind.

By the way, since you have commit access, you should be able to make a branch straight on this repo instead of forking the repo :) That should help if one of us wants to take the other's branch over.

Issue #1167, opened 2019-01-20
validate_presence_of fails on Rails 6


SimplySerenity:
validate_presence_of relies on the constant ActiveModel::SecurePassword::InstanceMethodsOnActivation

which has been removed in Rails 6 https://github.com/rails/rails/commit/08dde0f355e0450bc2cb7e0b4b95a0ec18d1a870#diff-a6f9af64a6d420eb1103e16c0a0a8f04L66

nicholasshirley:
Found the same.

Looks like the third check could also return false when it should return true because the attribute can be assigned by the user now instead of always being `password`.

 https://github.com/thoughtbot/shoulda-matchers/blob/f951455f0f66ebc09a80bb20089b1b253c2ac0fb/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb#L149-L153

joshRpowell:
related https://github.com/thoughtbot/shoulda-matchers/pull/1117

yagudaev:
It would be really nice to have this fixed with new version of rails around the corner

mcmire:
@yagudaev We plan on including support for Rails 6 in the next immediate version, so this is included in the list of upcoming changes.

nscmnto:
@yagudaev thanks for the workaround!

mcmire:
This is fixed now! ^

connorshea:
@mcmire can we get a new release now that Rails 6.0.0.rc1 is out? :) I just ran my test suite with the latest master on Rails 6 and it worked great 👍

PR #1169, opened 2019-01-29
Fix for validate_presence_of for Rails6 secure_password problem

itsterry:
Addresses #1167 

Logic is that Rails 6 secure password attributes require an additional attribute called [attribute]_digest

So if we're in Rails 6 and the attribute has a matching [attribute]_digest attribute, it's a secure password

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -157,7 +157,7 @@ module Shoulda
           ) &&
             @subject.class.ancestors.include?(
               ::ActiveModel::SecurePassword::InstanceMethodsOnActivation
-          ) &&
+            ) &&
             @attribute == :password
         end
 
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -121,7 +121,7 @@ module Shoulda
 
           possibly_ignore_interference_by_writer
 
-          if secure_password_being_validated?
+          if should_double_check_attribute?
             disallows_and_double_checks_value_of!(blank_value, @expected_message)
           else
             disallows_original_or_typecast_value?(blank_value, @expected_message)
@@ -133,7 +133,7 @@ module Shoulda
 
           possibly_ignore_interference_by_writer
 
-          if secure_password_being_validated?
+          if should_double_check_attribute?
             allows_and_double_checks_value_of!(blank_value, @expected_message)
           else
             allows_original_or_typecast_value?(blank_value, @expected_message)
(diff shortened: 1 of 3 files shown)


spec/support/unit/helpers/rails_versions.rb
```diff
@@ -38,5 +38,13 @@ def rails_lt_5?
     def rails_5_x?
       rails_version =~ '~> 5.0'
     end
+
+    def rails_lt_6?
```
mcmire: Good thinking, but what if we add these when we need them?

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -327,7 +327,7 @@ def model_creator
     end
   end
 
-  if rails_4_x?
+  if rails_4_x? || rails_5_x?
```
mcmire: Good thinking. In order to merge this, we will also need tests for the `#{attribute}_digest` case. Unfortunately, in order for you to even run those tests, we will need to add Rails 6 to the Appraisals file. We can't do this quite yet until we have 4.0 released, so we will have to come back around to this.

spec/support/unit/helpers/rails_versions.rb
mcmire: Yeah, let's go ahead and leave them out for now, and we can add them when it comes time.

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -147,11 +147,21 @@ def simple_description
         private
 
         def secure_password_being_validated?
-          defined?(::ActiveModel::SecurePassword) &&
+          rails_lt_6_secure_password_being_validated? ||
+            rails_6_secure_password_being_validated?
+        end
+
+        def rails_lt_6_secure_password_being_validated?
```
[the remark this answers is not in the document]
mcmire: If you don't mind changing this, that would be appreciated!

spec/support/unit/helpers/rails_versions.rb
itsterry: Have removed

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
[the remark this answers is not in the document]
itsterry: Changed per feedback

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -146,14 +146,28 @@ def simple_description
...
-        def secure_password_being_validated?
-          defined?(::ActiveModel::SecurePassword) &&
-            @subject.class.ancestors.include?(::ActiveModel::SecurePassword::InstanceMethodsOnActivation) &&
+        def should_double_check_attribute?
+          attribute_is_secure_password? ||
+            attribute_has_digest?
+        end
+
+        def attribute_is_secure_password?
+          defined?(
```
ricardotk002: I was reviewing Rails 6's new beta and stumbled with this issue too. I came up with the following question: shouldn't these checks be more optimistic about the attribute's name and check if it is `password` first, before doing the subsequent checks for `ActiveModel::SecurePassword`?

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -146,14 +146,28 @@ def simple_description
...
+            ::ActiveModel::SecurePassword::InstanceMethodsOnActivation
```
ricardotk002: Btw, this module was removed in https://github.com/rails/rails/pull/26764

mcmire:
Hi @itsterry we realized that #1117 already covered this! Sorry for not seeing this earlier. This is fixed in the PR above ^

itsterry:
Brilliant! Thanks @mcmire !

PR #1171, opened 2019-02-01
Improve README on how to integrate with RSpec

nic-lan:
Improve the documentation on how to integrate the should matchers with RSpec in non Rails projects.

Currently the docs don't specify that `shoulda-matcher` needs to be required before RSpec configuration block.

That leads to a failure when trying to `bundle exec rspec`
```
Failure/Error: config.include(Shoulda::Matchers::ActiveModel, type: :model)

NameError:
  uninitialized constant Shoulda
```

Putting `require 'shoulda-matchers'` fixed the issue in my case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ to use model matchers in certain example groups, you'll need to manually include
 them. Here's a good way of doing that:
 
 ```ruby
+require 'shoulda-matchers'
+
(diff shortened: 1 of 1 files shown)


mcmire:
Makes sense to me, thanks!

PR #1172, opened 2019-02-02
Fixes uniqueness validator syntax in documentation

sildur:
The documentation shows a wrong example of uniqueness validation:
```ruby
validates :slug, uniqueness: true, scope: :journal_id
```
The correct syntax is:
```ruby
validates :slug, uniqueness: { scope: :journal_id }
```


mcmire:
Hey, you're totally right! Thanks for correcting this.

PR #1177, opened 2019-02-20
Fix `without_validating_presence` documentation

berislavbabic:
`balongs_to.without_validating_presence` was introduced in #1170 and both the documentation and the changelog state that the method is called `without_presence_validation`.  Imho the implemented method name is better, so here is the documentation update, to avoid future confusion


mcmire:
Whoops! Wow, I must have spaced on this when I implemented this. Good catch!

Issue #1181, opened 2019-02-26
Review of PR 1180


n-rodriguez:
Hi there!

I review the https://github.com/thoughtbot/shoulda-matchers/pull/1180 : I don't know if doing a `chdir` is a good thing here. If you wanted to remove `setup` you could have removed the `executables` method call.
See Sidekiq for example : https://github.com/mperham/sidekiq/pull/4023

mcmire:
Hey, in making those changes I took a look at what Bundler is now doing whenever it generates gems and decided to borrow from that: https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt. Are you having issues with this now?

mcmire:
Left a comment on your PR :)

PR #1182, opened 2019-02-26
Improve gemspec

n-rodriguez:
See https://github.com/thoughtbot/shoulda-matchers/issues/1181

### DIFF
--- a/shoulda-matchers.gemspec
+++ b/shoulda-matchers.gemspec
@@ -1,5 +1,6 @@
-$LOAD_PATH << File.join(File.dirname(__FILE__), 'lib')
-require 'shoulda/matchers/version'
+# frozen_string_literal: true
+
+require_relative 'lib/shoulda/matchers/version'
 
 Gem::Specification.new do |s|
   s.name        = "shoulda-matchers"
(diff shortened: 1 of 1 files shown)


ashmaroli:
@n-rodriguez Apparently, the intention of #1180 was two-fold:
  - Remove `bin/setup` from being installed as a global executable.
  - Remove *every other* unnecessary stuff (including the `spec` directory) from the gem.

While I agree with you that `Dir.chdir` is not necessary, your recommendation for
```ruby
s.files = `git ls-files`.split("\n")
```
is not going to meet the second intention of #1180.

mcmire:
Hey @n-rodriguez, in making those changes I took a look at what Bundler is now doing whenever it generates gems and decided to borrow from that: https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt. What's the goal in reverting this, are you having issues with this now?

n-rodriguez:
> and decided to borrow from that: https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt

Ok, I didn't see that :+1: 

> What's the goal in reverting this, are you having issues with this now?

No. It's just I'm not comfortable with`chdir` but if even Bundler does it let's go for it :+1:

Issue #1114, opened 2018-07-02
validate_uniqueness_of with scoped_to fails when the scope is of data type time


mikeduynguyen:
The validation fails but it should pass when the scope has data type `time`. 

Example: model `Schedule` with the following attributes and their data types:
`day_of_week`: `integer`
`start_time`: `time`

```ruby
class Schedule
    validates :day_of_week, uniqueness: { scope: :start_time }
end
```

Spec:
```ruby
it { should validate_uniqueness_of(:day_of_week).scoped_to(:start_time) }

Failures:

  1) Schedule should require case sensitive unique value for day_of_week scoped to start_time
     Failure/Error: it { should validate_uniqueness_of(:day_of_week).scoped_to(:start_time) }
     
       Did not expect errors to include "has already been taken" when day_of_week is set to 0,
       got errors:
       * "has already been taken" (attribute: day_of_week, value: 0) (with different value of start_time)
```

This bug is caused by the method that calculates the next value to test for the uniqueness of the scope:
[`lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb`](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L705)
```ruby
module Shoulda
  module Matchers
    module ActiveRecord
    ...
    def next_scalar_value_for(scope, previous_value)
        ...
        elsif previous_value.respond_to?(:to_datetime)
            previous_value.to_datetime.next
    ...
```

When `previous_value` is of type `time`:

`previous_value` -> `2018-07-02 11:20:14`
` previous_value.to_datetime.next` -> `2018-07-03 11:20:14`

The date is incremented by 1, however the time remains the same. As a result when the next value is set on the new record the uniqueness validation fails since the 2 records have the same time value.

A possible fix would be to also increment the `previous_value` by 60 seconds before calling `next` to guarantee both the time and the date are different:
`previous_value.to_datetime.in(60).next`

I'm more than happy to open a PR for this fix! Let me know what you guys think about this.

mcmire:
Ah... yes that makes sense! We would definitely appreciate a PR. 👍

mikeduynguyen:
@mcmire I apologize for the looong the delay on this 😅. Finally got to open the PR to fix this bug here #1190 .

It's my first time contributing to an open source project so please let me know if you have any suggestions for the PR.

mcmire:
[...]

PR #1190, opened 2019-03-19
Bug fix: uniqueness validation fails with scoped attributes of data type time

mikeduynguyen:
**PR to fix issue #1114**  

### The bug

The validation fails but it should pass when the scope has data type `time`. 

Example: model `Schedule` with the following attributes and their data types:
`day_of_week`: `integer`
`start_time`: `time`

```ruby
class Schedule
    validates :day_of_week, uniqueness: { scope: :start_time }
end
```

Spec:
```ruby
it { should validate_uniqueness_of(:day_of_week).scoped_to(:start_time) }

Failures:

  1) Schedule should require case sensitive unique value for day_of_week scoped to start_time
     Failure/Error: it { should validate_uniqueness_of(:day_of_week).scoped_to(:start_time) }
     
       Did not expect errors to include "has already been taken" when day_of_week is set to 0,
       got errors:
       * "has already been taken" (attribute: day_of_week, value: 0) (with different value of start_time)
```

This bug is caused by the method that calculates the next value to test for the uniqueness of the scope:
[`lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb`](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L833)
```ruby
module Shoulda
  module Matchers
    module ActiveRecord
    ...
    def next_scalar_value_for(scope, previous_value)
        ...
    ...
```

When `previous_value` is of type `time`:

`previous_value` -> `2018-07-02 11:20:14`
` previous_value.to_datetime.next` -> `2018-07-03 11:20:14`

The date is incremented by 1, however the time remains the same. As a result when the next value is set on the new record the uniqueness validation fails since the 2 records have the same time value.

### The fix

This PR fixes the bug by adding `in(60)` to increment the timestamp by 60 second before incrementing the date with `next`:
`previous_value.to_datetime.in(60).next`

### DIFF
--- a/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
+++ b/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb
@@ -820,7 +820,7 @@ module Shoulda
           elsif previous_value.respond_to?(:next)
             previous_value.next
           elsif previous_value.respond_to?(:to_datetime)
-            previous_value.to_datetime.next
+            previous_value.to_datetime.in(60).next
           elsif boolean_value?(previous_value)
             !previous_value
           else
--- a/lib/shoulda/matchers/util.rb
+++ b/lib/shoulda/matchers/util.rb
@@ -92,7 +92,7 @@ module Shoulda
           when :datetime, :timestamp
             DateTime.new(2100, 1, 1)
           when :time
-            Time.new(2100, 1, 1)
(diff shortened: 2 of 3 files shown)


mcmire:
Sorry for the long delay on merging this. I tweaked your branch slightly to get all CI to fully pass. This is in master now.

Issue #1102, opened 2018-05-09
ActiveStorage matchers


a14m:
What is the consensus about adding macro matchers for the following?
- `has_one_attached` (from ActiveStorage Rails 5.1)
- `has_many_attached` (from ActiveStorage Rails 5.1)

mcmire:
I can't speak for the rest of the team, but I personally haven't used ActiveStorage yet. But if you're finding that you're using it a lot and you've worked out some matchers locally for them, and you'd be willing to be a sort of "champion" for this down the road, then we'd totally welcome the new matchers into this gem.

mad-raz:
Great, it's a new project and we opted in for using `ActiveStorage`, I'll try to work on the matcher later and keep this issue up to date with the progress (and reference to the commits)...
Still not quite familiar with ActiveStorage and it's moving parts, so it'll be good to have input here from other people who are interested.

mcmire:
@guialbuk What do you think if we closed this for now? I'm not sure how this would work, and considering there's still a bunch of bugfixes in the backlog to merge in first, I'm not sure if we'd be able to get to this any time soon.

schadenfred:
Without weighing in on whether or not adding this macro is a good idea, you can still use the has_many matcher to test ActiveStorage, given the following active record class in app/models/document.rb: 

`
class Document < ApplicationRecord
    has_many_attached :documents_versions
end
`
Using Minitest::Spec in test/models/document_test.rb you should still be able to do:

`
describe Document do 

  specify "has_many" do

    must have_many(:documents_versions_attachments)
  end
end
`

If you'd like to write a custom macro, open up a rails console and use this for inspiration:

`Document.reflect_on_all_associations(:has_many).map(&:name)`

Good luck!

abhchand:
Hey all - 

I just opened a PR with activestorage to formally add validations to the library. See https://github.com/rails/rails/pull/35390

The validations cover `content_type` and `size`. Validating `presence` is a bit difficult (explained in the PR) but I hope to add it at some point. 

I love the shoulda-matchers library (huge thanks to everyone who has helped build it). I think it would be great to add matchers in here. 

If I'm lucky enough to get that merged I'll consider adding matchers here if time permits 🤞 In the meantime any support on that PR (code review, a thumbs up, etc...) would help nudge it forward too. 

Thanks!

mcmire:
[...]

PR #1194, opened 2019-03-22
Started on ActiveStorage matchers

StefSchenkelaars:
I am working on issue #1102 to add matchers for `have_one_attached` and `have_many_attached`. I would like to receive some of your opinions about the structuring of the tests. Since the methods define some ActiveRecord associations, a scope, a callback and two methods. I've tried to cover all of the created methods however I don't want to start testing ActiveStorage itself.

### DIFF
--- a/lib/shoulda/matchers/active_record.rb
+++ b/lib/shoulda/matchers/active_record.rb
@@ -23,6 +23,7 @@ require "shoulda/matchers/active_record/accept_nested_attributes_for_matcher"
 require "shoulda/matchers/active_record/define_enum_for_matcher"
 require "shoulda/matchers/active_record/uniqueness"
 require "shoulda/matchers/active_record/validate_uniqueness_of_matcher"
+require "shoulda/matchers/active_record/have_attached_matcher"
 
 module Shoulda
   module Matchers
--- a/lib/shoulda/matchers/active_record/have_attached_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_attached_matcher.rb
@@ -1,0 +1,147 @@
+module Shoulda
+  module Matchers
+    module ActiveRecord
+      def have_one_attached(name)
+        HaveAttachedMatcher.new(:one, name)
+      end
+
+      def have_many_attached(name)
+        HaveAttachedMatcher.new(:many, name)
+      end
+
+      # @private
+      class HaveAttachedMatcher
+        attr_reader :name
+
+        def initialize(macro, name)
+          @macro = macro
+          @name = name
+        end
+
+        def description
+          "have a has_#{macro}_attached called #{name}"
+        end
+
+        def failure_message
+          <<-MESSAGE
+Expected #{expectation}, but this could not be proved.
+  #{@failure}
+          MESSAGE
+        end
+
+        def failure_message_when_negated
+          <<-MESSAGE
+Did not expect #{expectation}, but it does.
+          MESSAGE
+        end
+
+        def expectation
+          "#{model_class.name} to #{description}"
+        end
+
+        def matches?(subject)
+          @subject = subject
+          reader_attribute_exists? &&
+            writer_attribute_exists? &&
+            attachments_association_exists? &&
+            blobs_association_exists? &&
+            eager_loading_scope_exists?
+        end
+
+        private
+
+        attr_reader :subject, :macro
+
+        def reader_attribute_exists?
+          if subject.respond_to?(name)
+            true
+          else
+            @failure = "#{model_class.name} does not have a :#{name} method."
+            false
+          end
+        end
+
+        def writer_attribute_exists?
+          if subject.respond_to?("#{name}=")
+            true
+          else
+            @failure = "#{model_class.name} does not have a :#{name}= method."
+            false
+          end
+        end
+
+        def attachments_association_exists?
+          if attachments_association_matcher.matches?(subject)
+            true
+          else
+            @failure = attachments_association_matcher.failure_message
+            false
+          end
+        end
+
+        def attachments_association_matcher
+          @_attachments_association_matcher ||=
+            AssociationMatcher.new(
+              :"has_#{macro}",
+              attachments_association_name,
+            ).
+              conditions(name: name).
+              class_name('ActiveStorage::Attachment').
+              inverse_of(:record)
+        end
+
+        def attachments_association_name
+          case macro
+          when :one then
+            "#{name}_attachment"
+          when :many then
+            "#{name}_attachments"
+          end
+        end
+
+        def blobs_association_exists?
+          if blobs_association_matcher.matches?(subject)
+            true
+          else
+            @failure = blobs_association_matcher.failure_message
+            false
+          end
+        end
+
+        def blobs_association_matcher
+          @_blobs_association_matcher ||=
+            AssociationMatcher.new(
+              :"has_#{macro}",
+              blobs_association_name,
+            ).
+              through(attachments_association_name).
+              class_name('ActiveStorage::Blob').
+              source(:blob)
+        end
+
+        def blobs_association_name
+          case macro
+          when :one then
+            "#{name}_blob"
+          when :many then
+            "#{name}_blobs"
+          end
+        end
+
+        def eager_loading_scope_exists?
+          if model_class.respond_to?("with_attached_#{name}")
+            true
+          else
+            @failure = "#{model_class.name} does not have a " \
+                       ":with_attached_#{name} scope."
+            false
+          end
+        end
+
+        def model_class
+          subject.class
+        end
+      end
+    end
+  end
+end
--- a/spec/support/unit/helpers/rails_versions.rb
+++ b/spec/support/unit/helpers/rails_versions.rb
@@ -18,5 +18,9 @@ module UnitTests
     def rails_5_x?
       rails_version =~ '~> 5.0'
     end
+
+    def rails_gte_5_2?
+      rails_version >= 5.2
+    end
   end
 end
--- a/spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
@@ -1,0 +1,272 @@
+require 'unit_spec_helper'
+
+describe Shoulda::Matchers::ActiveRecord::HaveAttachedMatcher, type: :model do
+  if active_record_supports_active_storage?
+    before do
+      create_table :active_storage_blobs do |t|
+        t.string :key, null: false
+        t.string :filename, null: false
+        t.string :content_type
+        t.text :metadata
+        t.bigint :byte_size, null: false
+        t.string :checksum, null: false
+        t.datetime :created_at, null: false
+
+        t.index [:key], unique: true
+      end
+
+      create_table :active_storage_attachments do |t|
+        t.string :name, null: false
+        t.references :record, null: false, polymorphic: true, index: false
+        t.references :blob, null: false
+
+        t.datetime :created_at, null: false
+
+        t.index [:record_type, :record_id, :name, :blob_id],
+          name: 'index_active_storage_attachments_uniqueness', unique: true
+
+        # The original rails migration has a foreign key.
+        # Since this messes up the clearing of the database, it's removed here.
+        # t.foreign_key :active_storage_blobs, column: :blob_id
+      end
+    end
+
+    describe 'have_one_attached' do
+      describe '#description' do
+        it 'returns the message with the name of the association' do
+          matcher = have_one_attached(:avatar)
+          expect(matcher.description).
+            to eq('have a has_one_attached called avatar')
+        end
+      end
+
+      context 'when the attached exists on the model' do
+        it 'matches' do
+          record = record_having_one_attached(:avatar)
+          expect { have_one_attached(:avatar) }.
+            to match_against(record).
+            or_fail_with(<<-MESSAGE)
+Did not expect User to have a has_one_attached called avatar, but it does.
+          MESSAGE
+        end
+
+        context 'and the reader attribute does not exist' do
+          it 'matches' do
+            record = record_having_one_attached(:avatar, remove_reader: true)
+            expect { have_one_attached(:avatar) }.
+              not_to match_against(record).
+              and_fail_with(<<-MESSAGE)
+Expected User to have a has_one_attached called avatar, but this could not be proved.
+  User does not have a :avatar method.
+            MESSAGE
+          end
+        end
+
+        context 'and the writer attribute does not exist' do
+          it 'matches' do
+            record = record_having_one_attached(:avatar, remove_writer: true)
+            expect { have_one_attached(:avatar) }.
+              not_to match_against(record).
+              and_fail_with(<<-MESSAGE)
+Expected User to have a has_one_attached called avatar, but this could not be proved.
+  User does not have a :avatar= method.
+            MESSAGE
+          end
+        end
+
+        context 'and the attachments association does not exist' do
+          it 'matches' do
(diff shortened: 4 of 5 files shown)


lib/shoulda/matchers/active_record/have_attached_matcher.rb
```diff
@@ -0,0 +1,133 @@
...
+      # @private
+      class HaveAttachedMatcher
+        attr_reader :name
+
+        def initialize(macro, name)
+          @macro = macro
+          @name = name
+        end
+
+         def description
```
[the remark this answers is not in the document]
mcmire: Yup, this looks to be intended too much.

lib/shoulda/matchers/active_record/have_attached_matcher.rb
```diff
@@ -0,0 +1,133 @@
...
+          "has_#{macro}_attached #{name}"
+        end
+
+        def failure_message
+          "Expected #{expectation} (#{@failure})"
```
mcmire: This is good, but in a future update I'm going to be updating the messaging for the association matchers so they align with other matchers. The work I did in 47f27f01810e14bbf43705f67abdcfa7b2e9b08c I think serves as a good example. What do you think about something like this:

* If the reader attribute does not exist:
  > Expected Gallery to have many attached :photos, but this could not be proved.
  >   Gallery does not have a :photos method.
* If the writer attribute does not exist:
  > Expected Gallery to have many attached :photos, but this could not be proved.
  >   Gallery does not have a :photos= method.
* If either attachments or blobs association does not exist, you can re-use the failure message from the association matcher.
* If the eager loading scope doesn't exist:
  > Expected Gallery to have many attached :photos, but this could not be proved.
  >   Gallery does not have a .with_attached_photos method.

lib/shoulda/matchers/active_record/have_attached_matcher.rb
```diff
@@ -0,0 +1,133 @@
...
+              class_name('ActiveStorage::Attachment').
+              inverse_of(:record).
+              dependent(false)
+        end
+
+        def blobs_association_exists?
+          if blobs_association_matcher.matches?(subject)
+            true
+          else
+            @failure = "no valid association called #{blobs_association_name}"
```
mcmire: As mentioned above, what do you think about reusing the message that the matcher generates here? Either `blobs_association_matcher.failure_message` or `blobs_association_matcher.failure_message_when_negated`, depending on how _this_ matcher is called.

lib/shoulda/matchers/active_record/have_attached_matcher.rb
```diff
@@ -0,0 +1,133 @@
...
+
+        private
+
+        attr_reader :subject, :macro
+
+        def attachments_association_exists?
+          if attachments_association_matcher.matches?(subject)
+            true
+          else
+            @failure = "no valid association called #{attachments_association_name}"
```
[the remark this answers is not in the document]
mcmire: As mentioned above, what do you think about reusing the message that the matcher generates here? Either `attachments_association_matcher.failure_message` or `attachments_association_matcher.failure_message_when_negated`, depending on how _this_ matcher is called.

lib/shoulda/matchers/active_record/have_attached_matcher.rb
```diff
@@ -0,0 +1,133 @@
...
+            attachments_association_exists? &&
+            blobs_association_exists? &&
+            eager_loading_scope_exists?
+        end
+
+        private
+
+        attr_reader :subject, :macro
+
+        def attachments_association_exists?
```
mcmire: Can you reorganize these methods so that they appear in the same order that they're called in `matches?` That makes it easier to follow.

spec/support/unit/helpers/rails_versions.rb
```diff
@@ -38,5 +38,9 @@ def rails_lt_5?
     def rails_5_x?
       rails_version =~ '~> 5.0'
     end
+
+    def rails_gte_5_2?
```
mcmire: Instead of using a plain version conditional, I've started adding methods to [ActiveRecordVersions](https://github.com/thoughtbot/shoulda-matchers/blob/c47e6db7391963d27cb9b45c37d9aa94e53f0590/spec/support/unit/helpers/active_record_versions.rb) that describe the behavior the stuff inside the conditional is for. What do you think about adding a method there and then using it in your spec file?

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,116 @@
+require 'unit_spec_helper'
+
+# rubocop:disable Metrics/BlockLength
```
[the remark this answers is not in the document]
mcmire: I disabled this globally so you should be good here.

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        t.references :record,   null: false, polymorphic: true, index: false
+        t.references :blob,     null: false
+
+        t.datetime :created_at, null: false
+
+        t.index [ :record_type, :record_id, :name, :blob_id ], name: "index_active_storage_attachments_uniqueness", unique: true
+        t.foreign_key :active_storage_blobs, column: :blob_id
+      end
+
+      create_table(:users) do |t|
```
mcmire: When you use `define_model`, it will create a table for the model for you, so you don't usually need to do this.

StefSchenkelaars:
Hi @mcmire, 

Thanks for the review, but before I start working on it let's get some things straight. Especially the negative versions are quite interesting:

> * Negative versions of these matchers. This is something we don't have currently for association matchers, but it's logic for what happens when you say `should_not have_many_attached :whatevers` and it essentially consists of adding a `does_not_match?` method as a companion to `matches?`. It actually get can a bit tricky because you have to think through all the failure states of this too.

When do you expect the spec to fail? Let's say you have the `should_not have_many_attached :photos`  but in the model you did define a method `def photos`. I personally think that the test should just pass and only if all the conditions are true (so it has `photos`, `photos=`, etc), the spec should fail. What about you?

lib/shoulda/matchers/active_record.rb
```diff
@@ -21,6 +21,7 @@
 require "shoulda/matchers/active_record/define_enum_for_matcher"
 require "shoulda/matchers/active_record/uniqueness"
 require "shoulda/matchers/active_record/validate_uniqueness_of_matcher"
+require "shoulda/matchers/active_record/have_attached_matcher"
```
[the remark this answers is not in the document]
StefSchenkelaars: I leave this up to another PR to fix this for all strings in this file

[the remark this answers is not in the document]
mcmire: Yup, good call 👍

StefSchenkelaars:
Hi @mcmire, I started on refactoring the specs towards the `match_against` matcher. However with the current implementation of this matcher, I can't make the difference between a false positive or a false negative. Let me give an example:

This is the case when you do want to check if the model has an attached avatar (`should have_one_attached :avatar`). However it fails because the blobs relations is not set correctly. 
```ruby
context 'and the blobs association is invalid' do
  it 'matches' do
    record = record_has_one_attached(:avatar, invalidate_blobs: true)

    expect { have_one_attached(:avatar) }.
      not_to match_against(record).
        and_fail_with <<-MESSAGE
Expected User to have a has_one attached called avatar, but this could not be proved.
  Expected User to have a has_one association called avatar_blob (avatar_blob should resolve to ActiveStorage::Blob for class_name)
    MESSAGE
  end
end
```


This is the case where you want to check `should_not have_one_attached :avatar`:
```ruby
context 'when the attached does not exsist on the model' do
  it 'matches' do
    model = define_model('User') {}
    record = model.new

    expect { have_one_attached(:avatar) }.
      to_not match_against(record).
        and_fail_with <<-MESSAGE
Did not expect User to have a has_one attached called avatar, but it does.
    MESSAGE
  end
end
```

Both of these cases use the `to_not match_against` and thus use the `does_not_match?` of my matcher. This way the matcher cannot determine what failure message to show. So how do you see this? Change the `match_against` matcher to allow `to match_agains(record).and_fail_with`? This will now raise an error but it is basically what the first case should be. It matches against a record but it fails because it's false. You agree?

lib/shoulda/matchers/active_record/have_attached_matcher.rb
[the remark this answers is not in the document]
StefSchenkelaars: Now just used the `failure_message`. Negated is still an open point.

lib/shoulda/matchers/active_record/have_attached_matcher.rb
StefSchenkelaars: Done. I've put the sub-methods (so the private methods needed for a check for example `attachments_association_name`) directly below the associated methods. Let me know if you want to move them down so the methods of `matches?` follow each other directly.

spec/support/unit/helpers/rails_versions.rb
StefSchenkelaars: Done

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
StefSchenkelaars: Done

gmmcal:
I was planning on starting this code. Looking forward for using this.

trcarden:
@StefSchenkelaars / @mcmire is this still in progress or did it stall out? Looks like its ready for review and would be a nice addition.

StefSchenkelaars:
@trcarden Well, it’s far from ready. Even if I get it to completely work, it will definitely change the api. So it’s a bit stale now since I am also not 100% sure how to continue exactly. Basically it’s a complete rebuild. But I will extend the PR and get a functional branch! So we can at least keep moving forward!

mcmire:
Yup, thanks for reminding. I'm not immediately sure what's being asked above so I'll take another look at this.

StefSchenkelaars:
@mcmire I noticed that the issue was closed and that this PR has been stale for some time. I just rebased on master so the merge conflicts are out and IMO it's still in good enough shape to be merged. If there is anything you want to be changed, let me know, then I can have a look again 👍

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+          expect(matcher.description).
+            to eq('have a has_one_attached called avatar')
+        end
+      end
+
+      context 'when the attached exists on the model' do
+        let(:record) do
+          record_has_one_attached(:avatar)
+        end
+        it 'matches' do
```
mcmire, line 44: Just a style thing, but do you mind adding an empty line around before and after `it do ... end`? Same goes for `context do ... end`.

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+      reflections.delete("#{attached_name}_attachment")
+    end
+
+    if invalidate_blobs
+      reflections["#{attached_name}_blob"].options[:class_name] = 'User'
+    end
+
+    if remove_eager_loading_scope
+      instance_eval <<-CODE, __FILE__, __LINE__ + 1
+undef with_attached_#{attached_name}
```
mcmire, line 240: Does it not work to use `undef_method` here?

mcmire:
@StefSchenkelaars By the way, I just read back through your comments around the negative matchers. I think what you have here is probably fine, so I'm not too concerned about it. Negative versions of matchers are rarely used and I'm sure if someone discovers an issue then they will raise it and we can tweak this further. But from a big picture, this is good as a first version.

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+      describe '#description' do
+        it 'returns the message with the name of the association' do
+          matcher = have_one_attached(:avatar)
+          expect(matcher.description).
+            to eq('have a has_one_attached called avatar')
+        end
+      end
+
+      context 'when the attached exists on the model' do
+        let(:record) do
```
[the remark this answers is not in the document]
StefSchenkelaars: Will do 👍

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+      instance_eval <<-CODE, __FILE__, __LINE__ + 1
+undef with_attached_#{attached_name}
+      CODE
+    end
+  end
+
+  model.new
+end
+
+def record_has_many_attached(
```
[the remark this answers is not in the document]
StefSchenkelaars: Yea sounds better indeed 👍

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
StefSchenkelaars: Sure. Maybe an idea to add some rubocop checks for this kind of stuff?

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
StefSchenkelaars: Well in some way this was not working. The `with_attached_xxx` is a rails scope scope defined by the `has_
one_attached` method. In some way it's not actually defining a class method that could easily be removed. If I try the same with a method that is explicitly defined, it does work. So this is still the best I could come up with for now.

I again tried it now and it still doesn't work 😞

StefSchenkelaars:
@mcmire I've updated the code. I also noticed that the test suite was failing this time since of a foreign key constraint. I saw the structure of the creating and deleting of the tables and I thought the easiest thing to do is to just remove the foreign key since it doesn't affect the matcher nor the tests.

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
mcmire: Yeah, I had Hound enabled at one point but it doesn't seem to be working well anymore. I'll look to see if there is a new GitHub Action that someone has made that does the same thing.

spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
mcmire: Hmm... okay! No problem

StefSchenkelaars:
@mcmire Doesn't matter too much to me but it's always nice to have the correct author on the commit. So unless CI now suddenly fails, you are free to merge it 👍

mcmire:
Great! Thank you!

ruvaleev:
Execuse me, I see, that PR already merged, but my shoulda-matchers still trhrows error about `have_one_attached` matcher (
```
    Failure/Error: it { should have_one_attached :file }
       expected #<Attachment:0x00007fd21ea569f8> to respond to `has_one_attached?`
```
). I've try to remove version specification from Gemfile (change `gem 'shoulda-matchers', '~> 4.0'` to `gem 'shoulda-matchers'`), and even run `gem update 'shoulda-matchers'`, but nothing has changed. This feature is already merged to production? And how can I use it? Or we should to wait new release or something?

P.S.: Thanks!

gmmcal:
@ruvaleev this feature is still unreleased. You can see it on [CHANGELOG.md](CHANGELOG.md). You can either point your dependency to master and have this feature or wait for it to be officially released.

ruvaleev:
@gmmcal Ok, thanks )

Issue #1196, opened 2019-03-24
Document how to match a multi-column index with have_db_index


connorshea:
e.g. I have a migration like this:

```ruby
class AddUniqueIndexToGamePlatforms < ActiveRecord::Migration[5.2]
  def change
    add_index :game_platforms, [:game_id, :platform_id], unique: true
  end
end
```

And I have a matcher like this:

```ruby
it { should have_db_index([:game_id, :platform_id]).unique }
```

Currently, the documentation doesn't mention the ability to pass an array to the `have_db_index` matcher. It'd be nice if it did :)

https://github.com/thoughtbot/shoulda-matchers/blob/3635df70546f58152f1dc278f4385007f9903914/lib/shoulda/matchers/active_record/have_db_index_matcher.rb#L4-L26

rodrigovirgilio:
I created the PR, would that be it? https://github.com/thoughtbot/shoulda-matchers/pull/1197

connorshea:
This is resolved now, thanks! :)

PR #1197, opened 2019-03-26
Added documentation to accept an array of columns for have_db_index

rodrigovirgilio:
Added documentation to issue: https://github.com/thoughtbot/shoulda-matchers/issues/1196

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
@@ -24,7 +24,8 @@ module Shoulda
       #       should have_db_index(:user_id)
       #     end
       #
-      # has an index and an array of columns.
+      # It can also be used to test that the table that backs your model
+      # has an index on an array of specific columns.
       #
       #     class CreateBlogs < ActiveRecord::Migration
       #       def change
--- a/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
@@ -24,6 +24,29 @@ module Shoulda
       #       should have_db_index(:user_id)
       #     end
       #
+      # has an index and an array of columns.
+      #
+      #     class CreateBlogs < ActiveRecord::Migration
+      #       def change
+      #         create_table :blogs do |t|
+      #           t.integer :user_id
+      #           t.string :name
+      #         end
+      #
+      #         add_index :blogs, :user_id, :name
+      #       end
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Blog, type: :model do
+      #       it { should have_db_index([:user_id, :name]) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class BlogTest < ActiveSupport::TestCase
+      #       should have_db_index([:user_id, :name])
+      #     end
+      #
       # #### Qualifiers
       #
       # ##### unique


lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -65,6 +65,56 @@ module ActiveRecord
...
+      #     RSpec.describe Blog, type: :model do
+      #       it { should have_db_index([:user_id, :name]) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class BlogTest < ActiveSupport::TestCase
+      #       should have_db_index([:user_id, :name])
+      #     end
+      #
+      # #### Qualifiers
```
mcmire: Looks like you have some copypasta here. Can you remove this section? ✂️

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -65,6 +65,56 @@ module ActiveRecord
...
+      # has an index and an array of columns
+      #
+      #     class CreateBlogs < ActiveRecord::Migration
+      #       def change
+      #         create_table :blogs do |t|
+      #           t.integer :user_id
+      #           t.string :name
+      #         end
+      #
+      #         add_index :blogs, :user_id, :name
```
connorshea: Is this actually valid? I thought the syntax had to use an array, like this:

```ruby
add_index :blogs, [:user_id, :name]
```

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -65,6 +65,56 @@ module ActiveRecord
       #
       # @return [HaveDbIndexMatcher]
       #
+      # has an index and an array of columns
```
mcmire: Instead of putting this at the end, I would put this right after the first section, before `#### Qualifiers`.

Additionally, it'd be nice to flesh out the first line here so it introduces _this_ section a little more naturally.

connorshea: This was copied from above, needs to be changed to a wording that makes sense in this context.

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
connorshea: Let's remove the additions from this line and lower, it's probably unnecessary to include the `unique` qualifier again, plus it's just a repetitive section.

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
connorshea: This section should probably also be moved to be above the first `Qualifiers` header. It doesn't really make sense to have this under the `Qualifiers` section, and I think the return statement (`@return [HaveDbIndexMatcher]`) is always supposed to be at the bottom of the documentation anyway.

connorshea:
Seems I've pretty much repeated what @mcmire suggested, that's what I get for not typing fast enough :P I'll defer to their comments.

mcmire:
@connorshea Haha sorry I only realized you were leaving comments after I submitted my review :)

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
rodrigovirgilio: Not necessarily, but is valid.

rodrigovirgilio:
I made the adjustments, please check.

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -24,6 +24,29 @@ module ActiveRecord
       #       should have_db_index(:user_id)
       #     end
       #
+      # has an index and an array of columns.
```
connorshea: This should be a full sentence, like this:
```suggestion
      # It can also be used to test that the table that backs your model
      # has an index on an array of specific columns.
```

mcmire:
Hey thanks! I've merged this in 9a47fa3add95991e46eb59651b9832a166604b0f.

PR #1215, opened 2019-06-08
Add a Usage section to the README

### LINKED ISSUE #1029: Validations blow up when setting subject to described class
I typically write my specs by setting the `subject` to the class I'm writing specs for, like so:
```
Rspec.describe User, type: :model do
  subject { described_class }

  it { is_expected.to validate_presence_of(:username) }
end
```

It's just a personal style preference for my specs, however, when I do this and try to use the shoulda_matchers I get an error because shoulda_matchers is trying to set the attributes on `Class`. Like so: 
```
 NoMethodError:
       undefined method `username=' for #<Class:0x007ff38eef70c0>
```

I'm not sure if this is expected behavior or if it is something that should work. Either way I think it would be good to have it work or have it documented somewhere possibly. Let me know if I can help. Thanks.

mcmire:
Over time we've gotten questions around how matchers are intended to be
used, specifically around `subject`. This commit adds a proper Usage
section, mentioning `subject`, and also moves most of the Getting
Started section into Usage where it arguably really belongs.

---

Fixes #1029.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,11 @@ longer, more complex, and error-prone.
 
 * [Getting started](#getting-started)
    * [RSpec](#rspec)
-      * [Availability of matchers in various example groups](#availability-of-matchers-in-various-example-groups)
-      * [<code>should</code> vs <code>is_expected.to</code>](#should-vs-is_expectedto)
    * [Minitest](#minitest)
+* [Usage](#usage)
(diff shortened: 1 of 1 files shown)


mcmire:
@guialbuk What do you think about this?

guialbuk:
@mcmire I really like it. Common questions now have clear and easy to find answers 👍 🎉

Issue #962, opened 2016-09-17
Validating uniqueness on UUID fields


anthonator:
It looks like when validating uniqueness you just upcase the value to produce a new value that. That doesn't work with UUID fields since they are case insensitive. If you use `case_insensitive` it seems to work. This may be intended behavior but I found it confusing and might be worth some documentation.

``` ruby
it { should validate_uniqueness_of(:account_id).scoped_to(:user_id) }
```

produces

```
AccountUser did not properly validate that :account_id is
       case-sensitively unique within the scope of :user_id.
         After taking the given AccountUser, whose :account_id is
         ‹"eff94896-7d1d-11e6-b971-ffc3a4e8feb7"›, and saving it as the
         existing record, then making a new AccountUser and setting its
         :account_id to a different value,
         ‹"EFF94896-7D1D-11E6-B971-FFC3A4E8FEB7"› and its :user_id to a
         different value, ‹"effa716c-7d1d-11e6-b971-8bb7d7da04c6"›, the matcher
         expected the new AccountUser to be valid, but it was invalid instead,
         producing these validation errors:

         * account_id: ["has already been taken"]
```

mcmire:
Hey folks. In an effort to lighten our load as maintainers and be able to serve you better in the future, the shoulda-matchers team is working on cleaning out the cobwebs in this repo by pruning the backlog. As there are few of us, there are a lot of items that will simply never earn our attention in a reasonable time frame, and rather than giving you an empty promise, we think it makes more sense to focus on more recent issues. That means, unfortunately, that we must close this issue.

Don't take this the wrong way: our aim is not to diminish the effort people have made or dismiss problems that have been raised. If you feel that we should reopen this issue, then please let us know so that we can reprioritize it. Thanks!

PR #1221, opened 2019-06-12
WIP: Fix problem with UUIDs being case-insensitive

booch:
This is a fix for issue #962.


booch:
Thanks for the feedback, @mcmire. I've got a simple acceptance test in my actual Rails project, but I've had a hard time figuring out how to write the test with the testing framework in place. If I continue to struggle with that, I'll probably try to coordinate some time to pair up with someone to get it knocked out.

PR #1224, opened 2019-06-18
validate_presence_of(:active_storage_field) throws exception

### LINKED ISSUE #1223: validate_presence_of(:active_storage_field) throws exception
Gem version: `4.1.0`
Ruby version: `2.6.3`
Rails version: `6.0.0.rc1`
Rspec version: `3.8.1`

Given a model

```
class Test < ApplicationRecord
  has_one_attached :file
end
```

and a test

```
RSpec.describe Test, type: :model do
  it { is_expected.to validate_presence_of(:file) }
end
```

the test fails with this exception

```
1) Test should validate that :file cannot be empty/falsy
    Failure/Error: it { is_expected.to validate_presence_of(:file) }
    
    ActiveSupport::MessageVerifier::InvalidSignature:
        ActiveSupport::MessageVerifier::InvalidSignature
    # <gems_path>/gems/activestorage-6.0.0.rc1/app/models/active_storage/blob.rb:49:in `find_signed'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/changes/create_one.rb:63:in `find_or_build_blob'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/changes/create_one.rb:19:in `blob'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/changes/create_one.rb:42:in `find_attachment'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/changes/create_one.rb:38:in `find_or_build_attachment'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/changes/create_one.rb:15:in `attachment'
    # <gems_path>/gems/activestorage-6.0.0.rc1/lib/active_storage/attached/one.rb:13:in `attachment'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb:140:in `attribute_changed_value?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb:166:in `raise_attribute_changed_value_error?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb:88:in `set'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setter.rb:67:in `set!'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setters_and_validators.rb:38:in `matches?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setters_and_validators.rb:24:in `each'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setters_and_validators.rb:24:in `detect'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher/attribute_setters_and_validators.rb:24:in `first_passing'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher.rb:534:in `public_send'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher.rb:534:in `run'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/allow_value_matcher.rb:401:in `does_not_match?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/disallow_value_matcher.rb:32:in `matches?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validation_matcher.rb:162:in `run_allow_or_disallow_matcher'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validation_matcher.rb:98:in `disallows_value_of'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb:233:in `disallows_original_or_typecast_value?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb:156:in `block in matches?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb:155:in `all?'
    # <gems_path>/gems/shoulda-matchers-4.1.0/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb:155:in `matches?'
    # ./spec/models/test_spec.rb:4:in `block (2 levels) in <top (required)>'
```

I've created an example repository https://github.com/gmmcal/shoulda-error and the error can be seen in https://travis-ci.com/gmmcal/shoulda-error/builds/115617758

This bug was introduced on `4.1.0`. On `4.0.1` I cannot see this issue happening.

gmmcal:
When trying to test validation of presence against a active storage relation, shoulda tries to set the attribute as empty string. This makes the association fail to be set and breaks the test.

I've introduced a block that checks if the field is an active storage relation and only allows `nil` to be set to the object.

Fixes #1223

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -236,6 +236,8 @@ validation for you? Instead of using `validate_presence_of`, try
         def disallowed_values
           if collection?
             [Array.new]
+          elsif attachment?
+            [nil]
           else
             values = []
 
@@ -341,12 +343,22 @@ validation for you? Instead of using `validate_presence_of`, try
             model.reflect_on_association(@attribute)
         end
 
+        def attachment?
+          model.respond_to?(:reflect_on_association) &&
+            model_reflect_on_associations?(["#{@attribute}_attachment", "#{@attribute}_attachments"])
+        end
+
         def attribute_type
           if model.respond_to?(:attribute_types)
             model.attribute_types[@attribute.to_s]
           else
             LegacyAttributeType.new(model, @attribute)
           end
+        end
+
+        def model_reflect_on_associations?(associations)
+          reflections = associations.map { |association| model.reflect_on_association(association) }
+          !reflections.compact.empty?
         end
 
         def presence_validation_exists_on_attribute?
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -345,7 +345,7 @@ validation for you? Instead of using `validate_presence_of`, try
 
         def attachment?
           model.respond_to?(:reflect_on_association) &&
-            model_reflect_on_associations?(["#{@attribute}_attachment", "#{@attribute}_attachments"])
+            model_has_associations?(["#{@attribute}_attachment", "#{@attribute}_attachments"])
         end
 
         def attribute_type
@@ -356,9 +356,8 @@ validation for you? Instead of using `validate_presence_of`, try
           end
         end
 
-        def model_reflect_on_associations?(associations)
-          reflections = associations.map { |association| model.reflect_on_association(association) }
-          !reflections.compact.empty?
+        def model_has_associations?(associations)
+          associations.any? { |association| !!model.reflect_on_association(association) }
         end
 
         def presence_validation_exists_on_attribute?
--- a/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
@@ -204,6 +204,89 @@ could not be proved.
     end
   end
 
+  # validate_presence_of an attachment only works on rails >= 6
+  if rails_gte_6_0?
+    context 'a has_one_attached association with a presence validation' do
+      it 'requires the attribute to be set' do
+        expect(has_one_attached_child(presence: true)).to validate_presence_of(:child)
+      end
+
+      it_supports(
+        'ignoring_interference_by_writer',
+        tests: {
+          accept_if_qualified_but_changing_value_does_not_interfere: {
+            changing_values_with: :nil_to_blank
+          },
+          reject_if_qualified_but_changing_value_interferes: {
+            model_name: 'Example',
+            attribute_name: :attr,
+            changing_values_with: :never_falsy,
+            expected_message: <<-MESSAGE
+Expected Example to validate that :attr cannot be empty/falsy, but this
+could not be proved.
+  After setting :attr to ‹nil› -- which was read back as ‹"dummy value"›
+  -- the matcher expected the Example to be invalid, but it was valid
+  instead.
+
+  As indicated in the message above, :attr seems to be changing certain
+  values as they are set, and this could have something to do with why
+  this test is failing. If you've overridden the writer method for this
+  attribute, then you may need to change it to make this test pass, or
+  do something else entirely.
+            MESSAGE
+          }
+        }
+      )
+    end
+
+    context 'a has_one_attached association without a presence validation' do
+      it 'requires the attribute to be set' do
+        expect(has_one_attached_child(presence: false)).
+          not_to validate_presence_of(:child)
+      end
+    end
+
+    context 'a has_many_attached association with a presence validation' do
+      it 'requires the attribute to be set' do
+        expect(has_many_attached_children(presence: true)).to validate_presence_of(:children)
+      end
+
+      it_supports(
+        'ignoring_interference_by_writer',
+        tests: {
+          accept_if_qualified_but_changing_value_does_not_interfere: {
+            changing_values_with: :nil_to_blank
+          },
+          reject_if_qualified_but_changing_value_interferes: {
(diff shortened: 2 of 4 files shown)


lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -236,6 +236,8 @@ def disallows_original_or_typecast_value?(value)
         def disallowed_values
           if collection?
             [Array.new]
+          elsif attachment?
```
mcmire: Instead of adding a new clause here, what if you were to add another condition to `attribute_accepts_string_values`?

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
```diff
@@ -349,6 +356,11 @@ def attribute_type
           end
         end
 
+        def model_reflect_on_associations?(associations)
```
mcmire: Maybe this ought to be called `model_has_associations?` ?

And instead of getting all given associations and testing that it's not empty after compacting, why not say:

``` ruby
associations.any? { |association| !!model.reflect_on_association(association) }
```

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
```diff
@@ -204,6 +204,89 @@ def model_creator
     end
   end
 
+  # validate_presence_of an attachment only works on rails >= 6
```
mcmire: Instead of adding another version qualifier to RailsVersions (which really should go away), what if you were to add a condition to [ActiveRecordVersions](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/helpers/active_record_versions.rb)? That should allow you to be more expressive here (and you wouldn't need the comment anymore).

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
gmmcal: you mean a `&& !attachment?` there? I agree that a `if` condition here is bad, but `attribute_accepts_string_values` is also quite complex already. IMO, I'd just move complexity and not make readable. I'm also worried about future side-effects... maybe we can do early returns instead of `if..else` statements, like:

```
return [Array.new] if collection?
return [nil] if attachment?

(...)
```

I think it would make the method more readable.

gmmcal:
@mcmire I'll wait on your reply to item 1 and push the fixes :)

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
gmmcal: I'm creating a condition on `ActiveStorageVersions` class. then the helper is `active_storage_supports_validate_presence?`, which, IMO, makes more readable

lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
mcmire: Hmm, I'm not a fan of early returns. I think it looks pretty, but at the same time it makes the logic more difficult to untangle and reason about later. With `if`-`else`s I find I can mentally draw a flowchart in my head a lot more quickly, even if it looks "uglier".

This is probably fine for now, until we have other requirements.

spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb
mcmire: Ah, I'd forgotten that ActiveStorage was added in Rails 5.2. In that case, is it really the case that the presence validation didn't work on `has_one_attached` or `has_many_attached` before Rails 6?

gmmcal: Yeah. That's how I noticed the issue in the first place. I had `validate_presence_of` commented since I migrated to ActiveStorage. After upgrading to Rails 6, I uncommented and validation started to work. Then I uncommented the test and it was passing. After pushing to latest version of shoulda gem, I noticed this issue. You can see that it fails with rails 5 here https://travis-ci.com/gmmcal/shoulda-error/jobs/208187338

I never tracked the solution on master Rails repo, but somehow it was fixed.

gmmcal: I just noticed that I have to roll this back to `ActiveRecordVersions` because tests fails on rails < 5.2 due to non-existing `ActiveStorage::VERSION::STRING`. I'll stick to your suggestion and not create `ActiveStorageVersions` class

PR #1226, opened 2019-06-23
Fix installation of Heroku gem when using Homebrew

booch:
I doubt that any of the dependency installations were working on Linux, either, as the setup code was trying to install things with names like `apt=libpq-dev` instead of `libpq-dev`.


mcmire:
Hey, thanks so much for fixing this!

PR #1227, opened 2019-06-23
Support for folks using chruby

booch:
This is a bit tricky to support, since `chruby` is actually a function, not a command. And the script doesn't interact so well with `set -u`. (Thus the `PREFIX=''` and the `chruby ''`.)

Also, chruby doesn't come with its own Ruby installer. Most people will use `ruby-install` though.


mcmire:
Thank you!

PR #1237, opened 2019-08-20
Add `optional(false)` support to `belong_to`

### LINKED ISSUE #1219: `belong_to` does not support `optional(false)`
For a while I thought that `required` was the main option on `belong_to`, but it's `optional`. Therefore we should support passing `false` to `optional`. Right now if you have

``` ruby
belongs_to :account, optional: false
```

you have to test it with:

``` ruby
it { should belong_to(:account).required }
```

and that seems weird.

guialbuk:
Allows `belongs_to` associations with `optional: false`

```ruby
belongs_to :copany, optional: false
```

to be tested with

```ruby
it { should belong_to(:copany).optional(false) }
```

It also adds support for `optional(true)` which is the same as just `optional`.

Fixes #1219

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -1096,12 +1096,12 @@ module Shoulda
           self
         end
 
-        def optional
+        def optional(optional = true)
           remove_submatcher(AssociationMatchers::RequiredMatcher)
           add_submatcher(
             AssociationMatchers::OptionalMatcher,
             name,
-            true,
+            optional,
           )
           self
         end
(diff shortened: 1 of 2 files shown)


mcmire:
Oh nice! I just did a quick pass of this and it looks good but I will take a closer look soon.

bbugh:
Hurray! 🎉  Our app has `belongs_to_required_by_default = false` and only sets `optional: false` when required, which means there's no way for us currently to test this with `shoulda-matchers`.  We need to explicitly test `optional(false)`. Thanks for doing this PR!

XavierMeunier:
Thanks for your PR
I'm also waiting for this feature to be released, is there any planned date for the release of 4.2 ?

This will also be usefull for conditional optional in Rails 5: 
`belongs_to :category, optional: -> { draft? }`

PR #1238, opened 2019-08-20
When an unrelated error is seen with negated allow_value, give a hint

### LINKED ISSUE #919: negative allow_value matcher incorrectly fails when validation message is custom but not passed to matcher
If I have a validation like this in a model Customer:
`validate -> { errors.add(:base, 'cannot modify') }`

and then I test it with this:
`is_expected.not_to allow_value(1).for(:some_attribute)`

then I get this error:

```
       Failure/Error: is_expected.not_to allow_value(1).for(:some_attribute)

         After setting :some_attribute to ‹1›, the matcher expected the
         Customer to be invalid, but it was valid instead.
```

However, the record is not valid, it's just that `.with_message('cannot modify', against: :base)` should have been passed to the matcher.

antifun:
In cases when an unrelated error is seen that makes `allow_value` fail, the validation message was using the erroneous text "but it was valid instead". This changes the validation message to indicate which attribute(s) actually failed validation.

Fixes #919 (kinda sorta).

### DIFF
--- a/spec/support/unit/record_with_unrelated_error_builder.rb
+++ b/spec/support/unit/record_with_unrelated_error_builder.rb
@@ -1,0 +1,90 @@
+require_relative 'helpers/model_builder'
+
+module UnitTests
+  class RecordWithUnrelatedErrorBuilder
+    include ModelBuilder
+
+    def initialize(options)
+      @options = options.reverse_merge(default_options)
+    end
+
+    def attribute_that_receives_error
+      options[:attribute_that_receives_error]
+    end
+
+    def attribute_to_validate
+      options[:attribute_to_validate]
+    end
+
+    def message
+      options[:message]
+    end
+
+    def message=(message)
+      options[:message] = message
+    end
+
+    def model
+      @_model ||= create_model
+    end
+
+    def model_name
+      'Example'
+    end
+
+    def record
+      model.new
+    end
+
+    def valid_value
+      'some value'
+    end
+
+    protected
+
+    attr_reader :options
+
+    private
+
+    def context
+      {
+        validation_method_name: validation_method_name,
+        valid_value: valid_value,
+        attribute_to_validate: attribute_to_validate,
+        attribute_that_receives_error: attribute_that_receives_error,
+        message: message
+      }
+    end
+
+    def create_model
+      _context = context
+
+      define_model model_name, model_columns do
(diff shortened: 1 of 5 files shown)


spec/support/unit/record_with_unrelated_error_builder.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    def create_model
+      _context = context
+
+      define_model model_name, model_columns do
+        validate _context[:validation_method_name]
+
+        define_method(_context[:validation_method_name]) do
+          self.errors.add(_context[:attribute_that_receives_error], _context[:message])
```
antifun: This is similar to RecordWithDifferentErrorAttributeBuilder, but the behavior I want is for there to be an error that has nothing to do with the attribute to validate. If there's a better way to shoehorn that in the other class *or* to represent this condition in a builder, please let me know.

mcmire: I think this is fine!

spec/unit/shoulda/matchers/active_model/allow_value_matcher_spec.rb
```diff
@@ -347,6 +347,27 @@
...
+          builder = builder_for_record_with_unrelated_error
+          assertion = lambda do
+            expect(builder.record).
+              not_to allow_value(builder.valid_value).
+              for(builder.attribute_to_validate)
+          end
+
+          message = <<-MESSAGE
+After setting :attribute_to_validate to ‹"some value"›, the matcher
+expected the Example to be invalid, but it had errors involving other
```
mcmire, line 362: This is good, although I'm wondering whether we might need to go into a bit more detail here to avoid confusion. For instance:

```
After setting :attribute_to_validate to ‹"some value"›, the matcher expected the Example to be invalid, placing a validation error on :attribute_to_validate. The Example was invalid, but it had errors involving other attributes:

* attribute_that_receives_error: ["some message"]
```

user512:
@mcmire, hi, I came across with this issue related to `allow_value` and looks like this PR will fix it and ready to go.
Is there anything I can help with?

mcmire:
Revisiting this PR. I'm good with the direction of this PR but the only thing I would want changed is the wording because otherwise I feel like it could mislead the developer on how this matcher actually works under the hood.

PR #1241, opened 2019-08-30
Fix typo in documentation of allow_values

pacop:
Example in allow_values was wrong. It provides a valid value while using
should_not allow_values, so the test was failling but it was claiming in
documentation that the test will pass


mcmire:
Good catch, thanks!

PR #1248, opened 2019-10-20
Update documentation for Minitest

JapArt:
Update instructions from https://thoughtbot.com/blog/shoulda-matchers-3-0

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,23 @@ group :test do
 end
 ```
 
-Now you're ready to [use matchers in your tests](#usage)!
+You no longer need to say `require: false` in your Gemfile.
+You’ll need to add the following somewhere inside Test::Unit in your test_helper:
+
+```ruby
+Shoulda::Matchers.configure do |config|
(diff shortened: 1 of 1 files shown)


mcmire:
Usually people that use Minitest include `shoulda` in their Gemfile (in fact it's mentioned a paragraph above your additions). Shoulda (and `shoulda-context`) don't support shoulda-matchers 3.x yet. So I hesitate to include these instructions here because I don't want to confuse people. 

The real fix for this would be to get Shoulda up to date with `shoulda-matchers`, and then we can use the same set of instructions for both RSpec and Minitest. I'm working on that. :)

JapArt:
Yes I read the documentation but I couldn't get it to work until I read the instructions in the blog. No problem I agree with you, wait until the update in `Shoulda`. Cheers!

mcmire:
I've just released Shoulda 4.0.0 and updated the README to match the RSpec instructions, so this is no longer needed. :)

PR #1253, opened 2019-11-19
Fix appraisal command in CONTRIBUTING.md file

VSPPedro:
Hello, I was following the CONTRIBUTING.md file.

At some point it suggest to run this command:

```bash
bundle exec appraisal 5.2 rspec spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```

But instead of running the tests, it returned:

```bash
bundler: command not found: 5.2
```

So I tried to see the available list of appraisals by running:

```bash
bundle exec appraisal list
```

And this is what I got:

```bash
rails_4_2
rails_5_0
rails_5_1
rails_5_2
rails_6_0
```

Then I replaced the `5.2` with `rails_5_2` and the command worked.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,7 +142,7 @@ is created and loaded every time you start running tests.
 To run a unit test, you might say something like:
 
 ```bash
-bundle exec appraisal 5.2 rspec spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
+bundle exec appraisal rails_5_2 rspec spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
 ```
 
 ### Acceptance tests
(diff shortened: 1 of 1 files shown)


VSPPedro:
Should I run the CI again? I only updated the CONTRIBUTING.md file, I think this should not affect the tests.

The problem:
```
1) Shoulda::Matchers::ActiveRecord::ValidateUniquenessOfMatcher when the model has a scoped uniqueness validation when one of the scoped attributes is an integer column when more than one record exists that has the next version of the attribute's value accepts

     Failure/Error: expect(record).to validate_uniqueness.scoped_to(:scope)

     

       Expected Example to validate that :attr is case-sensitively unique

       within the scope of :scope, but this could not be proved.

         Given an existing Example, after setting its :attr to ‹"dummy value"›,

         then making a new Example and setting its :attr to ‹"dummy value"› as

         well and its :scope to a different value, ‹nil›, the matcher expected

         the new Example to be invalid, but it was valid instead.

     Shared Example Group: "it supports scoped attributes of a certain type" called from ./spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb:761

     # ./spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb:64:in `block (4 levels) in <top (required)>'
```

guialbuk:
I ran just the failing build again. Everything is fine now.
Thank you for your contribution 🎉

Issue #1252, opened 2019-11-10
Add `have_rich_text`


VSPPedro:
Hello, first of all, thank you all for maintaining and improving this gem.

So, I was wondering if there it's a possibility to add something like this:

```
it { is_expected.to have_rich_text(:body) }
```

To test ActionText with RSpec. Example:

```
class Post < ApplicationRecord
  has_rich_text :body
end
```

I would love to work on this if by chance this suggestion is accepted.

Thank you for your time.

EDIT: For now I'm using a custom matcher:

```
RSpec::Matchers.define :have_rich_text do |expected|
  match do |actual|
    actual.send(expected).class == ActionText::RichText
  end
end
```

guialbuk:
Hi @VSPPedro 

Your contribution would be appreciated. Feel free to submit a PR, or even a draft one to receive feedback during your progress.

And feel free to ask any questions 🙂

VSPPedro:
Hi, @guialbuk

Can I ask you a question, please? 

So I started to study this project and now I'm wondering if I should create a file named `have_rich_text.rb` or if I should  add this logic inside the `association_matcher.rb`?

VSPPedro:
I'll close this issue because the PR with the solution was merged! :tada:

O330oei:
🤪😎🤪

Sent Samsung S8 Droid
________________________________

PR #1263, opened 2019-11-23
Add have rich text matcher

VSPPedro:
Hi, this PR solves the issue #1252. 

It's a draft for now, I would love to know your opinion about it.

One of the things that needs to be improved in this PR is the creation of the `action_text_rich_texts` table. I'm out of ideas for now, any suggestions?


mcmire:
@VSPPedro How does `action_text_rich_texts` usually get generated in a Rails app? Is it using a generator, or does it get generated automatically on the fly? (Sorry, I've never used ActionText so I don't really know how it works.)

VSPPedro:
Hi, @mcmire, to make the action text to work in a Rails app it's necessary to run: `rails action_text:install`. This command it will install the yarn package necessary and add a migration with the `action_text_rich_texts` table:
```
# This migration comes from action_text (originally 20180528164100)
class CreateActionTextTables < ActiveRecord::Migration[6.0]
  def change
    create_table :action_text_rich_texts do |t|
      t.string     :name, null: false
      t.text       :body, size: :long
      t.references :record, null: false, polymorphic: true, index: false

      t.timestamps

      t.index [ :record_type, :record_id, :name ], name: "index_action_text_rich_texts_uniqueness", unique: true
    end
  end
end
```

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,77 @@
...
+            ":#{rich_text_attribute} is a #{@subject.send(rich_text_attribute).class}"
+          end
+        end
+
+        def has_attribute?
+          @subject.respond_to?(rich_text_attribute.to_s)
+        end
+
+        def has_expected_action_text?
+          @subject.send(rich_text_attribute).class == ActionText::RichText
```
guialbuk: This can raise `uninitialized constant` if `ActionText::RichText` is not available for the Rails version or available but not loaded.
A solution would be to compare `class.name` with a `'ActionText::RichText'` string

spec/support/unit/helpers/active_record_versions.rb
```diff
@@ -50,5 +50,9 @@ def active_record_supports_expression_indexes?
     def active_record_supports_validate_presence_on_active_storage?
       active_record_version >= '6.0.0.beta1'
     end
+
+    def active_record_supports_has_rich_text?
+      active_record_version >= '6.0.0.beta1'
```
guialbuk: It's possible to have Rails 6 and not load Action Text:

https://github.com/rails/rails/blob/34d66ca6bab109540299a7b30980627bea37e323/railties/lib/rails/all.rb#L16

So to check if Action Text is available, you can check for `ActionText::RichText` or `ActionText::Engine`.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'unit_spec_helper'
+
+describe Shoulda::Matchers::ActiveRecord::HaveRichText, type: :model do
+
+  if active_record_supports_has_rich_text?
```
guialbuk: It would be nice to have tests for falsy `active_record_supports_has_rich_text?`.

mcmire:
@VSPPedro So instead of adding a migration in the test itself, what I would do instead is to actually run the generator you mentioned. To explain how you would do this, before you run unit tests, a full Rails application is actually generated in the background. You can see that happening here: https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/load_environment.rb and here is the main method that generates the app: https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/rails_application.rb#L74.

Given this, if you add a new method to UnitTests::RailsApplication#generate that does something like this:

``` ruby
if rails_version >= 6 && bundle.includes?('actiontext')
  fs.within_project do
    run_command! `rails action_text:install`
  end
end
```

and then call it in #generate, then that may solve the issue for you. Migrations get run in the #load method after the #create method is called, so you shouldn't have to take care of that step. 

The benefit of this approach is that if Rails for some reason changes ActionText to use a different table, we shouldn't have to update shoulda-matchers in the future to accommodate this — everything should just continue to work. 

Let me know if you have any more questions on this.

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
VSPPedro: True!

I'll update it! Thank you!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Sure! Before this branch is merged I'll added that!

spec/support/unit/helpers/active_record_versions.rb
VSPPedro: You're right!

I'll add this validation!

VSPPedro:
First of all, thanks for all the knowledge you two are sharing with me, I am learning a lot.

@mcmire , I tried to use this:

```
if rails_version >= 6 && bundle.includes?('actiontext')
  fs.within_project do
    run_command! `rails action_text:install`
  end
end
```

 inside the #generate, but without success. These errors were popping out:

```ruby
RAILS_ENV=development environment is not defined in config/webpacker.yml, falling back to production environment
  rake aborted!
  ActiveRecord::AdapterNotSpecified: The `development` database is not configured for the `development` environment.
```

 Only worked when I tried to call it inside the `load` method right after the `load_environment` and with this it worked like a charm.

Another problems that I faced was that the command to install action_text only works with webpacker installed. I saw that the command to create the rails project skip the webpacker installation so I look for other options and I founded the possibility to copy only the migration file with this command: `action_text:install:migrations`. With this it's not more necessary copy paste the migration file inside the tests. I'm kinda happy with this result! Thanks a lot! :+1:

Another thing, I tried to use this method, `bundle.includes?('actiontext')`, but only returned false. It returned true when I added the actiontext as gem in the gemfile, but I don't think this is a good thing to do.

I replaced with: `Gem.loaded_specs.has_key?('actiontext')`. What do you think? Or should I give another try with  `bundle.includes?('actiontext')`? Or should I add the `actiontext` to the gemfile?

mcmire:
@VSPPedro Thanks for looking into this further. I'm glad that you were able to get the migrations created! With regard to `bundle.includes?('actiontext')`, I'd forgotten that ActionText is an optional dependency, so that makes that that would fail. Replacing that with `Gem.loaded_specs.has_key?('actiontext')` is fine, although the `bundle` object can ask the question from a Bundler perspective, so I'd prefer we use that so that we don't get false positives. I think you would have to add ActionText to some kind of Gemfile to get this to work, and that's okay. There are two steps to this:

* Instead of the root Gemfile, you'd probably want to add the gem to the Appraisals file, which is sort of a mesh of Gemfiles we maintain for different Rails versions. If you add it to [this spot here](https://github.com/thoughtbot/shoulda-matchers/blob/4369abdd6ff1d2d21a5bc00960323ec1250c16cf/Appraisals#L133), you can run `appraisal install` to install it locally. This will update `gemfiles/rails_6_0.gemfile` and `gemfiles/rails_6_0.gemfile.lock` but that is perfectly okay!
* You may have to ensure that it shows up in the generated application's Gemfile as well. This may not be completely necessary, but if not, then I think it's at least helpful for consistency in case you need to debug this stuff later. You can do that by adding

    ``` ruby
    if rails_version >= 6
      bundle.add_gem 'actiontext'
    end
    ```

    after [this line](https://github.com/thoughtbot/shoulda-matchers/blob/4369abdd6ff1d2d21a5bc00960323ec1250c16cf/spec/support/unit/rails_application.rb#L197).

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
VSPPedro: Done!

spec/support/unit/helpers/active_record_versions.rb
VSPPedro: I updated to this:
```
    def active_record_supports_has_rich_text?
      active_record_version >= '6.0.0.beta1' &&
        defined?(ActionText::RichText)
    end
```
Was that what you had in mind?

guialbuk: I don't think you need to check the rails version (some people are even backporting ActionText to 5.2).
Just `defined?(ActionText::RichText)`should be enough 👍

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Done!

Please, take a look into the if conditions. The difference between the second and the third is the name of the class inside the message.  

In Rails 4 the test returns a NillClass, but in Rails 5 or greater the test return a String class.

This is what I got when I run the tests with Rails 4: 

``` 
activerecord-4.2.11.1/lib/active_record/connection_adapters/abstract_adapter.rb:84: warning: deprecated Object#=~ is called on Integer; it always returns nil
```

I'll create some `shared examples` for the duplications.

VSPPedro:
It helped a lot, @mcmire! Thanks!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    
+      it 'does not matches when lacking has_rich_text ' \
+          'attribute' do
+        create_table_posts
+    
+        invalid_record = new_post(has_rich_text: false)
+        expected_message = 'Expected Post to have :invalid_attribute as a ' \
+          'ActionText::RichText but :invalid_attribute does not exist'
+    
+        aggregate_failures do
```
mcmire: Instead of using `aggregate_failures` and having to repeat yourself twice, there's a new-ish matcher you can use called `match_against` which checks that a matcher matches when used in the negative and does not match when used in the positive. You can see that in use here: https://github.com/thoughtbot/shoulda-matchers/blob/0011a3bd9a6b07f8234954d0418a74bdeb393db7/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb#L529-L536. So in this case you would say:

``` ruby
expect { have_rich_text(:invalid_attribute) }.
  not_to match_against(invalid_record).
  and_fail_with(expected_message)
```

By the way, if you want to thoroughly test both versions of the failure message (`failure_message` and `failure_message_when_negated`), you can also do the same thing with the positive cases, so you could say:

``` ruby
expect { have_rich_text(:content) }.
  to match_against(valid_record).
  or_fail_with(expected_message)
```

and this will test that the matcher matches when used in the positive and does not match when used in the negative.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      message = /undefined method `has_rich_text' for Post/
+        
+      expect(&invalid_record).to raise_error(
+        NoMethodError,
+        message
+      )
+    end
+
+    it 'does not matches when the provided attribute it\'s not a ' \
```
mcmire: What if you used double quotes here? I feel like backslashing the apostrophe makes it less readable.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  
+      aggregate_failures do
+        expect(invalid_record).not_to have_rich_text(:content)
+        expect { expect(invalid_record).to have_rich_text(:content) }.
+          to fail_with_message(expected_message)
+      end
+    end
+  end
+
+  def create_table_posts(options = {})
```
mcmire: What do you think about using keyword arguments to shorten this a bit, e.g. `def create_table_posts(has_content_attribute: false)`? Same for `new_post`.

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+
+        def failure_message_when_negated
+          "Did not expect #{@subject.class} to have a rich text " \
+          ":#{rich_text_attribute}"
+        end
+
+        def matches?(subject)
+          @subject = subject
+          @error = run_checks
```
mcmire: I like how you've done this — that's a good pattern we should use for future matchers.

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,77 @@
...
+        def initialize(rich_text_attribute)
+          @rich_text_attribute = rich_text_attribute
+        end
+
+        def description
+          "have :#{rich_text_attribute} as a ActionText::RichText"
+        end
+
+        def failure_message
+          "Expected #{@subject.class} to #{description} but #{@error}"
```
mcmire: What do you think about making private `attr_reader`s for these instance variables? And also making the `attr_reader` for `@rich_text_attribute` private?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  if rich_text_is_defined?
+    it 'returns the message including the name of the provided attribute' do
+      matcher = have_rich_text(:content)
+      expect(matcher.description).
+        to eq('have :content as a ActionText::RichText')
+    end
+
+    context 'a has_rich_text association' do
+      it 'matches when the subject configures has_rich_text for a custom ' \
```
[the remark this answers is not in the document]
VSPPedro: No problem! Done!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Much better! Thanks!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: True. Done.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: I like it your suggestion. Is more precise then `options = {}`. Thanks!

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
VSPPedro: I don't see a problem. I'll add it now. Thanks!

VSPPedro:
Next thing to do: add [shared_examples](https://relishapp.com/rspec/rspec-core/v/2-0/docs/example-groups/shared-example-group).

My task for tomorrow.

VSPPedro:
I removed the WIP, I think is almost done. What do you think?

Tomorrow I'll look for the possibility of adding [shared_examples](https://relishapp.com/rspec/rspec-core/v/2-0/docs/example-groups/shared-example-group).

VSPPedro:
Done!

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,76 @@
...
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class PostTest < ActiveSupport::TestCase
+      #       should have_rich_text(:content)
+      #     end
+      #
+      # @return [HaveRichText]
+      #
+
```
mcmire: Can you remove this extra line? I'm not certain that YARD will realize that the above comment block goes with the method.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        and_fail_with(expected_message)
+    end
+  end
+
+  def self.rich_text_is_defined?
+    defined?(ActionText::RichText)
+  end
+
+  if rich_text_is_defined?
+    it 'returns the message including the name of the provided attribute' do
```
mcmire: Mind wrapping this test in a `describe '#description'` so it's clear that's what's being tested instead of the results of the matcher?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      matcher = have_rich_text(:content)
+      expect(matcher.description).
+        to eq('have :content as a ActionText::RichText')
+    end
+
+    context 'a has_rich_text association' do
```
mcmire: This context is a tad confusing. Usually when we make contexts we are describing a scenario in which something changes. This sounds like we are changing what the subject of the test is. Also, usually whenever you have a context, you have a countercontext. For instance, if you have "when something is the case" then you will also have "when something is not the case". Here you don't have that. So what if you were to unwrap these tests and remove the context?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      include_examples "invalid attribute", 'String'
+
+      it 'matches when the subject configures has_rich_text for a custom attribute' do
+        create_table_posts
+    
+        valid_record = new_post(has_rich_text_association: true)
+    
+        expect(valid_record).to have_rich_text(:content)
```
mcmire: Right now you have a separate test that tests `failure_message_when_negated`. However, you can also use the `match_against` matcher to do this, just as you've done on line 65. So what if you were to say:

```
expect { have_rich_text(:content) }.
  to match_against(valid_record).
  or_fail_with(expected_message)
```

only here, `expected_message` would be the `failure_message_when_negated`.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          to fail_with(expected_message)
+      end
+    end
+  elsif active_record_version >= 4 && active_record_version < 5
+    include_examples 'undefined rich text', 'NilClass'
+  else
+    include_examples 'undefined rich text', 'String'
+  end
+
+  def create_table_posts(has_content_attribute: false)
```
mcmire: Can you combine these two methods together? `define_model` calls `create_table` under the hood.

For instance:

``` ruby
def new_post(has_content_attribute: false, is_rich_text_attribute: false)
  columns = {}

  if has_content_attribute
    columns[:content] = :string
  end

  define_model 'Post', columns do
    if has_content_attribute && is_rich_text_attribute
      has_rich_text :content
    end
  end.new
end
```

Also, note that I've renamed the `has_rich_text_association`. We need to be careful when using the word "association". The `has_rich_text` macro does _not_ make `content` an association, so I feel like that could be confusing if we use that word.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'unit_spec_helper'
+
+describe Shoulda::Matchers::ActiveRecord::HaveRichText, type: :model do
+  shared_examples 'undefined rich text' do |kind|
+    it 'raise a undefined method error when action text is not defined' do
+      create_table_posts
+  
+      invalid_record = -> { new_post(has_rich_text_association: true) }
```
mcmire: Can we rename `has_rich_text_association` to `is_rich_text_attribute`? Also see my comment on `create_table_posts` together.

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      # @private
+      class HaveRichText
+        def initialize(rich_text_attribute)
+          @rich_text_attribute = rich_text_attribute
+        end
+
+        def description
+          "have :#{rich_text_attribute} as a ActionText::RichText"
```
mcmire: This seems a bit clunky from an English perspective. What if we were to rename this to "have configured :#{rich_text_attribute} as a ActionText::RichText attribute"? Then the whole thing reads:

> Expected Post to have configured :content as a ActionText::RichText attribute

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'unit_spec_helper'
+
+describe Shoulda::Matchers::ActiveRecord::HaveRichText, type: :model do
+  shared_examples 'undefined rich text' do |kind|
```
mcmire: I like that you're trying to keep things DRY in this file, but since each of these shared example groups only have one test in them, I feel like it actually makes things a bit harder to follow when reading over these tests. Would you mind unfolding them? It doesn't feel like a lot of duplication anyway.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    it 'raise a undefined method error when action text is not defined' do
```
mcmire: "raise" -> "raises"
"action text" -> "ActionText"

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        NoMethodError,
+        message
+      )
+    end
+
+    include_examples "invalid attribute", kind
+  end
+
+  shared_examples "invalid attribute" do |kind|
+    it "does not matches when the provided attribute it's not a RichText class" do
```
mcmire: "matches" -> "match"
"it's" -> "is"
"is not a RichText class" -> "is not a RichText attribute", maybe?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+      it 'matches when the subject configures has_rich_text for a custom attribute' do
+        create_table_posts
+    
+        valid_record = new_post(has_rich_text_association: true)
+    
+        expect(valid_record).to have_rich_text(:content)
+      end
+    
+      it 'does not matches when lacking has_rich_text attribute' do
```
mcmire: "matches" -> "match"

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        invalid_record = new_post
+        expected_message = 'Expected Post to have :invalid_attribute as a ' \
+          'ActionText::RichText but :invalid_attribute does not exist'
+        
+        expect { have_rich_text(:invalid_attribute) }.
+          not_to match_against(invalid_record).
+          and_fail_with(expected_message)
+      end
+
+      it 'fails with the appropriate message when negated' do
```
mcmire: This should be covered already if you use the `match_against` matcher as I've mentioned on line 55.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Done!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Thank you for the lesson!

I refactored the context of the tests!

Tell me what do you think, please.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Much better!

Thanks!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Hi, @mcmire, thank you for your support!

If i’m not wrong has_rich_text macro create a association between the model and the ActionTextRichText as a has_one. e.g.: Post has one ActionText through content.

The problem here was some silly mistakes made by me.

One was the poor description of the tests caused by my rusty English, I need to work on it.

The second was because I used content for the relationship with RichText and used the same name for a attribute. It is easy to think that they are the same...

The content attribute is used for failures tests only. It's for a scenario when the developer try to use a attribute of the model and the matcher raise a warning informing something like: this is not valid, it was expecting a ActionText, but this is a Integer/String/Etc.

I refactored the tests and add your suggestion.

What do you think?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        matcher = have_rich_text(:content)
+        expect(matcher.description).
+          to eq('have configured :content as a ActionText::RichText association')
+      end
+    end
+
+    context 'when the model does not have a RichText association' do
+      it 'does not match when provided with a model attribute' do
+        invalid_record = new_post(has_invalid_content: true)
+        expected_message = 'Expected Post to have configured :invalid_content as a ' \
```
mcmire: Okay, getting closer. Now that I see this message in its full form, I'm now wondering if it's is as helpful as it could be. For instance, `invalid_content` is defined a string column, but happens to be unset at the time this test is run. Does that information help the user realize what they need to change? I guess what I'm getting at is, would it be enough to say "Expected Post to have configured :invalid_content as a ActionText::RichText association" and not include the type at the end?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      it 'returns the message including the name of the provided association' do
+        matcher = have_rich_text(:content)
+        expect(matcher.description).
+          to eq('have configured :content as a ActionText::RichText association')
+      end
+    end
+
+    context 'when the model does not have a RichText association' do
+      it 'does not match when provided with a model attribute' do
```
mcmire: In both of these tests, we are providing a model attribute to the matcher. It seems like the thing that's different between these two tests is that in one, the attribute being tested exists on the model, whereas in the other, it does not. What if you were to have the names of these tests reflect that?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        expected_message = 'Expected Post to have configured :invalid_attribute as a ' \
+          'ActionText::RichText association but :invalid_attribute does not exist'
+
+        expect { have_rich_text(:invalid_attribute) }.
+          not_to match_against(invalid_record).
+          and_fail_with(expected_message)
+      end
+    end
+  else
+    it 'raises a undefined method error when ActionText is not defined' do
```
mcmire: Sort of a similar thing here — for _both_ of these tests, not just this one, ActionText is not defined (that's why you have them in an `else`). Can you have the names reflect the fact that in one, the attribute being tested exists, whereas in the other it does not?

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        valid_record = new_post(is_rich_text_association: true)
+
+        expected_message = 'Did not expect Post to have ActionText::RichText :content'
+
+        expect { have_rich_text(:content) }.
+          to match_against(valid_record).
+          or_fail_with(expected_message)
+      end
+
+      it 'returns the message including the name of the provided association' do
```
mcmire: Do you mind pulling this test out into its own `describe` block so that we know it has to do with `description` and not with running the matcher itself? I'm not sure it needs to be in this `if` block (it should be able to run regardless of whether `ActionText` is available or not). You could put the `describe` before the `if` statement on line 8.

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
mcmire: You're totally right about the association, I should have checked that. That's what you get for making assumptions, I guess :/ So thanks for clarifying that. I checked over your changes again!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
VSPPedro: Done!

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
mcmire: Hehe 😄 so I might have miscommunicated here. What I meant was that I felt like only _this_ particular message should be changed. The other one that says ":invalid_content does not exist", I think that one makes sense to keep. It's important for people to know that an attribute doesn't exist (we do that for other matches as well), but perhaps not so important to know what the current value of the attribute is. So I think it still makes sense for lines 42 and 63 to say "Expected Post to have configured :invalid_attribute as an ActionText::RichText association, but :invalid_attribute does not exist".

VSPPedro: No you did not.

The problem was, I didn't know how to refactor this:

```
def run_checks
   if !has_attribute?
     ":#{rich_text_attribute} does not exist"
  elsif !has_expected_action_text?
    ":#{rich_text_attribute} is a #{@subject.send(rich_text_attribute).class}"
  end
end
```

The first question that pop up on my mind was: what will be returned by the second condition now? Because it does not need the description in the string. So maybe a empty string? I don't like that.

So in the end I removed the conditional and decided to make all the code more clean, but I you're right again. It's important to know that an attribute doesn't exist. 

I'll try again.

Thanks!

VSPPedro: Done!

What do you think?

O330oei:
asap

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
mcmire: I like your solution! It's concise and yet understandable.

mcmire:
Everything LGTM now! The only thing left, it looks like, is to fix the merge conflicts. I'm not certain but it might have to do with the generated gemfile.lock file. Would you find looking into that and seeing if you can fix those? Other than that, @guialbuk you have any last thoughts on this?

VSPPedro:
Sorry, but I'm not sure what to do.

It's not showing any conflict to me.

`This branch has no conflicts with the base branch`

Regardless of that, I updated my branch with master. Did that work out?

mcmire:
@VSPPedro What if you were to try rebasing the branch with master instead of merging master? If you would like to squash you are free to do that as well.

VSPPedro:
Done!

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
```diff
@@ -0,0 +1,79 @@
...
+          @subject.send(rich_text_attribute).class.name == 'ActionText::RichText'
+        end
+
+        def error_description
+          error == :default ? description : "#{description} but #{error}"
+        end
+      end
+    end
+  end
+end
```
guialbuk: Just some nitpicking: missing am empty end lime here

spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      columns[:invalid_content] = :string
+    end
+  
+    define_model 'Post', columns do
+      if is_rich_text_association
+        has_rich_text :content
+      end
+    end.new
+  end
+end
```
guialbuk: And here

lib/shoulda/matchers/active_record/have_rich_text_matcher.rb
VSPPedro: Ops, thanks!

Done! :tada:

mcmire:
Looks like CI is green. Merging now!

PR #1282, opened 2020-03-21
Add support to numericality: :other_than

gmmcal:
Add support to validate numericality of a field when set :other_than comparison


gmmcal:
Travis build has passed, but github was not properly notified =/

guialbuk:
@gmmcal I'm restarting the build then 🚀

PR #1286, opened 2020-04-03
Fix setup script for RedHat based Linux

biinari:
`bin/setup` seems to have a partial implementation for RedHat based Linux distributions (i.e. those using RPM packages).

This completes what is there and as a side note drops the seemingly unused `heroku` requirement.

* Use `rpm=...` packages with `yum`
* Check for `yum` as a package manager
* Drop requirement on `heroku`

**Please note: I am not using a RedHat based Linux at the moment but have tested this out in a docker image for fedora:31 using rbenv to build ruby 2.6.5.**

Yum seems to be currently available in notable RPM distros:

* Fedora 22+ has adopted `dnf` in favour of `yum` but `yum` still seems to be available there.
* RHEL and CentOS 6 to 8 use `yum`

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -154,11 +154,6 @@ install-dependencies() {
     start postgresql
   fi
 
-  if ! has-executable heroku; then
-    banner 'Installing Heroku'
-    install brew=heroku/brew/heroku heroku
-  fi
-
   if has-executable rbenv; then
     if ! (rbenv versions | grep $RUBY_VERSION'\>' &>/dev/null); then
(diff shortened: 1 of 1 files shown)


mcmire:
Sorry it took me so long to get to this. The current incarnation of the setup script comes from my [`setup_script_generator`](https://github.com/mcmire/setup_script_generator) project and that I think is where the Heroku dependency snuck in. I'll backport your changes over there but this should do for now!

biinari:
Nice one @mcmire, thanks.

PR #1289, opened 2020-04-28
Read output while wating for spec command runner

### LINKED ISSUE #1287: Specs hang when trying to run locally
Running specs locally hangs with no output when run as `bundle exec rake`

Trying running a unit spec with `DEBUG_COMMANDS=1` to show where the problem lies:

```sh
$ DEBUG_COMMANDS=1 bundle exec appraisal rails_5_2 rspec spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
>> BUNDLE_GEMFILE=/home/bill/code/v2/shoulda-matchers/gemfiles/rails_5_2.gemfile bundle exec rspec spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb

Changing to directory: /home/bill/code/v2/shoulda-matchers
Running command: rails new /tmp/shoulda-matchers-acceptance/test-project --skip-bundle --no-rc --skip-webpack-install
```

It looks like the `IO.pipe` used to capture the stdout + stderr from the test command runner is filling up a buffer.

It looks like `rails new` is generating [4143 characters of output](https://gist.github.com/biinari/5e1af3b5155b91fbb7447e5abb2be0ed). Assuming the specs are running fine elsewhere, this seems to suggest the ruby pipe may have a buffer size of 4096 bytes and that rails output may just have increased a little to tip over the balance.

## Version Info

ruby 2.6.5
rbenv 1.1.2
shoulda-matchers git master at 390c2ca161b35bce7171c65a6ec3b653b7c2b877

## Potential fix

I have been experimenting with variations of `Open3.capture2e` (which I think won't play well with timeouts) and `Open3.popen2e`. Or alternatively can just stick with `spawn` and create a thread to wait for the process to complete while consume output on the main thread (or vice versa).

biinari:
Fixes #1287 - an issue that appears to involve the child process blocking on writing output and the parent waiting for the child to finish.

In spec helpers command runner, read from the stdout/stderr pipe in a separate thread while waiting for the spawned process to finish.

## Test code

As the affected code is a spec helper, it does not seem appropriate to add a spec for this behaviour. Here is a minimal example to showcase the fix:

```ruby
require './spec/support/tests/command_runner'

n = 65_536
runner = Tests::CommandRunner.new('ruby', '-e', %('print(("a" * #{n}) + ".")'))
runner.run_successfully = true
runner.run_quickly = true
runner.call
puts "actual length: #{runner.output.length}, expected: #{n+1}"
```

### DIFF
--- a/spec/support/tests/command_runner.rb
+++ b/spec/support/tests/command_runner.rb
@@ -25,6 +25,7 @@ module Tests
 
     def initialize(*args)
       @reader, @writer = IO.pipe
+      @command_output = ''
       options = (args.last.is_a?(Hash) ? args.pop : {})
       @args = args
       @options = options.merge(
@@ -84,7 +85,7 @@ module Tests
     def output
       @_output ||= begin
         stop
-        without_colors(reader.read)
+        without_colors(command_output)
       end
     end
 
@@ -126,7 +127,7 @@ Output:
 
     protected
 
-    attr_reader :args, :reader, :writer, :wrapper
+    attr_reader :args, :command_output, :reader, :writer, :wrapper
 
     private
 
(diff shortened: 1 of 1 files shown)


mcmire:
Do you want to squash your commits before I merge or leave them as is? Not sure if you have a preference or not.

biinari:
Thanks @mcmire. I've squashed the commits up as the second was a fixup to the first. Ready for you to merge if you're happy.

I've certainly learned more about the interaction of IO and forked processes through this and a similar fix in another gem.

mcmire:
Great! Merged!

PR #1290, opened 2020-05-06
Truncate long values in descriptions.  Fixes #1041

### LINKED ISSUE #1041: Improve performance of determine_where_to_break_line
I have the following line of code in my model:

```
validates :content, length: { maximum: 20.megabytes }
```

and the following test:

```
it { should validate_length_of(:content).is_at_most(19.megabytes) }
```

The process hangs while it's generating the failure message.  Using a debugger, I traced it to this line:
https://github.com/thoughtbot/shoulda-matchers/blob/v3.1.2/lib/shoulda/matchers/util/word_wrap.rb#L194

When dealing with very long lines, we should limit how far we go trying to find some whitespace.

What do you think of changing this line from:

```
while line[index] !~ /\s/ && (0...line.length).cover?(index)
```

to:

```
line_length_or_reasonable_limit = [line.length, REASONABLE_LIMIT].min
while line[index] !~ /\s/ && (0...line_length_or_reasonable_limit).cover?(index)
```

and what would be the `REASONABLE_LIMIT` (I'm leaning towards 500)?

betesh:
@mcmire Thanks for reminding me about #1041 by closing it!

Here's the commit I referenced there, rebased onto the latest master.

### DIFF
--- a/lib/shoulda/matchers/util.rb
+++ b/lib/shoulda/matchers/util.rb
@@ -4,6 +4,8 @@ module Shoulda
   module Matchers
     # @private
     module Util
+      MAXIMUM_LENGTH_OF_VALUE_TO_DISPLAY = 500
+
       def self.deconstantize(path)
         if defined?(ActiveSupport::Inflector) &&
           ActiveSupport::Inflector.respond_to?(:deconstantize)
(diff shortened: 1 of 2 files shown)


mcmire:
Thanks! Merged!

PR #879, opened 2016-01-11
Change in? to include?

oogali:
Use ```include?``` to check membership of an array instead of ```in?```.

Fixes #877 

All tests passing in this branch.
```
>> bundle check --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.0.0.gemfile' || bundle install --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.0.0.gemfile'
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
The Gemfile's dependencies are satisfied
>> bundle check --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.0.1.gemfile' || bundle install --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.0.1.gemfile'
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
The Gemfile's dependencies are satisfied
>> bundle check --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.1.gemfile' || bundle install --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.1.gemfile'
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
The Gemfile's dependencies are satisfied
>> bundle check --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.2.gemfile' || bundle install --gemfile='/Users/oogali/shoulda-matchers/gemfiles/4.2.gemfile'
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
The Gemfile's dependencies are satisfied
>> BUNDLE_GEMFILE=/Users/oogali/shoulda-matchers/gemfiles/4.2.gemfile bundle exec rake --trace
** Invoke default (first_time)
** Execute default
rake spec:unit --trace
** Invoke spec:unit (first_time)
** Execute spec:unit

Randomized with seed 62006
....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 1 minute 4.08 seconds (files took 7.64 seconds to load)
[...]

Issue #877, opened 2016-01-11
Hidden dependency on active_support


oogali:
**TL;DR version: Consider changing references from the active_support-specific ```in?``` to the pure Ruby form: ```include?```**

I have a project that uses Sinatra and ActiveRecord, but not Rails. For testing, I'm using RSpec, Factory Girl, and shoulda-matchers.

I'm attempting to write a test to validate the uniqueness of a column with a scope:
```
it { should validate_uniqueness_of(:priority).scoped_to(:publisher_id) }
```

but running into an exception on each run:
```
     Failure/Error: it { should validate_uniqueness_of(:priority).scoped_to(:publisher_id) }
     NoMethodError:
       undefined method `in?' for 5:Fixnum
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:533:in `boolean_value?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:528:in `block (2 levels) in all_scopes_are_booleans?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:528:in `each'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:528:in `all?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:528:in `block in all_scopes_are_booleans?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:527:in `each'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:527:in `all?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:527:in `all_scopes_are_booleans?'
     # ./gems/gems/shoulda-matchers-3.0.1/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb:441:in `validate_after_scope_change?'
[...]

PR #1291, opened 2020-05-11
Fix exception raised by ActiveSupport Object#in?

jmromer:
PR #879 refactored to replace invocations of `ActiveSupport`'s' `#in?`
method. This patch follows suit to fix the following exception, which 
occurs when `ActiveSupport` isn't a project dependency:

```
1) Stocking is expected to validate that :bin cannot be empty/falsy
    Failure/Error: it { should validate_presence_of(:bin) }

    NoMethodError:
      undefined method `in?' for :belongs_to:Symbol
      Did you mean?  nil?
```

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -327,8 +327,8 @@ validation for you? Instead of using `validate_presence_of`, try
         end
 
         def collection_association?
-          association? && association_reflection.macro.in?(
-            [:has_many, :has_and_belongs_to_many],
+          association? && [:has_many, :has_and_belongs_to_many].include?(
+            association_reflection.macro
           )
         end
 


mcmire:
This gem depends on ActiveSupport (see the [gemfile](https://github.com/thoughtbot/shoulda-matchers/blob/master/shoulda-matchers.gemspec#L37)), so I think it's reasonable to assume that it's loaded in some fashion. How are you using this gem, if I may ask?

jmromer:
> so I think it's reasonable to assume that it's loaded in some fashion

That assumption seems like a leap to me, and if nothing else it imposes an undocumented dependency (not on `ActiveSupport` per se but on using it in one's app) on any non-Rails app using it.

[Internal consistency](https://github.com/thoughtbot/shoulda-matchers/pull/879) is also desirable on its own:

```shell
shoulda-matchers master % ag -l '\.include\?' | wc -l
27

shoulda-matchers master % ag -l '\.in\?' | wc -l
1
```

> How are you using this gem, if I may ask?

Following the [README](https://github.com/thoughtbot/shoulda-matchers/blob/master/README.md)'s installation for a non-Rails app:

> Non-Rails apps
> 
> If your project isn't a Rails app, but you still make use of ActiveRecord or ActiveModel, you can still use this gem too! In that case, you'll want to place the following configuration at the bottom of `spec/spec_helper.rb`:
> 
> ```rb
> Shoulda::Matchers.configure do |config|
>   config.integrate do |with|
>     with.test_framework :rspec
> 
>     # Keep as many of these lines as are necessary:
>     with.library :active_record
>     with.library :active_model
>   end
> end
> ```

[This](https://github.com/thoughtbot/shoulda-matchers/issues/877) seems sensible:

> The thing is, of course, that just because you're requiring the active_support gem doesn't mean you're requiring all of it, so whatever file provides in? probably isn't getting required as well.

mcmire:
The part you quoted was the next thing I was going to say :) ActiveSupport is loaded whether you're explicitly using it or not (because both ActiveModel and ActiveRecord depend on it), but it's designed to be quasi-modular, so the exact file that provides `.in?` may not be required yet.

In any case, you've convinced me on the consistency argument. The only thing I'd like to see here is a test that exercises the scenario you're in so that we don't run into this problem again. We already some some high-level tests which set up an app and double-check that certain matchers are able to be used. We have a [scenario](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/acceptance/multiple_libraries_integration_spec.rb) which creates a Rails app and plugs into `:active_record` and `:active_model`, and we also have a [non-Rails app scenario](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/acceptance/active_model_integration_spec.rb) which only plugs into `:active_model`, but what we don't have is a scenario for a non-Rails app which plugs into `:active_record` and `:active_model`. So, what if you did this:

* Comment out the change you made.
* Add a new file `spec/support/acceptance/helpers/active_record_helpers.rb` that's similar to [`active_model_helpers`](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/acceptance/helpers/active_model_helpers.rb) but adds an `active_record_version` helper method.
* Require and include the new module in [this helpers file](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/acceptance/helpers/active_model_helpers.rb).
* Add a new [step helper](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/acceptance/helpers/step_helpers.rb#L15) called `create_active_record_project` that is similar to `create_active_model_project` but also adds the active_record gem in addition to active_model.
* Make a new file in `spec/acceptance` that's like `multiple_libraries_integration_spec.rb`, only instead of `create_rails_application` on [line 5](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/acceptance/multiple_libraries_integration_spec.rb#L5) you use your new `create_active_record_application` helper.
* Run the new test with: `bundle exec appraisal rails_6_0 rspec spec/acceptance/name_of_your_file_spec.rb`. Watch it fail with the same error you posted.
* Uncomment your change from earlier and re-run the test. It should now pass.

jmromer:
I'm a bit pressed for time at the moment, so unfortunately I can't do that — feel free to close if you're short on bandwidth.

VSPPedro:
I'd like to work on the tests if it's okay with you

mcmire:
@VSPPedro That would be great — go for it!

PR #1314, opened 2020-07-08
word_wrap: fix Ruby 2.7 warning

fatih:
This fixes the following warning:

```
vendor/gems/ruby/2.7.0/gems/shoulda-matchers-4.3.0/lib/shoulda/matchers/util/word_wrap.rb:8: 
  warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

### DIFF
--- a/lib/shoulda/matchers/util/word_wrap.rb
+++ b/lib/shoulda/matchers/util/word_wrap.rb
@@ -5,7 +5,7 @@ module Shoulda
       TERMINAL_WIDTH = 72
 
       def word_wrap(document, options = {})
-        Document.new(document, options).wrap
+        Document.new(document, **options).wrap
       end
     end
 


mcmire:
We're working on adding Ruby 2.7 to the test matrix (see #1310), so we'll make sure to merge this once that's merged in.

fatih:
@mcmire thanks for the heads up. Let me know if there is anything else I need to do. I have a similar PR here in another `shoulda-` repo: https://github.com/thoughtbot/shoulda-context/pull/64

mcmire:
I've merged #1310, but the same issue that held up https://github.com/thoughtbot/shoulda-context/pull/64 is also holding this one up too (the bit of code that detects for warnings and fails the build is also not working here). I am fixing this in #1328, so once that's merged then we should be clear to merge this.

mcmire:
Actually based on my findings in #1328 I can already guarantee this will fix the warning so I'll go ahead and merge this ahead of that PR :) Thank you!

fatih:
Thank you @mcmire

PR #1316, opened 2020-07-09
Reflect latest version (4.3) in README

VSPPedro:
Hi, I hope all of you who maintain this gem are doing well.

I noticed that the Readme is not reporting the latest version.

This PR aims to update this information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Start by including `shoulda-matchers` in your Gemfile:
 
 ```ruby
 group :test do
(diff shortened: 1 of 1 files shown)


mcmire:
Hi @VSPPedro! `~> 4.0` is equivalent to `>= 4.0, < 5.0`, and so you should be getting the latest 4.x version when adding that to your Gemfile. Is this not what you're seeing?

VSPPedro:
I did not know that!

My bad, thank you!

Issue #1317, opened 2020-07-11
Use autoloading instead of requiring all files up front


mcmire:
We have a lot of files which are like [active_record.rb](https://github.com/thoughtbot/shoulda-matchers/blob/f50f70e9c8679b4e3625f445478743df875616c9/lib/shoulda/matchers/active_record.rb) where there are a bunch of `require`s at the top of the file. I've never liked this because there are different files at different levels being required. This is not as organized as it could be, and if you add a new file, it's confusing where to add the require. I've started using `autoload` in other projects and I find it easier to manage. We should do the same thing here.

For instance, for `active_record.rb`, we might change this to something like:

``` ruby
module Shoulda
  module Matchers
    module ActiveRecord
      autoload :AssociationMatcher, "shoulda/matchers/active_record/association_matcher"
      autoload :AssociationMatchers, "shoulda/matchers/active_record/association_matchers"
      autoload :HaveDbColumnMatcher, "shoulda/matchers/active_record/have_db_column_matcher"
      # ...
    end
  end
end
```

VSPPedro:
Hi, @mcmire, I would love to work on this!

mcmire:
@VSPPedro Okay, that would be awesome! I will assign this to you :)

VSPPedro:
Thanks!

n-rodriguez:
@mcmire why not using [Zeitwerk](https://github.com/fxn/zeitwerk)?

mcmire:
@n-rodriguez I'm not sure that Zeitwerk would be a good choice here. It seems like it's really well suited for applications, but it seems overkill for gems IMO. It would have to be added as a runtime dependency and I don't feel like people should have to bring in Zeitwerk to use shoulda-matchers.

n-rodriguez:
> but it seems overkill for gems IMO.

IMO this PR is the perfect use case for Zeitwerk since you do exactly what it does.

VSPPedro:
@mcmire First, sorry for the problems with the PR and thank you for releasing the fix so fast.

For now, I will focus on adding the tests first, as described by you here: https://github.com/thoughtbot/shoulda-matchers/issues/1333#issuecomment-680980460. After adding these tests I will continue with this PR - if that's okay with you.

mcmire:
@VSPPedro Yeah, that's fine with me!

mcmire:
I'm going to close this issue. It seems like this is a relatively minor pain point and is actually more complicated than I thought it would be.

PR #1320, opened 2020-07-13
Use autoloading instead of requiring all files up front

VSPPedro:
Hi,

First, thanks for the opportunity to work on this issue: #1317

There still some files being loaded with require, I tried to use autoload but got some errors.

I'll keep digging for a solution.

### DIFF
--- a/lib/shoulda/matchers/active_record/uniqueness.rb
+++ b/lib/shoulda/matchers/active_record/uniqueness.rb
@@ -1,14 +1,14 @@
 module Shoulda
   module Matchers
-    module ActiveModel
+    module ActiveRecord
       # @private
       module Uniqueness
+        autoload :Model, 'shoulda/matchers/active_record/uniqueness/model'
+        autoload :Namespace, 'shoulda/matchers/active_record/uniqueness/namespace'
+        autoload :TestModelCreator, 'shoulda/matchers/active_record/uniqueness/test_model_creator'
+        autoload :TestModels, 'shoulda/matchers/active_record/uniqueness/test_models'
       end
     end
   end
 end
 
-require 'shoulda/matchers/active_record/uniqueness/model'
-require 'shoulda/matchers/active_record/uniqueness/namespace'
-require 'shoulda/matchers/active_record/uniqueness/test_model_creator'
-require 'shoulda/matchers/active_record/uniqueness/test_models'
--- a/lib/shoulda/matchers/independent/delegate_method_matcher.rb
+++ b/lib/shoulda/matchers/independent/delegate_method_matcher.rb
@@ -176,6 +176,9 @@ module Shoulda
 
       # @private
       class DelegateMethodMatcher
+        autoload :StubbedTarget, 'shoulda/matchers/independent/delegate_method_matcher/stubbed_target'
+        autoload :DelegateObjectNotSpecified, 'shoulda/matchers/independent/delegate_method_matcher/target_not_defined_error'
+
         def initialize(delegating_method)
           @delegating_method = delegating_method
 
--- a/lib/shoulda/matchers/integrations/test_frameworks.rb
+++ b/lib/shoulda/matchers/integrations/test_frameworks.rb
@@ -1,15 +1,17 @@
-require 'shoulda/matchers/integrations/test_frameworks/active_support_test_case'
 require 'shoulda/matchers/integrations/test_frameworks/minitest_4'
 require 'shoulda/matchers/integrations/test_frameworks/minitest_5'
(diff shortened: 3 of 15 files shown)


lib/shoulda/matchers/integrations/registry.rb
```diff
@@ -26,3 +26,5 @@ def find_class!(name)
     end
   end
 end
+
+require 'shoulda/matchers/integrations/libraries'
```
mcmire, line 30: Perhaps this should remain in `shoulda/matchers/integrations`? If this is here then it won't get loaded until Registry is loaded.

lib/shoulda/matchers/integrations/test_frameworks.rb
```diff
@@ -1,15 +1,15 @@
...
 require 'shoulda/matchers/integrations/test_frameworks/missing_test_framework'
 require 'shoulda/matchers/integrations/test_frameworks/rspec'
-require 'shoulda/matchers/integrations/test_frameworks/test_unit'
 
 module Shoulda
   module Matchers
     module Integrations
       # @private
       module TestFrameworks
+        autoload :ActiveSupportTestCase, 'shoulda/matchers/integrations/test_frameworks/active_support_test_case'
```
mcmire, line 9: I think this is a case where these files can't be autoloaded, because there is a bit of magic here where upon being loaded, they will register themselves as possible test frameworks. This makes it possible to say e.g.

``` ruby
Shoulda::Matchers.configure do |config|
  config.integrate do |with|
    with.test_framework :rspec   # or whatever
  end
end
```

If the appropriate file that corresponds to the symbol you're setting isn't loaded then it doesn't know what to do and the matchers will never be integrated into the appropriate test framework. Perhaps there is another way to accomplish this to where we can autoload them, but that's how they work now. This may be why the tests are failing.

lib/shoulda/matchers/independent/delegate_method_matcher.rb
```diff
@@ -176,6 +173,9 @@ def delegate_method(delegating_method)
 
       # @private
       class DelegateMethodMatcher
+        autoload :StubbedTarget, 'shoulda/matchers/independent/delegate_method_matcher/stubbed_target'
+        autoload :DelegateObjectNotSpecified, 'shoulda/matchers/independent/delegate_method_matcher/target_not_defined_error'
```
mcmire, line 177: Never noticed that this constant isn't named the same as the file, whoops! 😅

lib/shoulda/matchers/active_record/uniqueness.rb
```diff
@@ -1,14 +1,14 @@
 module Shoulda
   module Matchers
-    module ActiveModel
+    module ActiveRecord
```
mcmire, line 3: Oops, thanks :)

VSPPedro: No problem!

lib/shoulda/matchers/independent/delegate_method_matcher.rb
VSPPedro: Should I rename the class or the file?

lib/shoulda/matchers/integrations/test_frameworks.rb
VSPPedro: I made a silly mistake. I was only running unit tests.

I think the best strategy is to create a new branch of the master and make the changes little by little to see what went wrong.

This time I'll run all the tests at every change! :sweat_smile: 

Thanks for the info!

lib/shoulda/matchers/integrations/registry.rb
VSPPedro: No problem! I'll do that!

lib/shoulda/matchers/integrations/test_frameworks.rb
VSPPedro: Done! Tests fixed!

lib/shoulda/matchers/independent/delegate_method_matcher.rb
mcmire: I think this is fine as-is — some future update can fix this.

mcmire:
This looks good to me... does it look good to you? If so feel free to rebase/squash whenever you want to and I'll merge it in.

VSPPedro:
Hi!

It does look good to me!

I was only waiting for a feedback on the constant that does not have the same name as the file.

Conflict solved and rebase/squash done!

mcmire:
Awesome, thank you for your work on this!

Issue #1332, opened 2020-08-26
v4.4.0 NoMethodError: undefined method `belong_to'


baburdick:
With the release of v4.4.0, I'm seeing test failures with errors resembling this:
```
v4.4.0 NoMethodError: undefined method `belong_to' for #<RSpec::ExampleGroups::MyClass::Associations:0x00007fd5bdfc5518>
```
Reverting to the prior release fixes the issue.

jarkko:
Ditto. Seems like pretty much all the association matchers broke for us.

```
  1) SaxoBlockTradeOrder is expected to have many :trade_orders
     Failure/Error: it { should have_many(:trade_orders) }
       expected #<SaxoBlockTradeOrder:0x00007f7eeb8b9ab8> to respond to `has_many?`
     # ./spec/models/saxo_block_trade_order_spec.rb:6:in `block (2 levels) in <main>'
     # -e:1:in `<main>'

  2) SaxoBlockTradeOrder 
     Failure/Error: it { should belong_to(:saxo_allocation_key) }
     
     NoMethodError:
       undefined method `belong_to' for #<RSpec::ExampleGroups::SaxoBlockTradeOrder:0x00007f7ee1154cc8>
     # ./spec/models/saxo_block_trade_order_spec.rb:5:in `block (2 levels) in <main>'
     # -e:1:in `<main>'
```

arthurborisow:
Faced the same issue. Seems like the problem with autoloading. When just writing `Shoulda::Matchers::ActiveRecord::AssociationMatcher` and `Shoulda::Matchers::ActiveRecord::AssociationMatcher` before the tests resolve the issue but that's ugly

denisj:
Same here. So this might be this [commit](https://github.com/thoughtbot/shoulda-matchers/commit/ce78101362d8826d0bfaf4e7935aa0643f3e4c53#diff-a629c86e49f9eabce177fbd99635ae8a)?

n-rodriguez:
> Same here. So this might be this [commit](https://github.com/thoughtbot/shoulda-matchers/commit/ce78101362d8826d0bfaf4e7935aa0643f3e4c53#diff-a629c86e49f9eabce177fbd99635ae8a)?

why not using [Zeitwerk](https://github.com/fxn/zeitwerk)?

timlapluie:
It seems to be a feature, not a bug.

The release notes of `4.0.0` contain the following:
```
[...]

Issue #1331, opened 2020-08-26
have_one with dependent is no longer working from 4.4.0


prrrnd:
Using `dependent` on a `have_one` matcher doesn't seem to work on 4.4.0, but worked on 4.3.0.

```ruby
RSpec.describe Person, type: :model do
  it { should have_one(:contract).dependent(:destroy) }
end
```

```bash
$ rspec
Randomized with seed 30332

  1) Person 
     Failure/Error: it { should have_one(:contract).dependent(:destroy) }

     NoMethodError:
       undefined method `dependent' for #<RSpec::Matchers::BuiltIn::Has:0x0000564a1e04f5d8>
```

Volosh1n:
Same with `have_many`:
```
Failure/Error: it { is_expected.to have_many(:products).dependent(:destroy) }
      
      NoMethodError:
        undefined method `dependent' for #<RSpec::Matchers::BuiltIn::Has:0x000055e2152c1710>
```

VSPPedro:
Thank you for creating this issue.

There is already a PR to solve this problem.

Soon, we will have a new release with this fix.

mcmire:
Fixed in v4.4.1!

PR #1334, opened 2020-08-26
Revert "Use autoloading instead of requiring all files up front"

### LINKED ISSUE #1333: v4.4.0  introduces undefined method errors
After upgrading to version 4.4.0 we see errors like
```
     NoMethodError: undefined method `validate_length_of' for #<RSpec::ExampleGroups::Foo
```
in many projects. It is the same in Rails and non Rails projects.

As I see some similar issues (#1332 and #1331) it might be a general problem. Maybe it has to do with https://github.com/thoughtbot/shoulda-matchers/pull/1320?

VSPPedro:
This PR reverts the commit ce78101362d8826d0bfaf4e7935aa0643f3e4c53.

Closes #1333  #1332 #1331.

### DIFF
--- a/lib/shoulda/matchers/active_record/uniqueness.rb
+++ b/lib/shoulda/matchers/active_record/uniqueness.rb
@@ -1,14 +1,14 @@
 module Shoulda
   module Matchers
-    module ActiveRecord
+    module ActiveModel
       # @private
       module Uniqueness
-        autoload :Model, 'shoulda/matchers/active_record/uniqueness/model'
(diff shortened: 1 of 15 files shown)


lib/shoulda/matchers/active_record/uniqueness.rb
```diff
@@ -1,14 +1,14 @@
 module Shoulda
   module Matchers
-    module ActiveRecord
+    module ActiveModel
       # @private
```
VSPPedro, line 4: Later I intend to create a PR to send this change and rename the file name which is different from the constant if that's ok with you.

PR #1340, opened 2020-08-27
fix: Replace http with https [ci skip]

KapilSachdev:
Though being insignificant but took liberty to update the example with it as well.


mcmire:
Great, thank you so much!

PR #1341, opened 2020-08-27
Increase rails integration tests

VSPPedro:
Adding more rails integration tests similar to those in [Shoulda](https://github.com/thoughtbot/shoulda/blob/master/test/acceptance/integrates_with_rails_test.rb).

The lists below track which tests are missing.

ActiveModel Matchers:

- [x] allow_value
- [x] have_secure_password
- [x] validate_absence_of
- [x] validate_acceptance_of
- [x] validate_confirmation_of
- [x] validate_exclusion_of
- [x] validate_inclusion_of
- [x] validate_length_of
- [x] validate_numericality_of
- [x] validate_presence_of

ActiveRecord Matchers:

- [x] accept_nested_attributes_for
- [x] belong_to
- [x] define_enum_for
- [x] have_and_belong_to_many
- [x] have_db_column
- [x] have_db_index
- [x] have_implicit_order_column (Rails 6 +)
- [x] have_many
- [x] have_many_attached (Rails 5.2 +)
- [x] have_one
- [x] have_one_attached (Rails 5.2 +)
- [x] have_readonly_attribute
- [x] have_rich_text (Rails 6 +)
- [x] serialize
- [x] validate_uniqueness_of

ActionController Matchers:

- [x] filter_param
- [x] permit
- [ ] redirect_to
- [ ] render_template
- [ ] render_with_layout
- [ ] rescue_from
- [x] respond_with
- [ ] route
- [ ] set_session
- [ ] set_flash
- [ ] use_after_action
- [ ] use_around_action
- [ ] use_before_action

### DIFF
--- a/spec/support/tests/rails_versions.rb
+++ b/spec/support/tests/rails_versions.rb
@@ -19,8 +19,8 @@ module Tests
       rails_version =~ '~> 5.0'
     end
 
-    def rails_gt_5?
-      rails_version > 5
+    def rails_6_x?
+      rails_version =~ '~> 6.0'
     end
 
     def rails_gte_5_2?
--- a/spec/support/tests/rails_versions.rb
+++ b/spec/support/tests/rails_versions.rb
@@ -1,4 +1,4 @@
-module UnitTests
+module Tests
   module RailsVersions
     extend self
 
--- a/spec/acceptance/rails_integration_spec.rb
+++ b/spec/acceptance/rails_integration_spec.rb
@@ -8,18 +8,6 @@ describe 'shoulda-matchers integrates with Rails' do
     # TODO: ActionController matchers
     # configure_routes_with_single_wildcard_route
 
-    # if rails_gt_5? && bundle.includes?('actiontext')
-    #   run_rake_tasks!('action_text:install:migrations')
-    # end
-
-    # if rails_gte_5_2? && bundle.includes?('activestorage')
-    #   run_rake_tasks!('active_storage:install:migrations')
-    # end
-
-    # TODO: Fix uninitialized constant Rails
-
-    run_rake_tasks!('action_text:install:migrations')
-    run_rake_tasks!('active_storage:install:migrations')
     run_rake_tasks!('db:drop', 'db:create', 'db:migrate')
   end
 
@@ -90,42 +78,7 @@ describe 'shoulda-matchers integrates with Rails' do
   end
 
   def run_tests_for_n_unit
-    write_file 'test/unit/user_test.rb', <<-FILE
-      require 'test_helper'
-
-      class UserTest < ActiveSupport::TestCase
-        # ActiveModel matchers
-        should allow_value('https://foo.com').for(:website_url)
-        should have_secure_password
-        should validate_absence_of(:first_name)
-        should validate_acceptance_of(:terms_of_service)
-        should validate_confirmation_of(:email)
-        should validate_exclusion_of(:age).in_array(0..17)
-        should validate_inclusion_of(:role).in_array(%w( admin manager ))
-        should validate_length_of(:password).is_at_least(10).on(:create)
-        should validate_numericality_of(:number_of_dependents).on(:create)
-        should validate_presence_of(:email)
-
-        # ActiveRecord matchers
-        should have_many(:issues)
-        should accept_nested_attributes_for(:issues)
-        should belong_to(:organization)
-        should define_enum_for(:status)
-        should have_and_belong_to_many(:categories)
-        should have_db_column(:email)
-        should have_db_index(:organization_id)
-        should have_implicit_order_column(:created_at)
-        should have_many_attached(:photos)
-        should have_one(:profile)
-        should have_one_attached(:avatar)
-        should have_readonly_attribute(:username)
-        should have_rich_text(:description)
-        should serialize(:social_networks)
-        should validate_uniqueness_of(:email)
-      end
-    FILE
-
-    # TODO: ActionController matchers tests
+    create_files_for_minitest
 
     result = run_n_unit_test_suite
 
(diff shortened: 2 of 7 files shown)


VSPPedro:
For now, the tests are only passing on Rails 6.

For Rails 5.2 it's necessary to prevent `run_rake_tasks!('action_text:install:migrations')` command to be executed and disable the matchers tests for:

- have_implicit_order_column
- have_rich_text

For Rails 5.1 or below it's necessary to prevent the same things on 5.2 plus `run_rake_tasks!('active_storage:install:migrations')` and disable these tests as well:

- have_many_attached
- have_one_attached

VSPPedro:
I wold like to suggest to move the [rails_application.rb](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/rails_application.rb), [rails_version.rb](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/helpers/rails_versions.rb) and [load_environment.rb](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/unit/load_environment.rb) to outside of unit folder and use them both for unit tests and acceptance tests.

It seems to me that these three files should be shared to prevent duplications on the configuration of the project. For example this one for the migrations of the rails project:

```ruby
if rails_gt_5? && bundle.includes?('actiontext')
  run_rake_tasks!('action_text:install:migrations')
end
```

The unit tests already have this configuration, but acceptance tests don't.

You may be wondering how I got this idea. Or maybe why.

I tried to include the `rails_version.rb` for the acceptance tests to be able to use methods like `rails_gte_5_2?` like the example above, but then I got this error: 'Rails is not defined'.

Ok. So Rails is defined on `unit_spec_helper.rb`, but not on `acceptance_spec_helper.rb`.

I don't think I should add something like a `require 'rails'`, so I tried to understand what was going on. 

The [unit_spec_helper.rb](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit_spec_helper.rb#L1) first line told me where to look:

```ruby
 `require_relative 'support/unit/load_environment'`
```

Then these three lines from `load_envirohment.rb` caught my attention:

```ruby
$test_app = UnitTests::RailsApplication.new
$test_app.create
$test_app.load
```

And for my surprise, the load method is already doing almost everything that I need:

```ruby
def load
  load_environment

  if rails_version > 5 && bundle.includes?('actiontext')
    add_action_text_migration
  end

  run_migrations
end
```

It's only missing this piece of code:

```ruby
...
if rails_version >= 5.2 && bundle.includes?('active_storage')
  add_active_storage_migration
end
...

def add_action_text_migration
  fs.within_project do
    run_command! 'bundle exec rake active_storage:install:migrations'
  end
end
```

Should I refactor this in another PR? Or should I leave as it is?

mcmire:
Hey @VSPPedro. Sounds like you're finding some good stuff. You're right that there are some differences in how the unit tests set up the Rails applications vs. how the acceptance tests do it. This happened because I added both separately and didn't realize they were doing similar things. Since then, I've created a gem called [Snowglobe](https://github.com/mcmire/snowglobe) which encapsulates the work of creating a Rails application for use in tests. This gem actually came out of both [`shoulda`](https://github.com/thoughtbot/shoulda/blob/master/test/acceptance/integrates_with_rails_test.rb) and [`shoulda-context`](https://github.com/thoughtbot/shoulda-context/blob/master/test/shoulda/rerun_snippet_test.rb), but it was heavily based on UnitTests::RailsApplication here. So, ideally, the test setup in `shoulda-matchers` could be refactored to make use of Snowglobe and that would solve the inconsistency problem you're seeing going forward. It wouldn't be a fast job*, but it would unify the two things.

I think it's fine to include changes to the unit and/or acceptance test setup in this PR if it helps you write these tests better. In terms of whether or not to do the Snowglobe refactor, if synchronizing the two manually like you're doing isn't too much trouble, maybe it's best to stick with that approach for now and we can address the refactor in a future commit.

_* Snowglobe doesn't do everything that the `shoulda-matchers` test setup does, so it would likely have to be modified so that `shoulda-matchers` can make use of it. Also, in unit tests, the Rails application is loaded once into the same Ruby process in which the tests are running before all unit tests are run, whereas in acceptance tests, the Rails application is loaded within a completely separate Ruby process, for each acceptance test run. It's possible to use Snowglobe's RailsApplication class in both cases, but that distinction just has to be kept in mind when doing the refactor._

VSPPedro:
Thank you for sharing your thoughts with me.

I am interested in doing this factoring using Snowglobe, but I admit that there is still a long way to go to achieve this and I already promised other PRs here. hahaha

So, I will follow your suggestion and leave this refactoring for later.

For this PR, I see only the need to move `rails_versions.rb` to a common folder for unit tests and acceptance tests. On the last commit I moved this file to the `tests` folder, but if you think I should leave it somewhere else, just let me know.

spec/support/tests/rails_versions.rb
```diff
@@ -1,8 +1,8 @@
-module UnitTests
+module Tests
   module RailsVersions
     extend self
 
-    def self.configure_example_group(example_group)
+    def configure_example_group(example_group)
```
mcmire, line 5: Heads up that the `self.` was intentional to communicate that `configure_example_group` is callable at the class level but shouldn't be available within tests.

mcmire:
@VSPPedro I think moving `rails_versions.rb` to `tests` makes sense — that's where I've been keeping things that are common to all kinds of tests. I made a small comment above but it looks like you're on the right track :)

spec/support/tests/rails_versions.rb
VSPPedro: I see.

To be honest with you I did not know much about `extend self` until I saw this file. After reading about it I came to the conclusion that the `self.` was not necessary. I thought it was redundant. My bad. Thanks for the heads up.

I'll revert this change.

VSPPedro:
Would you like me to divide this PR into three?

- Increase rails integration tests for ActiveModel Matchers
- Increase rails integration tests for ActiveRecord Matchers
- Increase rails integration tests for ActionController Matchers

To make revising the code easier.

I think I'll have the time to finish this in two weeks. :)

VSPPedro:
I see know a better approach to this PR. I think it will help to make it leaner. :)

Let me describe what I did first and what I intend to do know, after that, I would like to hear your thoughts on this, please. When you have the time.

I've found three test scenarios so far:

1 - The rails version is equal to 6. In this case it will test all the matchers.
2 - The rails version is equal to 5.2. The matchers added for Rails 6 will be ignored: have_implicit_order_column, have_rich_text.
3 - The rails version is below 5.2. The matchers added for Rails 5.2 and 6 will be ignored: have_implicit_order_column, have_rich_text, have_many_attached, have_one_attached.

To make this happen I create two more models: `user.rb` and `photo_album.rb` with migrations, rspec and minitest.

In case of scenario 1, all the files would be added. In case os scenario 2, it would only add user.rb and photo_album.rb. In case of scenario 3, it would only add user.rb. Of course, for each scenario, the necessary migrations and test files were also configured - rspec and minitest. I thought that this type of organization would be an easy way to separate the configuration for each scenario and make the code simpler.

Now I don't like that idea very much because I think I found a better answer:  [def comment_lines_matching](https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/tests/filesystem.rb#L81https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/support/tests/filesystem.rb#L81).

So, I intend to remove the post.rb and photo_album.rb models. Move the code that was in them to user.rb and use the method I discovered to dynamically comment lines that are not compatible with the version of Rails being tested.

What do you guys think?

mcmire:
@VSPPedro Sorry I just saw your latest comment. Sounds like you're on a good track here!

PR #1343, opened 2020-08-29
fix(docs): Migrate to rouge [ci skip]

### LINKED ISSUE #1336: Generating documentation no longer works
We use Pygments along with YARD to highlight source code in documentation. Unfortunately this seems to have stopped working and upon running `yard doc` (which happens automatically when creating new releases) we are getting:

```
MentosError: No header received back
```

This seems to have been a [problem](https://github.com/tmm1/pygments.rb/issues/45) for a very long time and there is no known solution.

Perhaps instead of Pygments we should switch to [Rouge](https://github.com/rouge-ruby/rouge) instead?

KapilSachdev:
Closes #1336

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,6 @@ gem 'zeus', require: false
 
 # YARD
 gem 'fssm'
-gem 'pygments.rb'
+gem 'rouge'
 gem 'redcarpet'
 gem 'yard'
--- a/doc_config/yard/setup.rb
+++ b/doc_config/yard/setup.rb
@@ -1,20 +1,20 @@
 YARD::Templates::Engine.register_template_path(File.dirname(__FILE__) + '/templates')
 
-require 'pygments.rb'
+require 'rouge'
 
 module YARD
   module Templates
     module Helpers
       module HtmlSyntaxHighlightHelper
         def html_syntax_highlight_ruby(source)
-          highlight_with_pygments(:ruby, source)
+          highlight(:ruby, source)
         end
 
         private
 
-        def highlight_with_pygments(language, source)
-          html = Pygments.highlight(source, lexer: language.to_s)
-          html.sub(%r{\A<div class="highlight">\s*<pre>}, '').sub(%r{</pre>\s*</div>\Z}, '')
+        def highlight(language, source)
+          lexer = Rouge::Lexers.const_get(language.capitalize)
+          Rouge::Formatters::HTML.new.format(lexer.new.lex(source))
         end
       end
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,6 @@ GEM
     diff-lcs (1.3)
     fssm (0.2.10)
     method_source (1.0.0)
-    multi_json (1.14.1)
     parallel (1.19.2)
     parser (2.7.1.4)
       ast (~> 2.4.1)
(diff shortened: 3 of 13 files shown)


doc_config/yard/setup.rb
```diff
@@ -13,8 +13,7 @@ def html_syntax_highlight_ruby(source)
         private
 
         def highlight_with_pygments(language, source)
```
KapilSachdev, line 15: Do we need to support languages other than ruby?

Gemfile
```diff
@@ -13,6 +13,6 @@ gem 'zeus', require: false
 
 # YARD
 gem 'fssm'
-gem 'pygments.rb'
+gem 'rouge'
```
KapilSachdev, line 16: I'm not sure why we do not specify versions in the Gemfile, so continuing with it.
But i think it would be good to specify for all.

doc_config/yard/setup.rb
```diff
@@ -1,20 +1,20 @@
...
+          highlight(:ruby, source)
         end
 
         private
 
-        def highlight_with_pygments(language, source)
-          html = Pygments.highlight(source, lexer: language.to_s)
-          html.sub(%r{\A<div class="highlight">\s*<pre>}, '').sub(%r{</pre>\s*</div>\Z}, '')
+        def highlight(language, source)
+          lexer = Rouge::Lexers.const_get(language.capitalize)
```
KapilSachdev, line 16: Continuing support for language.

Gemfile
mcmire: The versions are locked in `Gemfile.lock` and `gemfiles/*.gemfile.lock`, and all of those are committed to the repo, so that's what keeps things in check from a development perspective. It would only be a potential problem if someone were to run `bundle update` and update everything across the board, but that happens pretty rarely.

doc_config/yard/setup.rb
mcmire: I don't think so — I think this is fine for now.

mcmire:
Oops, looks like the PR I just merged created conflicts, would you mind rebasing this?

doc_config/yard/setup.rb
KapilSachdev: I already added the support on later commit, should i revert back to using `ruby`?

Gemfile
KapilSachdev: Yes, thats true.
Just mentioning few points for thoughts.

- But using `pessimistic locking (~>)` is helpful in many ways like when bundle update is executed in case we need some fix by a gem and we have defined the version to minor `Major.minor.patch`, it will update the gem but do not introduce breaking changes.
- Defining also helps in knowing if a gem has some incompatible version with our use case, if any.
- In case `Gemfile.lock` has conflicts then one usually removes the content and regenerate the lock file, this might also introduce unwarranted versions as lock content was removed.

_That said, we can continue with this approach as it all depends on how the gem is being maintained._

doc_config/yard/setup.rb
mcmire: Oh, right, that makes sense. That's okay, what you did is fine too. :)

KapilSachdev:
@mcmire as all documentation related issues are fixed now, can you please release the missing documentation of all the versions released after `4.3.0`.

mcmire:
@KapilSachdev Yup! Regenerated docs for `v4.4.0` and `v4.4.1`. If you went to `http://matchers.shoulda.io/docs/` recently the redirect may be cached, but it should shortly redirect you to `v4.4.1`.

VSPPedro:
What do we need to do to update the list of matchers on this page? - https://matchers.shoulda.io/

If I'm not wrong, the accept_nested_attributes_for is deprecated.
And it's missing have_rich_text, have_one_attached, have_many_attached, have_implicit_order_column and maybe others have_implicit_order_column.

mcmire:
@VSPPedro Good point! I totally forgot about that page.

The code for that page is kept in the `gh-pages` branch along with the documentation. I have a notice in that branch that says you shouldn't update that branch directly but in this case it would be fine. The "home page" is kept in [`index.html`](https://github.com/thoughtbot/shoulda-matchers/blob/gh-pages/index.html) (sorry, I'd forgotten it's just one big HTML file — there might be a more maintainable way we can organize that code).

PR #1348, opened 2020-09-02
Update README [ci skip]

VSPPedro:
The association_matcher.rb is a big file and has more than one matcher inside. This PR update the links to redirect the user to where the description of the matcher starts.

This strategy is already used on the links of these matchers: use_after_action, use_around_action and use_before_action.

I'm still working on the PR #1341 and that's why I'm finding these details. I'm learning a lot more about this gem. Thanks.

Another possible contribution, if anyone is interested, is to add an example of using have_attached_matchers within the file like the other matchers. I think it would be great to add that.


mcmire:
I agree that it is pretty large and I think this will help a lot. Thank you!

PR #1351, opened 2020-09-06
fix(delegate_method): Support Class as a subject

### LINKED ISSUE #586: delegate_method doesn't allow me to test #to another class
I have this code that works (I'm not using STI for specific reason, instead I'm simulating MTI):

``` Ruby
class AbstractMessage
  def self.some_method
    #fdfdf
  end
end

class Message
  class << self
    delegate :some_method, to: AbstractMessage
  end
end

# Message.some_method will actually call AbstractMessage.some_method
```

This piece of code works in Ruby, and it actually delegates some_method to the AbstractMessage, however shoulda is not considering this special case.

KapilSachdev:
Closes #586 

TODO
- [ ] Test Case
- [ ] Changelog entry, if needed.

### DIFF
--- a/lib/shoulda/matchers/independent/delegate_method_matcher.rb
+++ b/lib/shoulda/matchers/independent/delegate_method_matcher.rb
@@ -200,9 +200,12 @@ module Shoulda
           ensure_delegate_object_has_been_specified!
 
           subject_has_delegating_method? &&
-            subject_has_delegate_object_reader_method? &&
-            subject_delegates_to_delegate_object_correctly? &&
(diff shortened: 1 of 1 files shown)


VSPPedro:
Hello @KapilSachdev, are you still working on this PR? If not, would you mind if I worked on it?

I asking you because I noticed that the milestone of version 4.5.0 is almost complete. Only this PR is missing.

So, if you need some help please let me know.

Thanks!

KapilSachdev:
Sure, please go ahead.

Issue #1353, opened 2020-09-11
Drop `git` in gemspec


utkarsh2102:
Hiya,

Thanks for your lovely work on this! ❤️ 

However, whilst trying to maintain this in Debian, we found that this library relies on `git` to produce a list of files.
Using `git` in gemspec is a bit troublesome and it's best to not use them!

More about this can be found here: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#packaginggemspecgit

If you agree with this, I'll be very happy to raise a PR for this! 😄 

CC: @mcmire

mcmire:
Hi @utkarsh2102. For context, the gemspec was originally created using Bundler. My guess as to why `git` was added to the gemspec was to prevent uncommitted changes from being released. I just checked Bundler, however, and it seems that `rake release` already checks against this. Therefore, I can't see a reason why we would need to use `git` in the gemspec. So I'd be fine with a PR that simplifies that part of the gemspec.

utkarsh2102:
Awesome, thanks!
We're working with the Bundler team to avoid using `git` by default (bundler itself doesn't use git!). And also, bundler now won't let you release if you have uncommitted changes, so it's pretty safe that way, too!

I'll work on the PR now :)

PR #1354, opened 2020-09-11
Drop `git ls-files` in gemspec

utkarsh2102:
Hi again 👋🏻 

As discussed in #1353, this PR drops `git` in gemspec, thereby also helping in smooth downstream maintenance (in Debian, Fedora, and so on).
Also adds the `Packaging` extension of RuboCop. More about it can be found on https://docs.rubocop.org/rubocop-packaging.

Closes: #1353 

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'pry-byebug'
 gem 'rake', '13.0.1'
 gem 'rspec', '~> 3.9'
 gem 'rubocop', require: false
+gem 'rubocop-packaging', require: false
 gem 'rubocop-rails', require: false
 gem 'warnings_logger'
 gem 'zeus', require: false
--- a/tasks/documentation.rb
+++ b/tasks/documentation.rb
@@ -1,4 +1,4 @@
-require_relative '../lib/shoulda/matchers/version'
+require 'shoulda/matchers/version'
 require 'erb'
 
 module Shoulda
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,6 @@
-require: rubocop-rails
+require:
+  - rubocop-packaging
+  - rubocop-rails
 AllCops:
   NewCops: disable
   TargetRubyVersion: 2.4
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -51,6 +51,8 @@ GEM
       unicode-display_width (>= 1.4.0, < 2.0)
     rubocop-ast (0.3.0)
       parser (>= 2.7.1.4)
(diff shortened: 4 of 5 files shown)


tasks/documentation.rb
```diff
@@ -1,4 +1,4 @@
-require_relative '../lib/shoulda/matchers/version'
+require 'shoulda/matchers/version'
```
KapilSachdev, line 1: Why this change is done? Seems unrelated to the PR.

Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'pry-byebug'
 gem 'rake', '13.0.1'
 gem 'rspec', '~> 3.9'
 gem 'rubocop', require: false
+gem 'rubocop-packaging', require: false
```
KapilSachdev, line 10: Though it misses a `gemlock` entry but why this has been added?
Has there been any reported issues or problems that this addition helps in resolving ?

tasks/documentation.rb
utkarsh2102: Each change has been committed separately with a verbose commit message.
Please let me know if they're not clear enough.

Gemfile
utkarsh2102: `Gemfile.lock` has now been updated, thanks!
It helps us in maintaining downstream (links in the commit message :)). More about these cops can be found at https://docs.rubocop.org/rubocop-packaging and this has a style guide hosted at https://packaging.rubystyle.guide.

KapilSachdev: Ok, so `require_relative` was part of this, got it :+1:

tasks/documentation.rb
VSPPedro: Nice commit message. I did not know that. Thanks! :+1: 

I'm just wondering, the project has 86 uses of require_relative, so why modify just this one?

If we intend to change several require_relative to require, I think it would be better to separate into another PR, but for now, I think we need to wait for more feedback on this change.

utkarsh2102: Hehe, nice! 😄 
The rationale for each cop is documented here: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#require-relative-hardcoding-lib-rationale

So it's still preferred to use `require_relative` but just not cross-lib. More specifically, defintiely not from the test code to `lib/`.

utkarsh2102:
@mcmire, should be ready for your review now, could you please take a look? 😄

mcmire:
Thank you for these changes! Merged!

PR #1356, opened 2020-09-14
Support custom has_secure_password attributes

krismichalski:
Starting with version 6, Rails accepts custom attributes for `has_secure_password` macro.
Without one it defaults to `password`, so everything should still work without breaking for older versions of Rails as well.

Rails pull request for allowing configurable attribute name on `#has_secure_password` - https://github.com/rails/rails/pull/26764


mcmire (approved):
This looks great to me! Thanks so much!

Issue #1281, opened 2020-03-11
have_db_column.().with_options() should raise error on misspelled argument


recurator:
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]
Rails 6.0.0
RSpec 3.8
  - rspec-core 3.8.2
  - rspec-expectations 3.8.4
  - rspec-mocks 3.8.1
  - rspec-rails 3.8.2
  - rspec-support 3.8.2
shoulda-matchers (4.1.2)

spec/models/comment_spec.rb:
```
require 'rails_helper'
RSpec.describe Comment, type: :model do
  it { is_expected.to have_db_column(:created_at).of_type(:datetime).with_options(preccision: 6, null: false)
end
```

*** Note spelling error: preccision

It is expected to raise error but no error is raised.

mcmire:
Closing this since #1358 was merged.

PR #1358, opened 2020-10-02
fix have_db_column.with_options misspelled options

rodriggochaves:
Hi!! This is my first contribuition to `shoulda-matchers`. If you got any feedback for this PR, I'll be very happy to listen.

This PR looks to solve the bug reported here #1281. 

### Solution descripton 

When there is an argument this matcher doesn't expect, it should raise an `ArgumentError`. I created a new protected method `validate_options` to be responsible for this.

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
@@ -84,6 +84,8 @@ module Shoulda
 
       # @private
       class HaveDbColumnMatcher
+        OPTIONS = %i(precision limit default null scale primary)
+
         def initialize(column)
           @column = column
           @options = {}
@@ -95,7 +97,8 @@ module Shoulda
         end
 
         def with_options(opts = {})
-          %w(precision limit default null scale primary).each do |attribute|
+          validate_options(opts)
+          OPTIONS.each do |attribute|
             if opts.key?(attribute.to_sym)
               @options[attribute.to_sym] = opts[attribute.to_sym]
             end
@@ -136,6 +139,13 @@ module Shoulda
         end
 
         protected
+
+        def validate_options(opts)
+          invalid_options = opts.keys.map(&:to_sym) - OPTIONS
+          if invalid_options.any?
+            raise ArgumentError, "Unknown option(s): #{invalid_options.map(&:inspect).join(", ")}"
+          end
+        end
 
         def column_exists?
           if model_class.column_names.include?(@column.to_s)
--- a/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
@@ -98,6 +98,14 @@ describe Shoulda::Matchers::ActiveRecord::HaveDbColumnMatcher, type: :model do
     end
   end
 
+  context 'with invalid argument option' do
+    it 'raises an error with the unknown options' do
+      expect {
+        have_db_column(:salary).with_options(preccision: 5, primaryy: true)
+      }.to raise_error("Unknown option(s): :preccision, :primaryy")
+    end
+  end
+
   def model(options = {})
     define_model(:employee, options).new
   end


lib/shoulda/matchers/active_record/have_db_column_matcher.rb
```diff
@@ -137,6 +138,13 @@ def description
 
         protected
 
+        OPTIONS = %i(precision limit default null scale primary)
```
mcmire, line 141: Would you mind putting this at the top of the class? That way it "pops" out.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
```diff
@@ -98,6 +98,14 @@
     end
   end
 
+  context 'with invalid argument option' do
+    it 'throws an ArgumentError' do
+      expect {
+        have_db_column(:salary).with_options(preccision: 5)
+      }.to raise_error(ArgumentError)
```
mcmire, line 105: What do you think about testing what the error message is going to be rather than what kind of error it is? When you do that you might see that error could be improved a bit. I believe that as `invalid_options` is an array, when it gets interpolated it will appear like an array in the message. Not sure if that's what you were going after or not.

lib/shoulda/matchers/active_record/have_db_column_matcher.rb
```diff
@@ -137,6 +138,13 @@ def description
...
+
+        def validate_options(opts)
+          invalid_options = opts.keys.map(&:to_sym) - OPTIONS
+          raise ArgumentError, "Unknown option(s) '#{invalid_options}'" unless invalid_options.empty?
```
mcmire, line 145: Just a little style thing: we don't tend to use modifiers or `unless` — would you mind formatting this as:

``` ruby
if invalid_options.any?
  raise ArgumentError, "Unknown option(s) '#{invalid_options}'"
end
```

(although see note below about possibly reformatting the message)

mcmire (commented):
Hi @rodriggochaves — thank you for this! It's nice when PRs are small :) Just had a few comments below.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
rodriggochaves: Ah! Good observation!! I just changed the test and improved the message raised. Let me know what did you think about  it?

lib/shoulda/matchers/active_record/have_db_column_matcher.rb
rodriggochaves: I moved but I've a question about the visibility of this constant. Shoud I use something to hide like `#private_constant`?

mcmire: Sure, that would make sense if you want to do that.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
```diff
@@ -98,6 +98,14 @@
     end
   end
 
+  context 'with invalid argument option' do
+    it 'raises an error with the unknown options' do
+      expect {
+        have_db_column(:salary).with_options(preccision: 5, primaryy: true)
+      }.to raise_error("Unknown option(s): 'preccision, primaryy'")
```
mcmire, line 105: Okay, so now that you're testing multiple options, it seems like we could improve this message a bit. What if you change the above to 

``` ruby
raise ArgumentError, "Unknown option(s): #{invalid_options.map(&:inspect).join(", ")}"
```

Then you should get something like:

``` ruby
expect {
  have_db_column(:salary).with_options(preccision: 5, primaryy: true)
}.to raise_error("Unknown option(s): :preccision, :primaryy")
```

mcmire (commented):
One more little thing but otherwise I'm good with this!

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
rodriggochaves: Ok! I improved the message!

mcmire (approved):
Sorry for the delay on this. This looks good to me, so thank you! It sounds like we may want to push this to the version after the next immediate one, so I'm going to hold off on merging but this looks ready to go once we are ready to merge.

VSPPedro:
Thank you very much, @rodriggochaves!

@mcmire, I noticed that this PR was selected to be part of the 4.5.0. So, I think we can merge it.

mcmire:
@VSPPedro Ah, sorry! For some reason I thought we were inserting a patch release before 4.5.0. In that case I'll merge this.

PR #1364, opened 2020-10-30
Use build  instead of create in validate_uniqueness

evanlouden:
Thank you for the awesome library!

You don't actually have to persist the record to get the validates_uniqueness test to pass. Everyone who copies the current code from here is slowing down their test suite unnecessarily.


mcmire:
You're right that you don't _need_ to create the record before giving it to the matcher. In terms of performance, I think it might be the same, at least for the uniqueness matcher — either way a record has to be present in the database for the matcher to compare against, so either you give us a record you've already saved, or you give us a new record and we'll save it for you. But you're right that we shouldn't be encouraging people to use created records for validation tests in general. I'll merge this. Thank you!

PR #1365, opened 2020-11-01
fix: Run actiontext migrations for rails >= 6

KapilSachdev:
... as actiontext is made available on rails 6.0

### DIFF
--- a/spec/support/unit/rails_application.rb
+++ b/spec/support/unit/rails_application.rb
@@ -26,9 +26,7 @@ module UnitTests
     def load
       load_environment
 
-      if rails_version > 5 && bundle.includes?('actiontext')
-        add_action_text_migration
-      end
+      add_action_text_migration if bundle.includes?('actiontext')
 
       run_migrations
     end


spec/support/unit/rails_application.rb
```diff
@@ -26,7 +26,7 @@ def create
     def load
       load_environment
 
-      if rails_version > 5 && bundle.includes?('actiontext')
+      if rails_version >= 6 && bundle.includes?('actiontext')
         add_action_text_migration
```
VSPPedro, line 30: But it can be used in versions below if included in the Gemfile that is why the other condition. `(bundle.includes?('actiontext'))`

For example: https://www.driftingruby.com/episodes/using-action-text-in-a-rails-5-2-application

Regardless, I see no problem with this adjustment because the actiontext is only tested on Rails 6.

Why not remove even the second condition?

And I think you can remove this line too: https://github.com/thoughtbot/shoulda-matchers/blob/master/gemfiles/rails_6_0.gemfile#L36

KapilSachdev: >I notice now that I forgot to test the actiontext on other versions. It's only tested on Rails 6. If we decide to not test on 5.2, I see no problem with this adjustment.

Unless we use a patched fork to work around, `rails/actiontext` gem dependencies for `rails < 6` can not be met from official sources.

We can also use `bundle.includes?('actiontext')` only rather than checking the version.

VSPPedro: Yeah, you are right. We can ignore the version. I don't see any problem with this change, but I think we can do even better. Let me take another look at the project.

KapilSachdev: Yes, maybe later it becomes a refactoring candidate as we do not rely on version but rather on definition of `actiontext`.

mcmire:
Is this build failing, is that what caused this change? I guess I'm wondering what this affects and how you knew to change this.

KapilSachdev:
Yes, I noticed it because of the build failure while fixing `Style/ClassEqualityComparison` offence.

Though I think I will have to add `defined?(ActionText::RichText)` at
https://github.com/thoughtbot/shoulda-matchers/blob/402abf81bb925e1c0e8807d4d29e70861261e9d0/lib/shoulda/matchers/active_record/have_rich_text_matcher.rb#L69-L71

VSPPedro:
I'm sorry, but I'm not sure how this fix will help you with the CI broking on #1350.

The problem there is ActionText::RichText being called without being defined. For that reason, we compared with a string as you can see in this comment: https://github.com/thoughtbot/shoulda-matchers/pull/1263#discussion_r354116064

So, the problem on CI is happening because there are tests when the ActionText is not defined: https://github.com/thoughtbot/shoulda-matchers/blob/master/spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb#L51

KapilSachdev:
While running tests of this repo, `actiontext` will not be available for `rails < 6`, so the condition is wrong `rails_version > 5`.

```ruby
defined?(ActionText::RichText) && @subject.send(rich_text_attribute).instance_of?(ActionText::RichText)
```
Yes, I would have to do that at #1350 .

I didn't wanted `rails` version changes to be part of #1350 because they don't belong there.
Though there had been some code  changes that rubocop forced to be done, but i'm trying for them to be minimal.

mcmire:
Ah, I see. The build isn't broken, this seems to be more of a consistency thing, as we don't check for a Rails version in the tests either: https://github.com/thoughtbot/shoulda-matchers/blob/59f7c0b6ac64312c5a55f67ddcf4c945c77cad4b/spec/unit/shoulda/matchers/active_record/have_rich_text_matcher_spec.rb#L5. Okay, that makes sense to me. Merging!

PR #1367, opened 2020-11-02
Install active_storage migrations

VSPPedro:
Suggestion: instead of creating the table manually, use the command 'bundle exec rake active_storage: install: migrations'.

### DIFF
--- a/spec/support/unit/rails_application.rb
+++ b/spec/support/unit/rails_application.rb
@@ -2,12 +2,15 @@ require_relative '../tests/bundle'
 require_relative '../tests/command_runner'
 require_relative '../tests/database'
 require_relative '../tests/filesystem'
+require_relative 'helpers/active_record_versions'
 require_relative 'helpers/rails_versions'
 
 require 'yaml'
 
 module UnitTests
   class RailsApplication
+    include UnitTests::ActiveRecordVersions
+
     def initialize
       @fs = Tests::Filesystem.new
       @bundle = Tests::Bundle.new
@@ -26,6 +29,7 @@ module UnitTests
     def load
       load_environment
 
+      add_active_storage_migration if active_record_supports_active_storage?
       add_action_text_migration if bundle.includes?('actiontext')
 
       run_migrations
@@ -168,6 +172,12 @@ end
       TEXT
     end
 
+    def add_active_storage_migration
+      fs.within_project do
+        run_command! 'bundle exec rake active_storage:install:migrations'
+      end
+    end
+
     def add_action_text_migration
       fs.within_project do
         run_command! 'bundle exec rake action_text:install:migrations'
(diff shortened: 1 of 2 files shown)


VSPPedro:
Just taking some notes.

Error on CI with Rails 5.2 and DATABASE_ADAPTER=postgresql:

```
Expected User to have a has_many_attached called avatars, but this could not be proved.

         Expected User to have a has_many association called avatars_attachments (ActiveStorage::Attachment does not have a record_id foreign key.)
```

VSPPedro:
This is weird. The active_storage tables are being created on postgres database instead of shoulda-matchers-test-development database on Rails 5.2. On Rails 6 it works and because of that, the tests pass. I will keep digging.

VSPPedro:
To solves this problem it's necessary to update Rails from 5.2.4.1 to 5.2.4.4.

I intend to do that in another PR and I noticed that last PR that updated the dependencies of this project was 10 months ago - https://github.com/thoughtbot/shoulda-matchers/pull/1269. I think it's time to do another one.

KapilSachdev:
`5.2.4.1` to `5.2.4.4` [diff](https://github.com/rails/rails/compare/v5.2.4.1...v5.2.4.4) does not show any related changes that might effect this issue.

I tried 
```ruby
DATABASE_ADAPTER=postgresql bundle exec appraisal rails_5_2 rspec spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```
with both the version and got the same errors.
Is upgrading working for you?

And about upgrading the dependencies, I too think we should do that.

VSPPedro:
No, I thought it worked, but now I see it didn't work. I probably ran with another version of Rails and thought it was 5.2. My bad.

Thank you for pointing this out. I will keep digging.

VSPPedro:
I don't know the reason yet but when running `bundle exec rake db:drop:all db:create:all db:migrate` on Rails 5.2 with postgres the last command don't work as expected.

The migrattion is not being runned on the test database - when I tried locally it was adding in the postgres database. Separating the commands work as you can see - the CI is green.

I tried to change to `bundle exec rake db:drop db:create db:migrate`, but I got this error as you can see in this [CI](https://travis-ci.org/github/thoughtbot/shoulda-matchers/jobs/741876695):

```
ActiveRecord::NoDatabaseError:

       FATAL:  database "shoulda-matchers-test_production" does not exist
```

I will continue digging to get a better understanding of why we need a production database at CI.

mcmire:
@VSPPedro I can answer that question. We need a "production" database (really just a secondary database, but managed via database.yml) in order to test the ability for `have_db_index` to be used with different models that connect to different databases, as added here: https://github.com/thoughtbot/shoulda-matchers/pull/1200.

VSPPedro:
Thank you for pointing this out.

I think it would be a good idea to change the names of these databases, but to be honest I can think of a better name now.

I'll leave that for later, in case I find better names and, of course, if you also think that this change would bring some value to the project.

PR #1366, opened 2020-11-02
Update examples

VSPPedro:
- [x] Add have_one_attached and have_many_attached examples
- [x] Replace is_expected.to with should (I noticed that the project has a preference for should and that is why I am sending this modification suggestion.)

PR #1368, opened 2020-11-02
Fix HaveRichTextMatcher class name

VSPPedro:
The name should be HaveRichTextMatcher and not HaveRichText.

Thank you, @KapilSachdev, you made me notice this mistake when you reviewed the #1366. Thank you very much.


mcmire:
Good catch :)

PR #1324, opened 2020-07-25
Release v4.4.0

mcmire:
### Features

* Add an `ignoring_check_for_db_index` qualifier to the `have_secure_token`
  matcher, since `has_secure_token` encourages use of an index but does not
  enforce it. ([#1278])

* Add `allow_blank` to `validate_length_of` to match other validation matchers.
  ([#725], [#1318])

* Add new matcher `have_implicit_order_column` which can be used to test the
  `implicit_order_column` setting for ActiveRecord models under Rails 6+.
  ([#1243])

* Add a new `is_other_than` qualifier to `validate_numericality_of` to be able
  to test the numericality validation's `:other_than` option. ([#1282])

* Add a new `have_one_attached` and `have_many_attached` matchers for testing
  the new model-level ActiveStorage macros in Rails 6. ([#1102])

[#1278]: https://github.com/thoughtbot/shoulda-matchers/pull/1278
[#725]: https://github.com/thoughtbot/shoulda-matchers/pull/725
[#1318]: https://github.com/thoughtbot/shoulda-matchers/pull/1318
[#1243]: https://github.com/thoughtbot/shoulda-matchers/pull/1243
[#1282]: https://github.com/thoughtbot/shoulda-matchers/pull/1282
[#1102]: https://github.com/thoughtbot/shoulda-matchers/pull/1102

### Bug fixes

* Fix performance of `allow_value` so that it doesn't hang if the given value is
  really long. ([#1290])

* Fix `have_many` so that it is possible to test an association that has a scope
  that takes an argument. ([#952], [#992])

* Update `validate_uniqueness_of` to use the public `validators_on` instead of
  the private `_validators` when reading validations off of a model. This
  enables shoulda-matchers to be used with the [schema_validations] gem.
  ([#995])

* Update `validate_uniqueness_of` to work with scopes that are `time` columns.
  ([#1190])

* Fix `have_and_belong_to_many` so that when using the `join_table` qualifier
  you can pass a symbol rather than a string. ([#1323])

[#1290]: https://github.com/thoughtbot/shoulda-matchers/issues/952
[#952]: https://github.com/thoughtbot/shoulda-matchers/issues/952
[#992]: https://github.com/thoughtbot/shoulda-matchers/pull/992
[schema_validations]: https://github.com/SchemaPlus/schema_validations
[#995]: https://github.com/thoughtbot/shoulda-matchers/pull/995
[#1190]: https://github.com/thoughtbot/shoulda-matchers/pull/1190
[...]

PR #1369, opened 2020-11-02
Use Ruby 2.7.2 for development and CI

VSPPedro:
I made the update following the changes made on this PR #1310.

What do you think about dropping the support for old Ruby versions and Rails 4.2 like suggested in this comment: https://github.com/thoughtbot/shoulda-matchers/pull/1324#issuecomment-664078521

One of the advantages would be the reduction of time of the CI.


mcmire:
@VSPPedro I think that depends on whether we are ready to talk about 5.0. If we're going to be talking about that then perhaps we should although think about other things we could possibly deprecate? For instance, I really wonder how often controller matchers get used (they're more useful in a Minitest context, but even controller tests are deprecated now in Rails). Do we really need database column matchers? etc. But maybe it's worth starting that discussion.

mcmire:
Aside from that, this looks good, so I'll merge :)

VSPPedro:
@mcmire I think it's worth starting that discussion too. Should we finish the open [milestone](https://github.com/thoughtbot/shoulda-matchers/milestone/14) first?

mcmire:
@VSPPedro Eh... maybe. 😆 I made that milestone a while back just so the next release would be somewhat focused instead of cramming as much stuff as possible into one release as I have done in years past. Also, with every new release, I have tried to make sure I am addressing old issues as well as new ones, but at this point that strategy has failed too 🙄

So... I say if we want to release a new major version, we should focus on that and everything that entails and not worry so much about what it is in the backlog (unless there's something we've always wanted to fix that requires a backward incompatible change — although I'm having a hard time thinking of an example right now).

VSPPedro:
I'll try to help to finish that milestone. :smile:

KapilSachdev:
Good catch @VSPPedro . I have created `4.5.0` milestone, and is adding PRs that are merged after 4.4.1 under it. I hope i don't miss any.

> So... I say if we want to release a new major version

I think our major release (`5.0.0`) will be the one where we would be removing the support for rails `4.2` for which we would be creating a deprecation workflow.
So I have taken the liberty to create a `4.5.0` milestone which would be our next version.

mcmire:
@VSPPedro I gave @KapilSachdev commit access and it sounds like you may want to do more work so I have given you commit access as well. I will let y'all coordinate on what you want to include in 4.5.0 :)

VSPPedro:
Thank you very much!

PR #1371, opened 2020-11-04
feat: Add github action for linting rubocop

### LINKED ISSUE #1345: Replace Hound with GitHub Action
In the past, this gem was using Hound (developed by some now ex-thoughtbotters) to run code through Rubocop. However, it doesn't seem to be maintained much these days — I actually disabled Hound a while back because it started choking on new versions of Rubocop, starting with the Rubocop team's decision to spin `rubocop-rails` off. Since then GitHub Actions has emerged as a more built-in solution for these types of tasks. I believe there is a linter action out there that integrates with Rubocop, and I think we should try to use that instead.

KapilSachdev:
Closes #1345

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,38 @@
+name: Rubocop
+
+on: [push, pull_request]
+
+jobs:
+  run-linters:
+    name: Run linters
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
+
+      - name: Set up bundle cache
+        uses: actions/cache@v2
+        with:
+          path: /vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('./Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+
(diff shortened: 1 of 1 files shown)


mcmire:
Any way to run this action with an intentional Rubocop violation? Wondering if the `[skip ci]` is also preventing the action from getting run.

KapilSachdev:
I did try its behaviour [before](https://github.com/KapilSachdev/shoulda-matchers/pull/1/files) creating a PR here, but it seems the action didn't run here which i thought would run.
I don't think [ci skip] have any effect, as I force pushed it twice and previous commit message does not had [ci skip] and the action didn't ran.
I will force push again without [ci skip], just to verify.

KapilSachdev:
Amending the commit and PR text with no [ci skip] does not had any effect.
Might be related to repository permissions? I'm not sure, but it seems to be working in my forked one.

mcmire:
Weird. Just checked out your repo though and everything looks good here.

By the way I see that you've been contributing a lot lately. Do you plan on doing more work? I can give you commit access if so.

KapilSachdev:
>Do you plan on doing more work? 

Yes.

mcmire (approved):

mcmire:
Okay you have commit access now, feel free to merge at will!

KapilSachdev:
Thank You !

KapilSachdev:
As per wearerequired/lint-action#13, The action will give 403 against the fork.
Taking liberty to merge thinking it would work. (as fork one is working properly)

PR #1372, opened 2020-11-04
Fix module

VSPPedro:
I find this PR([Move uniqueness validation to ActiveRecord module](https://github.com/VSPPedro/shoulda-matchers/commit/cf76d747eef4e04a787fb43badfaa2d459aa5021)) and it seems that this file was the only one missing the update on the module.


mcmire:
Good catch @VSPPedro! Thank you!

PR #1373, opened 2020-11-04
Fix 'required_ruby_version: unbound variable'

VSPPedro:
Running `.bin/setup` on ruby 2.7.2 was returning:

```ruby
./bin/setup: line 240: required_ruby_version: unbound variable
./bin/setup: line 241: required_ruby_version: unbound variable
```

This PR solves this problem, but please test it locally. Thank you.

PS: If you take a look at the [bin/setup](https://github.com/thoughtbot/shoulda-matchers/blob/master/bin/setup) file, you will notice that in some places the variable name is $REQUIRED_RUBY_VERSION and in others, it is $required_ruby_version.

Which one do you prefer?

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -17,7 +17,7 @@ provision-project() {
 
 ### DON'T MODIFY ANYTHING BELOW THIS LINE! #####################################
 
-# This setup script was generated with setup_script_generator 0.2.7,
+# This setup script was generated with setup_script_generator 0.3.1,
 # available on RubyGems.
 #
 # To regenerate this section, install the gem and run:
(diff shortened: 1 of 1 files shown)


mcmire:
@VSPPedro Yes, this seems to work for me. I would prefer `REQUIRED_RUBY_VERSION` (I think my original intent was that this would be a sort of global) so your change is fine with me.

By the way, this script is generated via my [`setup_script_generator`](https://github.com/mcmire/setup_script_generator) gem. Instead of making changes directly to the script like this, it might be better to regenerate the script using the gem, otherwise a future update might blow away those changes. I've actually removed the `rvm` section entirely in the latest update since I don't use RVM myself and haven't run across anyone (yet) who uses it, so this problem would go away with that (or, if you do want to use it, I could add that section back). The latest update also contains a fix for installing Bundler 1.x — it seems that that the `gem install GEM -v VERSION` syntax has been replaced with `gem install GEM:VERSION`.

With that in mind, what do you think about running:

``` ruby
gem install setup_script_generator
generate-setup bin/setup --provision ruby
```

VSPPedro:
Whops, I did notice this comment: [DON'T MODIFY ANYTHING BELOW THIS LINE!](https://github.com/thoughtbot/shoulda-matchers/blob/master/bin/setup#L18). My bad! :sweat_smile:

I use RVM, but you don't need to add that section for me. When I need it I will install ruby ​​manually, no problem.

> With that in mind, what do you think about running:
> 
> ```ruby
> gem install setup_script_generator
> generate-setup bin/setup --provision ruby
> ```

Done!

mcmire (approved):
@VSPPedro Oh sorry! If you use RVM I am happy to add that section (or a similar section) back to `setup_script_generator`. I just had no way of testing it myself so I couldn't know if it worked or not so that's why it was deleted.

In any case, I just tested `bin/setup` on this branch on my computer and it seems to work, so this looks good to me 👍

VSPPedro:
Thanks! Merging!

PR #1376, opened 2020-11-15
fix: with_foreign_key modifier fails with wrong error message

### LINKED ISSUE #1374: with_foreign_key modifier fails with no error message
Steps to replicate:

```ruby
# migration
create_table :foos do |t|
  t.references :bar, foreign_key: true
  t.timestamps
end

create_table :bars do |t|
  t.timestamps
end

# app/model/foo.rb
class Foo < ApplicationRecord
  belongs_to :pub, class_name: 'Bar', foreign_key: :bar_id
end

# app/model/bar.rb
class Bar < ApplicationRecord
  has_many :foos, inverse_of: :pub
end

# spec/models/foo_spec.rb
...
it { is_expected.to belong_to(:pub).class_name('Bar').with_foreign_key(:bar_id) }
...
```

The test passes but if I change the argument of `with_foreign_key` to anything else (e.g. `baz_id`) the spec fails with the following error message:

```
Expected Foo to have a belongs_to association called pub ()
```

If the `class_name` argument is changed to be incorrect there is a specific error message in the parentheses for the submatcher (`pub should resolve to Baz for class_name`), so I'm guessing that somehow the following lines are no longer working as expected: https://github.com/thoughtbot/shoulda-matchers/blob/b7e02184d99014dc57d9e37bc134c4b6c5254b69/lib/shoulda/matchers/active_record/association_matcher.rb#L1404-L1407

Hope that helps!

VSPPedro:
Closes #1374

I noticed that for the scenario described on issue #1374 the class_has_foreign_key?(klass) was entering in the first conditional and than running this piece of code:

```
option_verifier.correct_for_string?(
  :foreign_key,
  options[:foreign_key],
)
```

It seems to me that the problem with this flow is that it just returns false without updating the failure message. What do you think?

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -1394,17 +1394,26 @@ module Shoulda
         end
 
         def class_has_foreign_key?(klass)
-          if options.key?(:foreign_key)
-            option_verifier.correct_for_string?(
-              :foreign_key,
-              options[:foreign_key],
-            )
-          elsif column_names_for(klass).include?(foreign_key)
-            true
-          else
-            @missing = "#{klass} does not have a #{foreign_key} foreign key."
+          if options.key?(:foreign_key) && !foreign_key_correct?
+            @missing = foreign_key_failure_message(klass, options[:foreign_key])
             false
+          elsif !column_names_for(klass).include?(foreign_key)
+            @missing = foreign_key_failure_message(klass, foreign_key)
+            false
+          else
+            true
           end
+        end
+
+        def foreign_key_correct?
+          option_verifier.correct_for_string?(
+            :foreign_key,
+            options[:foreign_key],
+          )
+        end
+
+        def foreign_key_failure_message(klass, foreign_key)
+          "#{klass} does not have a #{foreign_key} foreign key."
         end
 
         def primary_key_correct?(klass)
--- a/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
@@ -1414,6 +1414,20 @@ Expected Parent to have a has_many association called children through conceptio
       expect(having_one_non_existent(:person, :detail, class_name: 'Detail')).not_to have_one(:detail)
     end
 
+    it 'rejects an association with a valid :class_name and a bad :primary_key option' do
+      define_model :person_detail
+      define_model :person do
+        has_one :detail, class_name: 'PersonDetail', foreign_key: :person_detail_id
+      end
+
+      expected_message = 'Expected Person to have a has_one association called detail ' \
+        '(PersonDetail does not have a custom_primary_id foreign key.)'
+
+      expect { have_one(:detail).class_name('PersonDetail').with_foreign_key(:custom_primary_id) }.
+        not_to match_against(Person.new).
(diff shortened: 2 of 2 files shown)


VSPPedro:
There is something wrong with the Run linters action:

```
Error: Error trying to create GitHub check for RuboCop: Received status code 403
    at createCheck (/home/runner/work/_actions/wearerequired/lint-action/v1/src/github/api.js:64:9)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
/home/runner/work/_actions/wearerequired/lint-action/v1/src/index.js:13
	throw new Error(`Exiting because of unhandled promise rejection`);
	^

Error: Exiting because of unhandled promise rejection
    at process.<anonymous> (/home/runner/work/_actions/wearerequired/lint-action/v1/src/index.js:13:8)
    at process.emit (events.js:210:5)
    at processPromiseRejections (internal/process/promises.js:201:33)
    at processTicksAndRejections (internal/process/task_queues.js:94:32)
```

I will try to understand the problem.

mcmire:
Hmm, it seems to be related to this problem: https://github.com/wearerequired/lint-action/issues/13. I think since you have commit access now you should be able to push a branch directly to the repo instead of making a PR from your fork. That may be a solution for now.

mcmire (approved):
I'm not sure why this bug wasn't caught before but your fix seems right to me!

VSPPedro:
I see. For the next PR I will not use my fork anymore. Thanks for the suggestion.

So, for now, if anyone opens a PR we will need to ask them to run the rubocop locally just to make sure that they're not adding any offense.

I did it on this PR and there is no offense. Can I merge it or do you prefer that I open a new PR not using my fork?

KapilSachdev:
> So, for now, if anyone opens a PR we will need to ask them to run the rubocop locally just to make sure that they're not adding any offense.

That will defeat the purpose of automating it.

> I'm not sure why this bug wasn't caught

I was aware of the issue but i didn't anticipated this effect. I misunderstood that issue in some way thinking that the token was of the fork so it would not add the annotation on the destination repo, but once #1371  is merged this will not happen.

@mcmire what do you think about removing `lint-action`? This issue arises because github does not allows it so we don't know the timeline of this fix. 
We can add a custom github action for rubocop, but without annotation (eg https://github.com/rails/rails/runs/1403456898).
Initially i was into adding this but we were interested in annotation so this was cancelled.

KapilSachdev:
`pull_request_target` does not looks like a reliable fix either. 
I created a [PR](https://github.com/KapilSachdev/shoulda-matchers/pull/5/checks?check_run_id=1407720196) in my fork from another fork and the rubocop action failed due to a reason reported in that same thread of lint-action and there does not seems to be any response against it.

So for now using a custom rubocop lint action without annotation seems a reasonable compromise as hound does not support [rubocop > 0.91](http://help.houndci.com/en/articles/2461415-supported-linters)

@mcmire Thoughts?

mcmire:
@KapilSachdev Dang, that's annoying. Well, okay, I guess a lint action that doesn't support annotations is better than not having one. So I'd be fine with that.

VSPPedro:
I noticed that a fix for the CI was recently added on master. I'll rebase this branch to check if it's now passing.

KapilSachdev:
Yes, the annotation based rubocop linting is replaced with a simple workflow to check rubocop offences.

VSPPedro:
It worked. Merging. Thanks!

PR #1377, opened 2020-11-16
fix: Remove lint-action github action and add custom RuboCop action

KapilSachdev:
- Annotations feature provided by lint-action is not working with PRs from fork repos
- pull_request_target feature is also not behaving reliably 
- Remove hound
- Note:Not using `bundler-cache: true` for caching as it uses uses `./vendor/bundle` as bundle path and `rubocop` lint vendor as well. So caching under `../vendor/bundle` using `actions/cache`
- Using `actions/cache@v1` rather `v2` as gems are not getting cached due to unknown reason

Related: #1345, #1371

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,38 +1,30 @@
-name: Rubocop
+name: RuboCop
 
 on: [push, pull_request]
 
 jobs:
-  run-linters:
-    name: Run linters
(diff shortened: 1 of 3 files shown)


mcmire (approved):
Okay! Looks good to me 👍

KapilSachdev:
Thank you.
Merging.

PR #1378, opened 2020-11-17
fix: actual_value_for(validate) returns false instead of true when there is no validate option

### LINKED ISSUE #1357: have_many(:model).validate(false) causes a false positive test
I run into a problem where I had to make sure that an association has a `validate: false` on the association. I started it with writing a test for it, but it passed immediately without making the required code change.

I created this code snippet to reproduce the issue: https://gist.github.com/kiskoza/da2b4a0ba6930e9454403ea36e3fd73c

I think the problem is that in my case there weren't a `validate` option prior to the task. When it was [validated as a boolean](https://github.com/thoughtbot/shoulda-matchers/blob/256d3b5ac9a9031af7b7096fbcbb72272784000b/lib/shoulda/matchers/active_record/association_matcher.rb#L1360-L1367), the `nil` value was simply [converted to](https://github.com/thoughtbot/shoulda-matchers/blob/256d3b5ac9a9031af7b7096fbcbb72272784000b/lib/shoulda/matchers/active_record/association_matchers/option_verifier.rb#L71-L72) `false`

Docs about the validate option: https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_many

VSPPedro:
Closes #1357

It seems to me that the problem here is that when there is no validate option on the relationship the value being returned by `actual_value_for(name)` is nil instead of true - the default value.

In case you want to check the default value for this macro: [documantation](https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_many) and [active_record/associations/has_many_association.rb#L56](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/associations/has_many_association.rb#L56).

To avoid this I added a check to validate the value and if it is null the default value will be returned instead, otherwise the code will follow its normal flow. I added some tests just to be sure about the behaviour. What do you think?

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matchers/option_verifier.rb
+++ b/lib/shoulda/matchers/active_record/association_matchers/option_verifier.rb
@@ -6,6 +6,11 @@ module Shoulda
         class OptionVerifier
           delegate :reflection, to: :reflector
 
+          DEFAULT_VALUE_OF_OPTIONS = {
+            has_many: {
+              validate: true,
+            },
+          }.freeze
           RELATION_OPTIONS = [:conditions, :order].freeze
 
           def initialize(reflector)
@@ -55,7 +60,7 @@ module Shoulda
               if respond_to?(method_name, true)
                 __send__(method_name)
               else
-                reflection.options[name]
+                actual_value_for_option(name)
               end
             end
           end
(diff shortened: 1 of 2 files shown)


mcmire (approved):
Okay! This looks like a good fix to me. 👍

VSPPedro:
Thank you! Merging.

PR #1383, opened 2020-12-02
Fix validate_absence_of failing for array columns

### LINKED ISSUE #1240: validate_absence_of fails on Postgres array columns
It looks like `validate_absence_of` doesn't work with array columns. For example, given a Rails application (using Postgres) that has migration with this line:

```ruby
t.string :widgets, array: true
```

and the following model validations:

```ruby
validates :widgets, absence: true
```

and the following test:

```ruby
it { should validate_absence_of(:widgets) }
```

the matcher fails and produces the error below:

```
Expected SomeModel to validate that :widgets is empty/falsy, but
this could not be proved.
  After setting :widgets to ‹"an arbitrary value"› -- which was
  read back as ‹[]› -- the matcher expected the SomeModel to be
  invalid, but it was valid instead.

  As indicated in the message above, :widgets seems to be
  changing certain values as they are set, and this could have something
  to do with why this test is failing. If you've overridden the writer
  method for this attribute, then you may need to change it to make this
  test pass, or do something else entirely.
```

I think this is happening because the value chosen to test for presence ([here](https://github.com/thoughtbot/shoulda-matchers/blob/c32307fe3309fefea0398f6cd4e850d8e992046c/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb#L111)) is the string `"an arbitrary value"`. Because it's setting an array column, the string is being converted to `[]`, so the test fails.

Ideally it would know that it's an array column and use something like `['an arbitrary value']` instead. There is an `array` method it could use in the `Column` class: `@subject.class.columns_hash[@attribute.to_s].array`.

thealiilman:
My very first time contributing to this gem. 🤓  I love that the [contributing guide](https://github.com/thoughtbot/shoulda-matchers/blob/master/CONTRIBUTING.md) is well-written! 🎉 

This fixes #1240.

## Changelog
- Made `ValidatingAbsenceOfMatcher#value` check whether column is array
- Added tests on array columns for `ValidateAbsenceOfMatcher`

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
@@ -107,6 +107,8 @@ module Shoulda
             else
               obj
             end
+          elsif array_column?
+            ['an arbitary value']
           else
             case column_type
             when :integer, :float then 1
@@ -136,6 +138,12 @@ module Shoulda
         def reflection
           @subject.class.respond_to?(:reflect_on_association) &&
             @subject.class.reflect_on_association(@attribute)
+        end
+
+        def array_column?
+          @subject.class.respond_to?(:columns_hash) &&
+            @subject.class.columns_hash[@attribute.to_s].respond_to?(:array) &&
+            @subject.class.columns_hash[@attribute.to_s].array
         end
       end
     end
--- a/spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb
@@ -48,6 +48,49 @@ describe Shoulda::Matchers::ActiveModel::ValidateAbsenceOfMatcher, type: :model 
         end
       end
 
+      if database_supports_array_columns? && active_record_supports_array_columns?
+        context 'when the column backing the attribute is an array' do
+          context 'of varchar' do
+            it 'still works' do
+              record = validating_absence_of(
+                :attr,
+                {},
+                type: :varchar,
+                options: { array: true, default: [], null: false },
+              )
+
+              expect(record).to validate_absence_of(:attr)
+            end
+          end
+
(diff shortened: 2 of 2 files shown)


thealiilman:
Hmm, the jobs for the build on Travis CI seem to be stuck in _Waiting to be queued_ state. Travis CI is operating though. 🤔

mcmire (approved):
@thealiilman Very nice, looks perfect to me! In terms of CI, Travis just implemented some changes recently to open-source projects, so it's possible that we've exceeded the limits or something. I've restarted the build to see if that helped. Once we can get this to pass, this is good to go.

thealiilman:
> @thealiilman Very nice, looks perfect to me! In terms of CI, Travis just implemented some changes recently to open-source projects, so it's possible that we've exceeded the limits or something. I've restarted the build to see if that helped. Once we can get this to pass, this is good to go.

Aha, I see. That may well be the cause.

Thanks for the approval!

mcmire:
Merged. Thank you so much!

VSPPedro:
Thank you, @thealiilman!

@mcmire, we now have three bug fixes on the master. What do you think about a new release(v4.4.2) for them or do you have another idea in mind? 

If you would like some help with that please let me know.

### Bug fixes

* https://github.com/thoughtbot/shoulda-matchers/pull/1376
* https://github.com/thoughtbot/shoulda-matchers/pull/1378
* https://github.com/thoughtbot/shoulda-matchers/pull/1383

mcmire:
@VSPPedro Oops, sorry, I thought I'd already replied to this. That might be best. I will hold off on merging anything else :) 

I am definitely not opposed to others handling releases, so if that's something you'd like to do, then that's totally cool. Let me know and I will make sure you are able to do this. Note that I have a PR around clarifying the instructions in `MAINTAINING.md` around releasing so hopefully it is relatively easy to follow them once that time comes: https://github.com/thoughtbot/shoulda-matchers/pull/1384

PR #1387, opened 2020-12-20
Bring CHANGELOG up to date

VSPPedro:
There were many PRs merged.

~I'm not sure if I should add all of them or just those that affect the developer using this gem in some way.~

If you would like to change anything, please, let me know.

EDIT: I'll add all.

[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,39 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+* Fix `bundle exec yard doc` replacing pygments.rb with rouge to keep the documentation up to date. ([#1343])
+* Fix have_db_column.with_options not raising an error when the user adds an invalid value. ([#1358])
+* Fix default value of assossiation_matcher.validate option. It was returning false instead of true. ([#1378])
+* Fix validate_absence_of failing for array columns. ([#1383])
+
+[#1340]: https://github.com/thoughtbot/shoulda-matchers/pull/1340
+[#1358]: https://github.com/thoughtbot/shoulda-matchers/pull/1358
+[#1378]: https://github.com/thoughtbot/shoulda-matchers/pull/1378
+[#1383]: https://github.com/thoughtbot/shoulda-matchers/pull/1383
+
+### Features
+
+* Add support for has_secure_password custom attributes. ([#1356])
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,46 @@
...
 
+## Unreleased
+
+### Bug fixes
+
+* Rename remaining NEWS to CHANGELOG. ([#1308])
+* Fix links in CHANGELOG. ([#1339])
+* Replace all links with http to https. ([#1340])
+* Fix have_db_column.with_options misspelled options. ([#1358])
+* Improve error message for assossiation_matcher.with_foreign_key option. ([#1376])
```
mcmire, line 11: In general it might be best to put updates to error messages — or what I would consider to be changes to "user experience" — in Improvements. You're not really fixing broken functionality, just making the software easier to use in some way.

mcmire (commented):
@VSPPedro I probably didn't clarify what should go inside of the CHANGELOG. In general I've only been adding changes that are useful for end users. Think about if you had an app in the App Store and you were listing things that people could see. So this would include things like:

* new matchers
* changes to existing matchers
* changes to user-visible documentation
* changes to gemspec (bc this could affect installation)

but would not include:

* updates to the setup script
* updates to Rubocop
* updates to the Travis config or GitHub Actions config
* fixes to the CHANGELOG itself (that's very meta 😄 )

When listing items in the CHANGELOG, I'd also highlight the end result (the why) instead of the specific thing being added (the what). What people want to know is, what problem did this fix for me, or what can I accomplish now that I couldn't before? That's really what this CHANGELOG exists for. So you have to understand each problem being faced and figure out a short way to describe them. (I know that people can look at the original issues/PRs themselves, but people including myself tend to be lazy.) In the future I would probably recommend adding to the CHANGELOG as new things are merged rather than doing a full sweep before a new version release, because otherwise you tend to forget or you have to look back at the original issues to reconstruct the context.

Anyway... I've highlighted some examples below, hopefully that clarifies what I mean.

VSPPedro:
> fixes to the CHANGELOG itself (that's very meta)

😆 

I think I understand now what I need to do.

@mcmire, thank you very much!

mcmire (approved):
Okay, looks good! Nice job.

Issue #1392, opened 2021-01-12
v4.5.0 breaks Composite Primary Keys


natematykiewicz:
We use [composite primary keys](https://github.com/composite-primary-keys/composite_primary_keys). I believe this PR has caused an incompatibility https://github.com/thoughtbot/shoulda-matchers/pull/1376.

# Spec
```ruby
it do
  is_expected.to belong_to(:vehicle).
    with_foreign_key(%i[company_id vehicle_id]).
    inverse_of(:vehicles_options)
end
```

# Model
```ruby
class VehiclesOption
  belongs_to :vehicle,
    foreign_key: %i[company_id vehicle_id],
    inverse_of: :vehicles_options
end
```

With v4.4.1 this spec would succeed.
With v4.5.0 it fails with:
> Expected VehiclesOption to have a belongs_to association called vehicle (VehiclesOption does not have a [:company_id, :vehicle_id] foreign key.)

As a test of v4.4.1 I tried removing the foreign key from the model, and also changing one of the values of the array:
Ex:

```ruby
belongs_to :vehicle,
  inverse_of: :vehicles_options

belongs_to :vehicle,
  foreign_key: %i[company_id foo_id],
  inverse_of: :vehicles_options
```

Both of these fail with:
> Expected VehiclesOption to have a belongs_to association called vehicle ()

This is the error message that #1376 aimed to fix. This also shows me that the foreign key check was indeed working (I was worried I had a false positive before).

Looking at the diff, I'm not sure why this isn't working anymore.

VSPPedro:
Thank you for all the details. I'll take a look at that as soon as possible.

natematykiewicz:
@VSPPedro I think I see the problem. With v4.4.1, if we were providing a `:foreign_key` option, it would only run the `option_verifier.correct_for_string?`. The result of that determined whether the test passes.

With v4.5.0 it runs `options.key?(:foreign_key) && !foreign_key_correct?`. Since we have a `options.key?(:foreign_key)`, it then checks `foreign_key_correct?`. Since `foreign_key_correct?` returns true (all is well), it then carries on to the `elsif`.

The `column_names_for(klass).include?(foreign_key)` part never ran in v4.4.1 when `options.key?(:foreign_key)` was provided.

In my case `foreign_key` is actually an array. It sort of needs to do:

```ruby
Array(foreign_key).all? { |f| column_names_for(klass).include?(f) }
```

[...]

PR #1393, opened 2021-01-13
refactor class_has_foreign_key method from association_matcher.rb

VSPPedro:
I'm just trying to clean up a mess I made. I believe that this way it is easier to read. What do you think?

It only passes if there is no error message.


VSPPedro:
I changed my mind. I think this refactor can be added with the change requested by the #1392

PR #1397, opened 2021-01-14
Add GitHub Action to replace Travis

### LINKED ISSUE #1395: Switch to Github Actions or Buildkite
I think we need to change the CI before this happens: `Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.`

You can see this warning at the header on their site: https://travis-ci.org/

I'm creating this issue so I don't forget that detail.

VSPPedro:
Closes #1395


VSPPedro:
There were two problems with this update:

1. It was not possible to access the Postgres database, if you want to see this behavior locally you can create a docker-compose.yml like this:

```
version: "3.7"

services:
  postgres:
    image: postgres:12.1
    ports:
      - 5432:5432
    environment:
      POSTGRES_PASSWORD: postgres
    volumes:
      - ./postgres_data:/var/lib/postgresql/data

volumes:
  postgres_data:
    driver: local
```

And then if you try to run this command:

```
DATABASE_ADAPTER=postgresql bundle exec appraisal rails_6_0 rake spec:unit
``` 

You will face this error:

```
shoulda-matchers git:(master) DATABASE_ADAPTER=postgresql bundle exec appraisal rails_6_0 rake spec:unit
>> BUNDLE_GEMFILE=/home/pedro/tempo-livre/open-source/shoulda-matchers/gemfiles/rails_6_0.gemfile bundle exec rake spec:unit

An error occurred while loading ./spec/unit/shoulda/matchers/action_controller/callback_matcher_spec.rb.
Failure/Error:
        raise <<-MESSAGE
  Command #{formatted_command.inspect} exited with status #{exit_status}.
  Output:
  #{divider('START') + output + divider('END')}
        MESSAGE

RuntimeError:
  Command "bundle exec rake db:drop:all db:create:all db:migrate" exited with status 1.
  Output:
  ---START----------------------------------------------------------------
  rake aborted!
  PG::ConnectionBad: could not connect to server: No such file or directory
  	Is the server running locally and accepting
  	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

The modification on `spec/support/tests/database_configuration.rb` is to fix this.

2. bundle exec rake --trace is running with the wrong gemfile - [example](https://github.com/VSPPedro/shoulda-matchers/runs/1715302869?check_suite_focus=true). The Gemfile for the example is rails_6_0, but as you can see the command is calling the rails_4_2:
```
Run bundle exec rake --trace
** Invoke default (first_time)
** Execute default
>> bundle check --gemfile='/home/runner/work/shoulda-matchers/shoulda-matchers/gemfiles/rails_4_2.gemfile' || bundle install --gemfile='/home/runner/work/shoulda-matchers/shoulda-matchers/gemfiles/rails_4_2.gemfile' --retry 1
Resolving dependencies...
Bundler can't satisfy your Gemfile's dependencies.
Install missing gems with `bundle install`.
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "actionpack":
  In snapshot (rails_4_2.gemfile.lock):
    actionpack (= 4.2.11.1)

  In rails_4_2.gemfile:
    actiontext (~> 6.0.2.1) was resolved to 6.0.2.1, which depends on
      actionpack (= 6.0.2.1)

    rails (= 4.2.11.1) was resolved to 4.2.11.1, which depends on
      actionpack (= 4.2.11.1)
```

So, instead of using the command `bundle exec rake` I split them on two and I'm informing the gemfile that it's suppose to use:

```
- name: Run Unit 
  run: bundle exec appraisal ${{ matrix.appraisal }} rake spec:unit --
- name: Run Acceptance Tests
  run: bundle exec appraisal ${{ matrix.appraisal }} rake spec:acceptance --trace
```

What do you think?

To be honest, I'll try to improve this. I'm just wondering what are your thoughts on this, @mcmire.

natematykiewicz:
If you can get it to work, I'd recommend not using actions/cache, and instead using the bundler-cache option in ruby/setup-ruby.

https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

natematykiewicz:
@VSPPedro, the bundler-cache option installs as well. So you shouldn't need to do that yourself.

VSPPedro:
Ops, nice! I'll try to use that. Somehow is breaking now. I think I made some silly mistake like a typo or something. Tomorrow I'll try again.

Thanks for your suggestion.

spec/support/tests/database_adapters/config/postgresql.yml
```diff
@@ -0,0 +1,19 @@
+default: &default
+  adapter: postgresql
+  encoding: unicode
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5) %>
+  host: <%= ENV.fetch("DB_HOST", "localhost") %>
+  username: <%= ENV.fetch("DB_USER", "postgres") %>
```
mcmire, line 6: You might want to double-check that this doesn't break local development. I don't know if we can assume a `postgres` user always exists locally (for instance, I think on Mac, when using Postgres from Homebrew, a role is created that matches your username). We might need to use `ENV["CI"]` to determine whether to use `postgres` or not.

mcmire:
With regard to why the acceptance tests aren't working, I think I might know why. All of the acceptance tests are set up such that they create a temporary Rails application on the fly, then run a subcommand within that Rails application. That is, the command that you see failing [here](https://github.com/thoughtbot/shoulda-matchers/pull/1397/checks?check_run_id=1725088097) is being run in a different directory than the main shoulda-matchers project directory. Don't worry that the gemfile is being overridden here — what matters is where the gems it expects to be there are located.

In this GitHub Action, it appears a custom path is being used to install gems (`vendor/bundle`). That's not a surprise, as that is the norm for CI. This path is set in this case via the [`setup-ruby` action via `bundle config`](https://github.com/ruby/setup-ruby/blob/master/bundler.js#L115). `bundle config` in turn saves the setting to `.bundle` Since `bundle config` is being run in the project directory, that means `.bundle` is also located in the project directory.

Everything is fine until we change directories. Once we go outside of the project directory, `.bundle` is not in the same directory that `bundle exec` is being run in. What I think is happening is that the `path` setting that `setup-ruby` sets is getting lost, and Bundler is trying to find the gems in the global space, which of course don't exist. (The reason we never ran into this issue on Travis is because we [overrode the `install` step](https://github.com/thoughtbot/shoulda-matchers/blob/master/.travis.yml#L39) to _not_ set a custom path but installed everything to the global space instead, then cached those global directories [here](https://github.com/thoughtbot/shoulda-matchers/blob/master/.travis.yml#L30-L33). This also explains why appraisal tests "just work" locally as well — again, we assume you're installing all gems to one common global area in development.)

So I'm not really sure what the best solution here is. We could certainly turn off the Bundler mechanism in `setup-ruby` and handle that step ourselves, but then we'd have to figure out where the global gem installation directory is under GitHub Actions so that we could cache it explicitly and hardcode it into the workflow. We could also have the acceptance tests look for a `.bundle` directory to exist in the project directory and copy it over to the app directory after generating it (modifying `.bundle/config` on the spot so that the location of `vendor/bundle` is correct). The first option is faster but more liable to break in the future; the second option requires adding some minutiae to our test suite that's specific to CI but may work better in the long run. 

What do we think?

natematykiewicz:
I'd personally recommend option 2. Here's what setup-ruby (a repo provided by the official Ruby account) says:


Caching bundle install manually

It is also possible to cache gems manually, but this is not recommended because it is verbose and very difficult to do correctly. There are many concerns which means using actions/cache is never enough for caching gems (e.g., incomplete cache key, cleaning old gems when restoring from another key, correctly hashing the lockfile if not checked in, OS versions, ABI compatibility for ruby-head, etc). So, please use bundler-cache: true instead and report any issue.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      DATABASE_ADAPTER: ${{ matrix.adapter }}
+      BUNDLE_GEMFILE: gemfiles/${{ matrix.appraisal }}.gemfile
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+      - name: Run Unit Tests
+        run: bundle exec appraisal ${{ matrix.appraisal }} rake spec:unit --trace
```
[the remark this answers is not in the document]
VSPPedro: Oops, fixed!

spec/support/tests/database_adapters/config/postgresql.yml
VSPPedro: > I don't know if we can assume a postgres user always exists locally (for instance, I think on Mac, when using Postgres from Homebrew, a role is created that matches your username).

For this kind of scenario, the user can configure the necessary variable. Example:

```
DATABASE_ADAPTER=postgresql DB_USER=vsppedro bundle exec appraisal rails_6_0 rake spec:acceptance
```
It will be used vsppedro as username instead of postgres, because the method .fetch tries to catch the environment variable first, if returns nill it uses the second value.

A double-check would be nice. I'll ask some friends to run this command locally. I only can test this on Linux. 😅

VSPPedro:
I'll try to make it work with the second option. I think I can learn one thing or two in the process.

VSPPedro:
I haven't given up yet. I just migrated to the action / cache to see if it still worked. Tomorrow I intend to comment on this [issue](https://github.com/ruby/setup-ruby/issues/136) of the setup-ruby about the challenge of this PR. Maybe I can find some help.

VSPPedro:
FWIW, in case anyone wants to know how the cache works. I can’t point to the exact line because github doesn't show the content of the [file](https://github.com/ruby/setup-ruby/blob/master/dist/index.js), it’s too big, but in case you want to check where the cache key is created go to line 44663.

I’ll use this [job](https://github.com/VSPPedro/shoulda-matchers/runs/1739207905) as an example. If you take a closer look at the log created by `bundle install` you will find the cache key:

```
Cache key: setup-ruby-bundler-cache-v3-ubuntu-18.04-ruby-2.7.2-gemfiles/rails_6_0.gemfile.lock-8b40e73806da41ad164d0da2df40a58151cfc65e60754995bbc5148c92886b8f
```

As you can see the path to the gemfile.lock is being used. Makes total sense. So, let’s talk about the tests.

Unit tests pass and acceptance tests don’t, why? Why is the cache found in only one of them? What is the difference between them?

And the answer is the BUNDLE_GEMFILE.

Value for the unit tests: gemfiles/rails_6_0.gemfile
Value for the acceptance tests: /tmp/shoulda-matchers-acceptance/test-project/Gemfile

In the first test it tries to find the cache with:

```
Cache key: setup-ruby-bundler-cache-v3-ubuntu-18.04-ruby-2.7.2-gemfiles/rails_6_0.gemfile.lock-8b40e73806da41ad164d0da2df40a58151cfc65e60754995bbc5148c92886b8f
```

But in the second it tries with:

```
Cache key: setup-ruby-bundler-cache-v3-ubuntu-18.04-ruby-2.7.2-/tmp/shoulda-matchers-acceptance/test-project/Gemfile.lock-{{ Creates another hash using the gemfile lock name }}
``` 

And that’s why it can’t find the cache.

“So, if we don’t change the gemfile the cache will work?” - yes and no - you can see the result [here](https://github.com/VSPPedro/shoulda-matchers/runs/1739207905). The cache works, but the acceptance tests fail because this new project is not supposed to be executed with the other Gemfile. ShouldaMatchers gem that is added dynamically is missing. Error message:

Screenshot from 2021-01-21 22-49-44

setup-ruby does not allow you to customize the cache key. So, I'll keep digging.

bin/ci/install
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+bundle config set path $PWD/vendor/bundle
+bundle install --jobs=3 --retry=3
+
+if ! which zeus &>/dev/null; then
```
mcmire, line 6: You can drop these three lines in this case. For super_diff I'm using zeus in CI to speed up integration tests, but we don't have that for shoulda-matchers.

spec/support/tests/database_adapters/config/postgresql.yml
mcmire: Yeah, I guess it's not very common to run the Postgres tests locally, so if the developer has to supply `DB_USER` it's probably a minor convenience. So I'd be fine with this.

natematykiewicz:
I'd just like to point out that this GitHub Actions CI passed in just under 13 mins https://github.com/thoughtbot/shoulda-matchers/actions/runs/497565911, where the Travis CI run took over 3 hours https://travis-ci.org/github/thoughtbot/shoulda-matchers/builds/754356680.

🚀

bin/ci/install
mcmire: EDIT: Well, we do use Zeus for development, but it's more of an optional thing, and we don't make use of it for CI.

VSPPedro: Done!

VSPPedro:
What do you think about merging this PR with action/cache?

By doing that we can solve the problem with Travis and have a CI faster for now. I think this is kind of urgent.

Of course, we keep looking for a way to make it work with setup-ruby. I'm afraid it won't be easy to find the answer and it will probably take some time. We can create an issue to remember to do this.

natematykiewicz:
I think you should definitely merge as-is. This is a vast improvement.

mcmire (approved):
I agree about the urgency here and am good with this! Nice work :)

VSPPedro:
Thanks! Merging!

eregon:
Maybe there should be a separate action that only does the "bundle install and caching" part, and then you could use it multiple times in the same workflow? Feel free to open an issue at https://github.com/ruby/setup-ruby/issues
Another way here would be to simply run these different Gemfiles in different jobs, then there would be no need to install both Gemfiles in the same job.

natematykiewicz:
@eregon In a repo at my work, we needed to install a private gem. I had to configure the auth token to get it to install. It took me a while to figure out the correct ENV to set to add our GitHub Packages credentials.

Per the GitHub Packages documentation, you're supposed to run:

```
bundle config https://rubygems.pkg.github.com/OWNER USERNAME:TOKEN
```

https://docs.github.com/en/packages/guides/configuring-rubygems-for-use-with-github-packages#authenticating-to-github-packages

The only solution I found was to set BUNDLE_RUBYGEMS__PKG__GITHUB__COM, but that doesn't set my token for just this `OWNER`, it sets the token for all packages hosted on GitHub (a bit looser than I'd like, but the only other option was the manual actions cache).

The trouble is, setup-ruby doesn't (from what I could tell) give you a way to run `bundle config` commands in setup-ruby when using the bundler cache option. I can't run bundle config before the action, because I don't have Ruby installed yet, much less bundler. But I can't run it after because it's already failed to install our private gem by then.

I feel like it would be nice if there was an action that installed the expected Ruby and Bundler versions. And a separate action that bundle installs and caches the output. That would give me the option to `bundle config` between the two.

That seems to be what you're suggesting as well. You could even keep the bundler_cache option as a shortcut (and backwards compatibility).

natematykiewicz:
I made an issue about this, in regards to something at my work. https://github.com/ruby/setup-ruby/issues/155

PR #1398, opened 2021-01-14
Test against ruby-head

voxik:
There is issue running shoulda-matchers test suite against Ruby 3.0 [[1]]. To prevent issues like this, it would be beneficial to test ahead the official release.

[1]: https://bugs.ruby-lang.org/issues/17536


voxik:
Travis was replaced by GH actions => closing this for now

PR #1406, opened 2021-01-28
Handle argument delegation for Ruby 3

gaetanm:
Hi,

I'm trying to get rid of all the warnings I have when I launch Rspec on my project. Like this one:

> ruby/gems/2.7.0/gems/shoulda-matchers-4.5.1/lib/shoulda/matchers/rails_shim.rb:68: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

That double splat does fix it.

### DIFF
--- a/lib/shoulda/matchers/rails_shim.rb
+++ b/lib/shoulda/matchers/rails_shim.rb
@@ -65,7 +65,7 @@ module Shoulda
               request_params
             end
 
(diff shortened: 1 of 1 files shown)


VSPPedro (approved):

VSPPedro:
Thank you very much for your contribution!

PR #1417, opened 2021-02-17
Standardize error message of association matcher

VSPPedro:
I noticed that the association matcher was returning a slightly different message between the tests below:

https://github.com/thoughtbot/shoulda-matchers/blob/9763c17d27221c5d3946b0b297b0c17b8f5644f5/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb#L1451-L1454

https://github.com/thoughtbot/shoulda-matchers/blob/9763c17d27221c5d3946b0b297b0c17b8f5644f5/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb#L1436-L1439

Pay attention to the foreign_keys, their types are different. This PR standardized the error message  - in both cases the foreign_keys will be symbols. Besides that, these modifications fix one of the four tests that are failing with Rails 6.1.

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -1394,23 +1394,25 @@ module Shoulda
         end
 
         def class_has_foreign_key?(klass)
+          @missing = validate_foreign_key(klass)
+
+          @missing.nil?
+        end
+
+        def validate_foreign_key(klass)
           if options.key?(:foreign_key) && !foreign_key_correct?
-            @missing = foreign_key_failure_message(klass, options[:foreign_key])
-            false
-          elsif !has_column?(klass, foreign_key)
-            @missing = foreign_key_failure_message(klass, foreign_key)
-            false
-          else
-            true
+            foreign_key_failure_message(klass, options[:foreign_key])
+          elsif !has_column?(klass, actual_foreign_key)
+            foreign_key_failure_message(klass, actual_foreign_key)
           end
         end
 
         def has_column?(klass, column)
           case column
           when Array
-            column.all? { |c| has_column?(klass, c) }
+            column.all? { |c| has_column?(klass, c.to_s) }
           else
-            column_names_for(klass).include?(column)
+            column_names_for(klass).include?(column.to_s)
           end
         end
 
@@ -1442,19 +1444,13 @@ module Shoulda
           end
         end
 
-        def foreign_key
-          key = if foreign_key_reflection
-                  if foreign_key_reflection.respond_to?(:foreign_key)
-                    foreign_key_reflection.foreign_key
-                  else
-                    foreign_key_reflection.primary_key_name
-                  end
-                end
+        def actual_foreign_key
+          return unless foreign_key_reflection
 
-          if key.is_a?(Array)
-            key.map(&:to_s)
+          if foreign_key_reflection.respond_to?(:foreign_key)
+            foreign_key_reflection.foreign_key
           else
-            key.to_s
+            foreign_key_reflection.primary_key_name
           end
         end
 
--- a/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
@@ -1350,7 +1350,7 @@ Expected Parent to have a has_many association called children through conceptio
       end
 
       expected_message = 'Expected Person to have a has_one association called detail ' \
-        '(PersonDetail does not have a ["company_id", "person_detail_id"] foreign key.)'
+        '(PersonDetail does not have a [:company_id, :person_detail_id] foreign key.)'
 
       expect do
         have_one(:detail).class_name('PersonDetail').


lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -1394,17 +1394,23 @@ def touch_correct?
         end
 
         def class_has_foreign_key?(klass)
+          @missing = validate_foreing_key(klass)
+
+          @missing.nil?
+        end
+
+        def validate_foreing_key(klass)
```
mcmire, line 1402: `validate_foreign_key`?

lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -1443,18 +1449,12 @@ def primary_key_correct?(klass)
         end
 
         def foreign_key
```
mcmire, line 1451: Should we rename this `actual_foreign_key`? I'm reading through this code (AssociationMatcher has gotten really large 😞 ) and I'm wondering if it's going to be confusing when comparing this to `options[:foreign_key]`.

mcmire (commented):
I just had two comments. This error message is not the greatest anyways, IMO I don't know if it always points users to what the problem actually is or what the solution is, but that is not your fault and seems out of scope for your change here, so I'm okay with letting it slide. 

Out of curiosity, which test failed on Rails 6.1 and why?

VSPPedro:
I'm opening this PR, https://github.com/thoughtbot/shoulda-matchers/pull/1418, just to show the errors that are happening with Rails 6.1.

One of them is:

```
3) Shoulda::Matchers::ActiveRecord::AssociationMatcher have_one rejects an association with a valid :class_name and :foreign_key option (CPK), but no columns
     Failure/Error:
       expect do
         have_one(:detail).class_name('PersonDetail').
           with_foreign_key(%i[company_id person_detail_id])
       end.not_to match_against(Person.new).and_fail_with(expected_message)
     
       Expected the positive version of the matcher not to match and for the failure
       message to be:
     
       ------------------------------------------------------------------------
       Expected Person to have a has_one association called detail (PersonDetail does not have a ["company_id", "person_detail_id"] foreign key.)
       ------------------------------------------------------------------------
     
       However, it was:
     
       ------------------------------------------------------------------------
       Expected Person to have a has_one association called detail (PersonDetail does not have a [:company_id, :person_detail_id] foreign key.)
```

Before Rails 6.1 this lines was returning an array of symbols:

https://github.com/thoughtbot/shoulda-matchers/blob/9763c17d27221c5d3946b0b297b0c17b8f5644f5/lib/shoulda/matchers/active_record/association_matcher.rb#L1448

But now is returning a string. Example:  "[:company_id, :person_detail_id]". The logic on the association matcher checks if this is an array:

https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/association_matcher.rb#L1453-L1458

This logic was supposed to update that value to "["company_id", "person_detail_id"], but it's not working because it's not returning an array anymore.

I think this was the commit that made this impact: https://github.com/rails/rails/commit/c0750fe2304de2c7bae028cbc465b6bf7e7d253e

lib/shoulda/matchers/active_record/association_matcher.rb
VSPPedro: I agree. I've thought about splitting into smaller files, but I still have doubts. If you have a suggestion, I'm all ears.

lib/shoulda/matchers/active_record/association_matcher.rb
VSPPedro: I'm not sure about this change. It seems to me that the `?` Should be used in methods that either return true or false, but in this case, this method is returning an error message - when there is a problem.

Would you like the name to be changed?

VSPPedro: Done!

lib/shoulda/matchers/active_record/association_matcher.rb
VSPPedro: Done!

lib/shoulda/matchers/active_record/association_matcher.rb
mcmire: Haha yeah sorry I should have been more clear :)

lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -1394,17 +1394,23 @@ def touch_correct?
...
-            false
-          else
-            true
+            foreign_key_failure_message(klass, options[:foreign_key])
+          elsif !has_column?(klass, to_s(actual_foreign_key))
+            foreign_key_failure_message(klass, actual_foreign_key)
           end
         end
 
+        def to_s(value)
```
mcmire, line 1410: Since `to_s` is a built-in method in Ruby, maybe we should name this something else other than `to_s` so people don't get confused? Maybe `normalize_foreign_key`?

mcmire:
@VSPPedro As long as we don't change what the developer is passing to `with_foreign_key` when printing out the failure message, this makes sense to me. I say that because if the developer passes `["company_id", "person_detail_id"]` but the failure message seems to indicate they passed `[:company_id, :person_detail_id]` then they might get confused (or, it would give us confusing information if we attempted to help them). It doesn't seem like that that's what you're doing here, though, so I'm good with what you've done so far here.

lib/shoulda/matchers/active_record/association_matcher.rb
VSPPedro: No problem. I agree with you. Thank you for the suggestions. I will use the first one.

VSPPedro:
I'll go back to this as soon as I finished dropping rails 4.2, 5.0 and 5.1.

VSPPedro:
Done!

VSPPedro:
I decided to merge this PR into #1418. I think it will be simpler to work with just one branch.

PR #1420, opened 2021-02-18
fix: issues affecting code quality

withshubh:
#### Description
This PR fixes a few issues that were affecting the code quality.

#### Summary of changes

- use `delegate` for delegation
- Time should be used with zone


lib/shoulda/matchers/action_controller/flash_store.rb
```diff
@@ -36,7 +36,7 @@ def has_value?(expected_value)
         end
 
         def empty?
-          flash.empty?
+          flash&.empty?
```
mcmire, line 39: I'm curious why this is necessary. Are there issues you noticed when using the gem?

lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
```diff
@@ -113,7 +113,7 @@ def value
             case column_type
             when :integer, :float then 1
             when :decimal then BigDecimal(1, 0)
-            when :datetime, :time, :timestamp then Time.now
+            when :datetime, :time, :timestamp then Time.zone.now
```
mcmire, line 116: This makes sense, but again, are there issues you noticed when using the gem?

lib/shoulda/matchers/action_controller/flash_store.rb
withshubh: This is a case of Rails delegation. The `empty?` method is being delegated to the `flush` object. We can replace the `empty?` method by either `delegate :empty?, to: :flush` or we can add the safe navigation operator as the Rails' `allow_nil` option checks not just for nil but also delegates if nil responds to the delegated method.

lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
withshubh: I did not face any issues while using gem.

lib/shoulda/matchers/action_controller/flash_store.rb
mcmire: I see. So, `&.` is a handy operator, but you have to keep in mind that it's really a form of defensive coding. What you're saying when you're using it is, "I expect at some point for this value to be nil, so if that happens, do this." Sometimes defensive coding is perfectly fine because you might have a method you wrote and you know that there is a chance it could return nil (because you wrote it that way). In this case, and in the other files you've changed here, I don't think it makes sense to use defensive coding. We don't want `flash`, `attribute_setter`, etc. to be nil, and there's no reason we ought to expect it to be nil. If nil pops up, that's an indication of a larger problem that should be addressed at a higher level. Adding `&.`, in my opinion, sweeps that problem under the rug, when it should be blaring. Therefore, I'm not sure we need these `&.` changes.

lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
mcmire: Okay, fair enough. I'm surprised Rubocop didn't catch this, actually. What about `Time.current`? It's shorter :)

mcmire (commented):
Thanks for the PR. Just a bit of feedback: I would be careful characterizing these sorts of things as issues in the future, as an issue signifies a bug to me. These are good things to point out, but I would view them more as style / best practice alignments. I say that because these sorts of changes are a lot more subjective than, say, a fix for a bug that is clearly causing problems. Just wanted to add some background to my comments below. Anyway, thanks for bringing these up. If you could address my comments then we're good to go.

mcmire:
Hi @withshubh, just pinging you about the changes above!

lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
withshubh: Agreed with what you just said, since `Time.current` is just a shorter version of `Time.zone.now`, we can use that instead.

lib/shoulda/matchers/action_controller/flash_store.rb
withshubh: Makes sense. Since we do not expect the `flash` object to be nil ever, and using the `&.` safe navigation operator will hide any errors, we can instead use the other method suggested above to perform Rails delegations, as so: `delegate :empty?, to: :flash` instead.

mcmire: Yup, that's a good point.

withshubh:
Hi @mcmire :wave: 

any update on this?

lib/shoulda/matchers/action_controller/session_store.rb
```diff
@@ -20,7 +20,7 @@ def has_value?(expected_value)
         end
 
         def empty?
-          session.empty?
+          session&.empty?
```
mcmire, line 23: Per the argument I was making before, I don't think we don't need these `&.` checks, would you mind removing these?

mcmire (commented):
I have one comment below but we should be good beyond that. Please note that CI is currently failing to the mimemagic nonsense that recently popped up around Rails. We are working on a fix for this so this PR is blocked until that PR is merged, but after that we should be good.

lib/shoulda/matchers/action_controller/session_store.rb
withshubh: Sure :sparkles:

mcmire (approved):
Perfect, thank you for your contribution!

mcmire:
Would you mind rebasing this against master? Looks like we cleared up the issues with CI. Just want to make sure all of the tests pass.

lib/shoulda/matchers/action_controller/flash_store.rb
```diff
@@ -34,10 +34,8 @@ def has_value?(expected_value)
             expected_value === actual_value
           end
         end
-
-        def empty?
-          flash.empty?
-        end
+        
+        delegate :empty?, to: :flash
```
VSPPedro, line 38: I think the problem will not be solved after a rebase. 

This is the error at the moment:

```
undefined method `delegate' for Shoulda::Matchers::ActionController::FlashStore:Class (NoMethodError)
```

I solved this after adding this line at the beginning of the file:

```
require 'active_support/core_ext/module/delegation'
```

I'm not sure if this is the best approach. What do you think, @mcmire?

mcmire: Hmmm weird. I wonder why that's happening. Perhaps we don't use `delegate` anywhere else. Anyway, your fix makes sense to me.

lib/shoulda/matchers/action_controller/flash_store.rb
```diff
@@ -34,10 +34,8 @@ def has_value?(expected_value)
             expected_value === actual_value
           end
         end
-
-        def empty?
-          flash.empty?
-        end
+
```
mcmire, line 37: It looks like Rubocop is detecting that there is extra whitespace here. Maybe there is an extra space? Can you check this line? Sorry

lib/shoulda/matchers/action_controller/flash_store.rb
mcmire: @withshubh If you could add that require then I think that would fix CI.

mcmire:
@withshubh There we go, that did it. Thanks for this change and sorry it took so long to merge!

PR #1421, opened 2021-02-23
Add Foreign Key support

lifeiscontent:
@mcmire I couldn't find an API in rails to check if a foreign key exists, I'm guessing this is why you guys previously didn't support it.

If you've got some notes on how I could make the PR better, feel free to share, otherwise now that I know what I know, I'd also be happy to close this PR and issue.

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
@@ -106,6 +106,24 @@ describe Shoulda::Matchers::ActiveRecord::HaveDbColumnMatcher, type: :model do
     end
   end
 
+  context 'with foreign_key option' do
+    it 'accepts a column that is has a foreign_key' do
+      expect(with_table(:author_id, :integer, foreign_key: true)).
+        to have_db_column(:author_id).with_options(foreign_key: true)
+    end
+
+    it 'accepts a column that is has a foreign_key with options' do
+      expect(with_table(:author_id, :integer, foreign_key: { to_table: :users })).
+        to have_db_column(:author_id).with_options(foreign_key: { to_table: :users })
+    end
+
+    it 'raises an error with the unknown options' do
+      expect {
+        have_db_column(:author_id).with_options(foreign_keyy: true)
+      }.to raise_error('Unknown option(s): :foreign_keyy')
+    end
+  end
+
   def model(options = {})
     define_model(:employee, options).new
   end
--- a/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
@@ -84,7 +84,8 @@ module Shoulda
 
       # @private
       class HaveDbColumnMatcher
-        OPTIONS = %i(precision limit default null scale primary).freeze
+        OPTIONS = %i(precision limit default null scale primary foreign_key).
+          freeze
 
         def initialize(column)
           @column = column
@@ -133,6 +134,9 @@ module Shoulda
           end
           if @options.key?(:precision)
             desc << " of precision #{@options[:precision]}"
+          end
+          if @options.key?(:foreign_key)
+            desc << " of foreign_key #{@options[:foreign_key]}"
           end
           desc << " of limit #{@options[:limit]}" if @options.key?(:limit)
           desc << " of default #{@options[:default]}" if @options.key?(:default)


spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
```diff
@@ -106,6 +106,24 @@
...
+  context 'with foreign_key option' do
+    it 'accepts a column that is has a foreign_key' do
+      expect(with_table(:author_id, :integer, foreign_key: true)).
+        to have_db_column(:author_id).with_options(foreign_key: true)
+    end
+
+    it 'accepts a column that is has a foreign_key with options' do
+      expect(with_table(:author_id, :integer, foreign_key: { to_table: :users })).
+        to have_db_column(:author_id).with_options(foreign_key: { to_table: :users })
+    end
```
VSPPedro, line 118: Hi there, thanks for your contribution!

Would you mind adding a test when the foreign_key does not exist?

Just to check that this is also working.

lifeiscontent: @VSPPedro is that even possible? based on my research into rails it doesn't look like they provide an API to look at the foreign keys from the model that are in the schema.

mcmire: I think this may be possible to check, but not through the Column class. It looks like you might be able to get a list of all foreign keys through `ActiveRecord::Base.connection`: https://github.com/rails/rails/blob/v5.0.0/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L581.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
```diff
@@ -106,6 +106,24 @@
...
+
+    it 'raises an error with the unknown options' do
```
mcmire, line 120: I'm not sure you need to repeat this test here; this is already taken care of on line 110.

spec/unit/shoulda/matchers/active_record/have_db_column_matcher_spec.rb
lifeiscontent: @mcmire I tried that but in the test it just returns an empty array

mcmire: Ohh, hmm, I'm wondering if this is limited to Postgres and MySQL. The tests run using SQLite as a default but you can switch to Postgres by prepending the test command with `DATABASE_ADAPTER=postgresql`. Maybe see how you fare with that?

lifeiscontent: @mcmire does it make sense to add this feature considering it has that as an edge case?

VSPPedro: I'm afraid that the way it is will give some false positive.

Examples:

```
    it 'why isn\'t it failing?' do
      # The column is not a foreign_key.
      expect(with_table(:value, :integer, foreign_key: false)).
        to have_db_column(:value).with_options(foreign_key: true)
    end

    it 'why isn\'t it failing?' do
      # The column is not a foreign key.
      expect(with_table(:value, :integer, {})).
        to have_db_column(:value).with_options(foreign_key: true)
    end

    it 'why isn\'t it failing?' do
      # The column is not a foreign key.
      expect(with_table(:value, :integer, invalid_option: false)).
        to have_db_column(:value).with_options(foreign_key: true)
    end

    it 'why isn\'t it failing?' do
      # Using a string
      expect(with_table(:value, :string, this_is_not_a_foreing_key: false)).
        to have_db_column(:value).with_options(foreign_key: true)
    end
```

All of these tests pass, correct me if I'm wrong, they shouldn't, because they don't have the correct value - foreign_key = true. I will try to understand the problem a little more here to perhaps suggest a solution.

lifeiscontent: @VSPPedro that's fine, I'll close and remove foreign keys from my test upstream, thanks everyone for the amazing feedback and support on this, really appreciate it!

mcmire: Oh ok! Well I was going to say it's up to you since you requested this feature on whether `have_db_column` should support foreign keys, but if you find you do need this then don't hesitate to submit another PR and we'll try to help you some more.

PR #1422, opened 2021-03-02
Remove Rails 4.2 support

VSPPedro:
I think we don't need this comment:

https://github.com/thoughtbot/shoulda-matchers/blob/178df4970cd80d63b5e800bdf5d3a37bb43506ea/spec/support/unit/capture.rb#L1-L4

But I'm not sure if I should do anything other than remove it.


mcmire:
@VSPPedro I think the method is worth keeping, because we make use of it, but perhaps the comment should be updated so that instead of talking about something that will happen, it talks about something that already did happen?

VSPPedro:
> @VSPPedro I think the method is worth keeping, because we make use of it, but perhaps the comment should be updated so that instead of talking about something that will happen, it talks about something that already did happen?

Something like this?

https://github.com/thoughtbot/shoulda-matchers/blob/eb360106ef89dc2cad61ea090ca3326b2b34085a/spec/support/unit/capture.rb#L1-L3

mcmire:
@VSPPedro Yeah, that works for me.

VSPPedro:
@mcmire, don't you want to change anything else in this PR? Just to make sure if I can merge as it is. 😅

mcmire (approved):
@VSPPedro Yeah, I guess I wasn't sure how you were going about this, but it looks like your plan is _just_ to remove Rails 4.2 from CI in this PR and do cleanup around removing all of the places where we're checking for Rails 4.2 in other PRs, right? If that's the case then this looks good to me.

VSPPedro:
Yes, that is exactly what I'm thinking of doing. Thanks! Merging!

PR #1423, opened 2021-03-02
Refactor ActiveRecordVersions

VSPPedro:
This PR is a continuation of the `Remove Rails 4.2 support` - https://github.com/thoughtbot/shoulda-matchers/pull/1422.

I'll keep this as a draft meanwhile the other PR is not merged.

I realized that we don't need some active record validations in the tests. The purpose of this PR is to remove conditionals that are no longer needed due to the current version that we support.

What do you think?

PR #1426, opened 2021-03-07
Remove unnecessary conditions for rails 5.0

VSPPedro:
This PR is a continuation of the Refactor ActiveRecordVersions- #1423.

~I'll keep this as a draft meanwhile the other PR is not merged.~

### DIFF
--- a/lib/shoulda/matchers/rails_shim.rb
+++ b/lib/shoulda/matchers/rails_shim.rb
@@ -3,10 +3,6 @@ module Shoulda
     # @private
     module RailsShim # rubocop:disable Metrics/ModuleLength
       class << self
-        def action_pack_gte_5?
-          Gem::Requirement.new('>= 5').satisfied_by?(action_pack_version)
-        end
-
         def action_pack_lt_5?
           Gem::Requirement.new('< 5').satisfied_by?(action_pack_version)
         end
@@ -58,12 +54,7 @@ module Shoulda
         end
 
         def make_controller_request(context, verb, action, request_params)
-          params =
-            if action_pack_gte_5?
-              { params: request_params }
-            else
-              request_params
-            end
+          params = { params: request_params }
 
           context.__send__(verb, action, **params)
         end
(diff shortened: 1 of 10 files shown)


VSPPedro:
Hi, @mcmire, I hope you're doing fine.

When you have some free time, would you mind taking a look at this PR and sharing your thoughts on it, please?

I think we're almost finishing removing all the support for EOL'ed Ruby and Rails releases. :tada:

VSPPedro:
I find it another file that has some conditional that we don't need anymore. I'll remove these methods and add to this PR as soon as possible.

https://github.com/thoughtbot/shoulda-matchers/blob/e746f2e370259bac0f86e80f3c1456c2caf86805/lib/shoulda/matchers/rails_shim.rb#L6-L12

https://github.com/thoughtbot/shoulda-matchers/blob/e746f2e370259bac0f86e80f3c1456c2caf86805/lib/shoulda/matchers/rails_shim.rb#L20-L22

mcmire:
@VSPPedro Okay, sounds good!

VSPPedro:
I think that's it. I can't find any more conditionals that need to be removed for now. What do you think?

lib/shoulda/matchers/rails_shim.rb
```diff
@@ -58,12 +46,7 @@ def generate_validation_message(
         end
 
         def make_controller_request(context, verb, action, request_params)
```
mcmire, line 48: Now that there is no logic in `make_controller_request`, `tables_and_views`, and `validation_message_key_for_association_required_option`, what do you think about folding this code into the place where they are used so that we can remove these methods?

VSPPedro: I think this is a good idea! I'll do that asap. Thanks!

VSPPedro: Done!

mcmire (approved):
Very nice! Looks great!

VSPPedro:
Thanks!

PR #1428, opened 2021-03-07
Remove Rails 5.0 support

VSPPedro:
To be honest, I think there is nothing more to be done with this version. I'm saying this after looking at the Unit Tests::ActiveRecordVersions and UnitTests::RailsVersions classes.

This PR depends on the merge of the https://github.com/thoughtbot/shoulda-matchers/pull/1426.


mcmire (approved):
That seems right to me. It seems like there were a lot more checks around Rails 4.2 than there were for 5.0.

VSPPedro:
We are getting close to the next release 😄

PR #1430, opened 2021-03-20
Remove allow_mass_assignment_of matcher

VSPPedro:
I noticed that we are not testing this matcher with the Rails versions that we support. So I decided to remove it. What do you think?


mcmire:
That makes sense! I feel like everyone should be using Strong Parameters at this point in time.

mcmire (approved):

VSPPedro:
> I feel like everyone should be using Strong Parameters at this point in time.

I agree! :+1:

Thanks! 😄

PR #1431, opened 2021-03-21
Remove use_before_filter, use_after_filter and use_around_filter matchers

VSPPedro:
They are only tested on a version that we don't support anymore. What do you think about removing them?


mcmire (approved):
Yes, this totally makes sense! ✂️ ✂️

VSPPedro:
We are almost there! I'm facing a weird bug with ruby 3.0. I'll detail that later.

Thanks again!

Issue #1179, opened 2019-02-25
gemspec problem makes boilerplate files globally executable


MadBomber:
The bin directory only contains common boilerplate files.  The gemspec file those common boilerplate files in the bin directory globally executable.  Delete the line in the gemspec that makes everything in the bin directory globally executable.

Delete this line:
 s.executables = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }

mcmire:
Hey! Thanks for bringing this to our attention. You are 100% right and I am super embarrassed by this. I just opened #1180 to fix this. As soon as the tests pass I'll merge it and issue another release.

mcmire:
I've released 4.0.1 with a correct gemspec. I've also yanked 4.0.0 so that hopefully this doesn't happen for anyone else.

PR #1434, opened 2021-03-31
Update rails

VSPPedro:
We are facing some problems with CI with rails 6.0 or 5.2:

```
 Fetching gem metadata from https://rubygems.org/.........
Your bundle is locked to mimemagic (0.3.5), but that version could not be found
in any of the sources listed in your Gemfile. If you haven't changed sources,
that means the author of mimemagic (0.3.5) has removed it. You'll need to update
your bundle to a version other than mimemagic (0.3.5) that hasn't been removed
```

In case you want to know more about the problem: https://github.com/rails/rails/issues/41750

To fix that it's necessary to update to the most recent version.


mcmire (approved):
Yeah we just had to upgrade our Rails app at work 😅  This looks good to me.

VSPPedro:
I hope you haven't had too many problems with mimemagic.

mcmire:
@VSPPedro It was definitely a pain as it broke our CI for a couple of days and no one could merge code. I guess yanking a version of a gem is more destructive than I'd previously thought.

VSPPedro:
> I guess yanking a version of a gem is more destructive than I'd previously thought.

I never thought much about it. Have you ever wondered what would happen if a similar problem happened to ShouldaMatchers?

mcmire:
@VSPPedro This did happen, once: after I released 4.0.0, I learned there was a [bug that caused `bin/setup` to be installed globally](https://github.com/thoughtbot/shoulda-matchers/issues/1179). I immediately yanked it, fixed the issue, and released 4.0.1. I justified this because I thought if someone were to mistakenly download 4.0.0, I would have gotten a slew of comments from people reporting the same exact issue. I didn't even think about the fact that if someone had just upgraded to 4.0.0, after the yank it would have immediately broken their CI. So I guess I could have gotten angry comments either way. I think the difference here was that 4.0.0 wasn't around for a long time before I pulled it. It's definitely something I'm going to watch out for in the future though.

PR #1438, opened 2021-04-19
Remove repeated sentence

eugesma:
The titles Rails apps and Non-Rails apps were repeated

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,39 +67,6 @@ Now you need to configure the gem by telling it:
 * that you're using RSpec so that it can make those matchers available in
   your example groups
 
-#### Rails apps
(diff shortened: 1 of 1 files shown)


mcmire:
I can see how you might think that, but there are separate instructions for both RSpec and Minitest. The section headers are the same within these sections, but the instructions are slightly different.

VSPPedro:
Woops. I did not notice. I thought it was repeated too. I even wondered how I hadn't seen it before.

eugesma:
my bad!

mcmire:
No worries, thanks for reporting it anyway :)

PR #1440, opened 2021-05-05
Upgrade Rails

VSPPedro:
I noticed that a new version of rails has been released.

About the support for Rails 6.1, I intend to go back to work with that soon. In two weeks at most.


mcmire (approved):
Sounds good. Thanks for making this update!

PR #1441, opened 2021-06-02
Typo in validate_inclusion_of_matcher

kleinjm:


kleinjm:
I misread the docs. The plural version is for the model, not for the matcher.

PR #1447, opened 2021-06-21
Use flatten when matching columns on a join table

scharbau:


scharbau:
Sorry for opening this PR, wanted to test it in my private fork

PR #1450, opened 2021-07-02
Ensure uniqueness validation matcher works with STI

### LINKED ISSUE #854: Issue with Shoulda::Matchers::ActiveRecord::Uniqueness::TestModels
I've just upgraded aasm gem to latest version and now I have a failing shoulda matcher (version 2.8.0) spec. The code is pretty standard, I have a `Bookmark` model which belongs to a user and a bookmarkable (polymorphic) item. A given user can only bookmark items once thus I have this spec:

```ruby
it { should validate_uniqueness_of(:user_id).scoped_to(:bookmarkable_id, :bookmarkable_type) }
```

Running this spec with latest aasm code raises an error. I tracked the error to [this line](https://github.com/aasm/aasm/blob/64a133407d271d66462082622b67017f60a72374/lib/aasm/persistence/active_record_persistence.rb#L137) in aasm code. The issue is that the bookmarkable item (passed in matcher by shoulda code) does not retain its original class. 

In my spec setup I create a Bookmark using a factory and matcher code uses that factory to compute values for assertion. My factory returns an instance of `Place` for bookmarkable while in some way the class of bookmarkable used in matcher is
```
class Shoulda::Matchers::ActiveRecord::Uniqueness::TestModels::Placf < ActiveRecord::Base
```
That name is likely generated by [this code](https://github.com/thoughtbot/shoulda-matchers/blob/8253d4b4c9b5489ea003784e638e04db353f6250/lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb), however I don't know its purpose.

In any case I think that test models generated by your matcher should behave like original one, so they should return original class when `#class` is called on their instance.

### LINKED ISSUE #1245: Issue with validates uniqueness and STI/polymorphic models on rails 6?
This feels similar to this long-ago bug - https://github.com/thoughtbot/shoulda-matchers/issues/203 - which was fixed here - https://github.com/thoughtbot/shoulda-matchers/pull/592

Model of:

```
class Classification < ApplicationRecord
  belongs_to :classifiable, polymorphic: true
  belongs_to :topic

  validates :classifiable_id, :uniqueness => { :scope=> [:topic_id, :classifiable_type] }
end
```

Spec of:

```
require "rails_helper"

describe Classification do
  it { should belong_to(:classifiable) }
  it { should belong_to(:topic) }

  it "ensures uniqueness" do
    create(:classification)
    should validate_uniqueness_of(:classifiable_id).scoped_to([:topic_id, :classifiable_type])
  end
end
```

Relevant output is:

```
Failures:

  1) Classification ensures uniqueness
     Failure/Error: should validate_uniqueness_of(:classifiable_id).scoped_to([:topic_id, :classifiable_type])
     
     ActiveRecord::SubclassNotFound:
       Invalid single-table inheritance type: Show is not a subclass of Shoulda::Matchers::ActiveRecord::Uniqueness::TestModels::Producu
     # ./spec/models/classification_spec.rb:9:in `block (2 levels) in <top (required)>'
```

That `::Producu` on the very end there in the error makes me suspect that we're calling `next` on a string with `Product` (or something?) somewhere in the codebase.

StefSchenkelaars:
The uniqueness validation matcher generates a fake class when the
scope name ends on _type. This fake class is a duplicate of the
original class. In most cases this is not a problem, however when
using single table inheritance (STI), this new fake class is not
a subclass of the original class and therefore ActiveRecord will
throw an error. See #1245

A similar thing happens when the original class is using an AASM
state machine. The state machine settings are stored in a global
registry. So when the fake class is initialized, AASM cannot find
the settings of the set state machines (due to duplication) in the
registry. See #854

These problems are fixed by making the fake class inherrit from the
original class. Therefore it closes #1245 and closes #854


gemfiles/rails_5_2.gemfile
```diff
@@ -36,3 +36,4 @@ gem "listen", "~> 3.0.5"
 gem "spring-watcher-listen", "~> 2.0.0"
 gem "pg", "~> 0.18"
 gem "sqlite3", "~> 1.3.6"
+gem "aasm"
```
StefSchenkelaars, line 39: Not sure what the policy is about these kind of tests that require external libraries. I now only require this gem in the relevant spec but it's still needed in the gemfiles. I'm happy to change / remove it.

mahdiar-naufal-shyftplan (approved):

matsales28:
@StefSchenkelaars thanks for working on this, do you still want to work on that? If so I think we can just remove the part of this PR that introduces `aasm`, we shouldn't be testing specific behaviour of gems.

I can continue this work and give you credits if you're not interested anymore.

StefSchenkelaars:
Hi @matsales28 
It has been some time ago I worked on this. Feel free to take over the PR. If you don't have time, let me know what things you would like to see different (next to removing the dependency). I can try to free up some time the following weeks but I'm kinda booked atm. 
Good to see that it's picked up 😊
Stef

PR #1451, opened 2021-07-02
Remove last references to rails 5.1

StefSchenkelaars:
Don't think this needs any explanation


VSPPedro:
Whoops, my bad. Thanks!

VSPPedro (approved):

PR #1429, opened 2021-03-17
Remove rails 5.1 support

VSPPedro:
TODO list for Shoulda-matchers 5.0:

For ruby:

- [x] remove support for 2.5
- [x] add support for 3.0

For rails:

- [x] remove support for 4.2
- [x] remove support for 5.0
- [x] remove support for 5.1
- [ ] add support for 6.1

After this PR, I'll start adding the missing support. 😄 

~Before reviewing this PR is necessary to merge the #1428.~

PR #1453, opened 2021-07-09
Add `without_scopes` to enum matcher.

dewyze:
Rails 6 added the `_scopes: false` option when defining enums. This will
prevent rails from creating class scopes for each of the values.

Previously, shoulda-matchers checked for the presence of singleton
methods as a way to check that the corresponding enum methods have been
defined. In rails 6, if someone were using `_scopes: false` the matcher
will fail.

Instead, we should check on the presence of the instance methods that
are generated: `#{value}!` and `#{value}?` as the default checker for
the enum values generating methods.

We can then check on the scopes/singleton methods generated when using
the new `without_scopes` matcher.

### DIFF
--- a/lib/shoulda/matchers/active_record/define_enum_for_matcher.rb
+++ b/lib/shoulda/matchers/active_record/define_enum_for_matcher.rb
@@ -116,7 +116,7 @@ module Shoulda
       ## ##### with_prefix
       #
       # Use `with_prefix` to test that the enum is defined with a `_prefix`
-      # option (Rails 5 only). Can take either a boolean or a symbol:
+      # option (Rails 6+ only). Can take either a boolean or a symbol:
       #
       #     class Issue < ActiveRecord::Base
       #       enum status: [:open, :closed], _prefix: :old
@@ -163,6 +163,30 @@ module Shoulda
       #         with_suffix
       #     end
       #
+      # ##### without_scopes
+      #
+      # Use `without_scopes` to test that the enum is defined with
+      # '_scopes: false' option (Rails 5 only). Can take either a boolean or a
+      # symbol:
+      #
+      #     class Issue < ActiveRecord::Base
+      #       enum status: [:open, :closed], _scopes: false
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Issue, type: :model do
+      #       it do
+      #         should define_enum_for(:status).
+      #           without_scopes
+      #       end
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class ProcessTest < ActiveSupport::TestCase
+      #       should define_enum_for(:status).
+      #         without_scopes
+      #     end
+      #
       # @return [DefineEnumForMatcher]
       #
       def define_enum_for(attribute_name)
@@ -173,7 +197,7 @@ module Shoulda
       class DefineEnumForMatcher
         def initialize(attribute_name)
           @attribute_name = attribute_name
-          @options = { expected_enum_values: [] }
+          @options = { expected_enum_values: [], scopes: true }
         end
 
         def description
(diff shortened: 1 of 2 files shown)


dewyze:
I am also unsure how to structure this so it only tests for rails 6 on this test. I couldn't find examples of that (based on grepping the words `rails` or `6` in the unit tests.

VSPPedro:
Hi, @dewyze, I hope you are doing fine.

I think this PR can help you: https://github.com/thoughtbot/shoulda-matchers/pull/1429/files

As you can see there was a method to check if the rails version supported active storage before running the tests.

dewyze:
@VSPPedro tests are passing, and I added a test that was missing. I had a hard time figuring out the cleanest way to handle the messages. I am also not sure if there needs to be updates with the messages. Both the instance methods and class scopes can have prefixes or suffixes, so it did seem to make sense to include that messaging in both.

Additionally, I felt torn about the `exclude_scopes?` method. It felt weird to define a method in the negative. But in most cases I am checking for a negative, and it felt equally weird to have a `include_scopes` method but to have all uses be `!include_scopes`.

Nevertheless, best I can tell this includes all use cases and does work. Let me know if there is any cleanup you'd like to see.

lib/shoulda/matchers/active_record/define_enum_for_matcher.rb
```diff
@@ -163,6 +163,30 @@ module ActiveRecord
       #         with_suffix
       #     end
       #
+      # ##### without_scopes
+      #
+      # Use `without_scopes` to test that the enum is defined with
+      # '_scopes: false' option (Rails 5 only). Can take either a boolean or a
```
VSPPedro, line 169: Thanks again for your contribution.

Taking a second look at this PR, I realized that maybe this description isn't accurate. Shouldn't it be Rails 6 instead of Rails 5?

VSPPedro:
Hi, @dewyze, I hope you're doing fine.

In case you don't have the time to continue working on this PR, I can help.

What do you think?

dewyze:
@VSPPedro sorry about the long silence! I fixed the rubocop errors, but the 2.6/2.7 errors were too old to see. If you can run the workflows I can take a look at the errors and get them fixed up.

VSPPedro:
No Problem.

dewyze:
@VSPPedro the issue was that I was adding the scopes parameter in the builder, but not checking the rails version, so it would fail on rails 5 versions. I updated it.

spec/unit/shoulda/matchers/active_record/define_enum_for_matcher_spec.rb
```diff
@@ -862,7 +939,17 @@ def build_record_with_enum_attribute(
...
     end
 
-    model.enum(enum_name => values, _prefix: prefix, _suffix: suffix)
+    params = {
+      enum_name => values,
+      _prefix: prefix,
+      _suffix: suffix,
+    }
+
+    if rails_version =~ '~> 6.0'
```
mcmire, line 948: Hi @dewyze, I'm wondering if this should be `rails_version >= '6.0'` instead? I would imagine Rails 7 has the same change as Rails 6 did.

dewyze: @mcmire annoyingly, in rails 7, they removed the leading underscore. https://api.rubyonrails.org/classes/ActiveRecord/Enum.html

So this won't work for rails 7, but there are no appraisals yet, so i wasn't going to worry about that.

mcmire: Ah, no problem then!

VSPPedro (approved):

VSPPedro:
Thank you for your contribution!

VSPPedro:
What do you think about squashing the commits into one?

VSPPedro:
Thank you! ❤️

PR #1458, opened 2021-07-28
Install ActiveStorage migrations

VSPPedro:
Suggestion: instead of creating the table manually, use the command 'bundle exec rake active_storage: install: migrations'.

### DIFF
--- a/spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
@@ -1,35 +1,6 @@
 require 'unit_spec_helper'
 
 describe Shoulda::Matchers::ActiveRecord::HaveAttachedMatcher, type: :model do
-  before do
-    create_table :active_storage_blobs do |t|
-      t.string :key, null: false
-      t.string :filename, null: false
-      t.string :content_type
-      t.text :metadata
-      t.bigint :byte_size, null: false
-      t.string :checksum, null: false
-      t.datetime :created_at, null: false
-
-      t.index [:key], unique: true
-    end
-
-    create_table :active_storage_attachments do |t|
-      t.string :name, null: false
-      t.references :record, null: false, polymorphic: true, index: false
-      t.references :blob, null: false
-
-      t.datetime :created_at, null: false
-
-      t.index [:record_type, :record_id, :name, :blob_id],
-        name: 'index_active_storage_attachments_uniqueness', unique: true
-
-      # The original rails migration has a foreign key.
-      # Since this messes up the clearing of the database, it's removed here.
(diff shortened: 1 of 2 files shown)


spec/unit/shoulda/matchers/active_record/have_attached_matcher_spec.rb
```diff
@@ -1,35 +1,6 @@
...
-      t.references :blob, null: false
-
-      t.datetime :created_at, null: false
-
-      t.index [:record_type, :record_id, :name, :blob_id],
-        name: 'index_active_storage_attachments_uniqueness', unique: true
-
-      # The original rails migration has a foreign key.
-      # Since this messes up the clearing of the database, it's removed here.
-      # t.foreign_key :active_storage_blobs, column: :blob_id
```
mcmire, line 29: I think this comment is the reason why the tables were added manually when these tests were originally added, but I'm wondering if there is a way around this.

VSPPedro: Fixed, what do you think?

The problem was happening only when using Postgres, Rails 5.2 and running this command: `bundle exec rake db:drop:all db:create:all db:migrate`.

I'm not sure why, but the migration was only executed on the database production. That's why the test could not found the table. Splitting this command in two did the job:

```
run_command! 'bundle exec rake db:drop:all db:create:all'
run_command! 'bundle exec rake db:migrate' 
```

I'll keep digging to have a better understanding of what is happening. 😄

VSPPedro: It seems to me that this PR can wait until rails 5.2 has been end-of-lifed. I'll keep this as a draft and I'll keep digging because I think I can learn one thing or two about database connection. Let's see.

### More information about the problems - just a reminder for me.

I thought that the params received by this class would help me understand what was happening:

https://github.com/rails/rails/blob/48661542a2607d55f436438fe21001d262e61fec/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L46-L48

I noticed that when I run `rake db:migrate` separately the `conn_params` has this value:

`{:user=>"postgres", :password=>"postgres", :dbname=>"shoulda-matchers-test_development", :host=>"localhost"}`

We can see that the PostgreSQL connection will be with the shoulda-matchers-test_development database. The task works just fine, but when I tried to run all the tasks together:  `bundle exec rake db:drop:all db:create:all db:migrate`.

For some reason, that I still don't understand, the task db:migrate has this values on `con_parrams`:

`{:user=>"postgres", :password=>"postgres", :dbname=>"postgres", :host=>"localhost"}`

And that's why it's creating the active storage table in the wrong database.

I'll keep digging,

PR #1463, opened 2021-10-07
Use bundler v2 in development

leoarnold:
Prior to this commit, the `Gemfile.lock` enforced the use of
Bundler v1.17.3 which was released in 2014.

On Ruby 3, this triggers the following deprecation warning:

```
/home/user/.rvm/gems/ruby-3.0.2@shoulda-matchers/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
```

Since the gem itself in v5.0.0 requires at least Ruby v2.6,
it seems fair to require a recent version of Bundler v2,
which in turn runs of every Ruby v2.3 or later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'appraisal', '2.2.0'
-gem 'bundler', '~> 1.1'
+gem 'bundler', '~> 2.0'
 gem 'pry'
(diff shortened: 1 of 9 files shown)


VSPPedro:
LGTM! Thank you!

PR #1467, opened 2021-12-03
CI: Use bundler-cache

olleolleolle:
This PR uses the `bundler-cache` feature of the Ruby-maintained Action `ruby/setup-ruby`.

[See `ruby/setup-ruby` parameters declaration](https://github.com/ruby/setup-ruby/blob/master/action.yml) for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,9 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - 3.0.2
-          - 2.7.4
-          - 2.6.8
+          - '3.0'
+          - 2.7
+          - 2.6
         appraisal:
           - rails_6_1
           - rails_6_0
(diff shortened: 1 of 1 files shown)


olleolleolle:
This is a very small change, but which avoids a Float-to-String loss of characters.

An example (not from this project) of what it can look like before this change, in this picture:

bild

Read more details, if you like: https://github.com/actions/runner/issues/849

olleolleolle:
If this PR is merged, it also does what #1466 did (in another way).

VSPPedro:
Thank you for working on this!

I tried to use the ruby/setup-ruby, but it did not work as you can see in this PR: https://github.com/thoughtbot/shoulda-matchers/pull/1397.

If you like to understand why the CI is failing take a look at this comment: https://github.com/thoughtbot/shoulda-matchers/pull/1397#issuecomment-765059033

I intended to test these options, but I forgot: https://github.com/thoughtbot/shoulda-matchers/pull/1397#issuecomment-782710793

What do you think about giving it a try?

PR #1478, opened 2022-01-15
Skip bootsnap on the test project creation

neilvilela:
- Use the Rails --skip-bootsnap option instead of commenting it on boot.rb after project creation

I'm closing #1475 in favor of this PR

### DIFF
--- a/spec/support/acceptance/helpers/step_helpers.rb
+++ b/spec/support/acceptance/helpers/step_helpers.rb
@@ -92,7 +92,7 @@ module AcceptanceTests
     end
 
     def rails_new_command
-      if rails_version =~ '~> 6.0'
+      if rails_version >= 6.0
         "bundle exec rails new #{fs.project_directory} --database=#{database.adapter_name} --skip-bundle --skip-javascript --no-rc --skip-bootsnap"
       else
         "bundle exec rails new #{fs.project_directory} --database=#{database.adapter_name} --skip-bundle --no-rc --skip-bootsnap"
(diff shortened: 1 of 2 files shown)


neilvilela:
I believe I should've done the same on `spec/support/unit/rails_application.rb`. The `spec/support/acceptance/helpers/step_helpers.rb` versions seems to be more up-to-date, as no version lower than 5 is supported and the unit tests version will never reach the `else` branch.

I'm thinking about keeping the `spec/support/acceptance/helpers/step_helpers.rb` duplicated and then a future refactoring could be done to remove this duplication. What do you think?

neilvilela:
Besides this, the line `if rails_version =~ '~> 6.0'` is evaluated to `false`  on Rails 6.1 or by a future Rails 7.0 support. I've tried changing it to `if rails_version >= 6.0` and it runs as expected. I can add this to this PR or open another PR.

mcmire:
> I believe I should've done the same on `spec/support/unit/rails_application.rb`. The `spec/support/acceptance/helpers/step_helpers.rb` versions seems to be more up-to-date, as no version lower than 5 is supported and the unit tests version will never reach the `else` branch.
> 
> I'm thinking about keeping the `spec/support/acceptance/helpers/step_helpers.rb` duplicated and then a future refactoring could be done to remove this duplication. What do you think?

Hi @neilvilela — you're right that there is unnecessary duplication between these two files. I would say for now your plan is good — keep the duplication in place and make the change in both places. 

> Besides this, the line `if rails_version =~ '~> 6.0'` is evaluated to `false` on Rails 6.1 or by a future Rails 7.0 support. I've tried changing it to `if rails_version >= 6.0` and it runs as expected. I can add this to this PR or open another PR.

Yes, I think that's a good change to make as well. Doing that here is fine with me.

neilvilela:
Hi @mcmire, I've just pushed the mentioned changes.

I don't know how consistent the CI runs will be from now on, but this was the fastest build in 5 months :tada: (by ~1 minute, but still)

PR #1479, opened 2022-01-18
Add support for text columns for secure matchers

maxehmookau:
## Issue

Currently, using the `text` datatype for the `has_secure_token` functionality is supported in Rails, but when used alongside shoulda-matchers it fails as the datatype must be `string`.

On codebases that use postgresql, there's no difference between using `text` or `string` and in most cases it makes sense to use `text`.

## Proposed solution

Add support for `text` datatypes when using the `have_secure_token` matcher.

### DIFF
--- a/lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
@@ -110,7 +110,7 @@ module Shoulda
         end
 
         def has_expected_db_column?
-          matcher = HaveDbColumnMatcher.new(token_attribute).of_type(:string)
+          matcher = HaveDbColumnMatcher.new(token_attribute).of_type(:string) || HaveDbColumnMatcher.new(token_attribute).of_type(:text)
           matcher.matches?(@subject)
         end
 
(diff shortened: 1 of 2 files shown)


lib/shoulda/matchers/active_record/have_secure_token_matcher.rb
```diff
@@ -110,7 +110,7 @@ def has_expected_instance_methods?
         end
 
         def has_expected_db_column?
-          matcher = HaveDbColumnMatcher.new(token_attribute).of_type(:string)
+          matcher = HaveDbColumnMatcher.new(token_attribute).of_type(:string) || HaveDbColumnMatcher.new(token_attribute).of_type(:text)
```
mcmire, line 113: This is always going to return the first one, the one for `:string`. The reason is that this line merely returns a matcher object; it doesn't actually perform the matching. That happens in the next line. 

What you probably want to do is:

``` ruby
matchers = [
  HaveDbColumnMatcher.new(token_attribute).of_type(:string),
  HaveDbColumnMatcher.new(token_attribute).of_type(:text)
]
matchers.any? { |matcher| matcher.matches?(@subject) }
```

krainboltgreene:
Mind if I copy this PR with the requested changes so we can get this merged in?

matsales28:
Closing this one in favor of https://github.com/thoughtbot/shoulda-matchers/pull/1719

PR #1483, opened 2022-01-25
Update bin/setup to detect chruby installed from homebrew on M1

technicalpickles:
With Apple moving to M1, homebrew took the opportunity to change the
install location from /usr/local to /opt/homebrew

So check for /opt/homebrew in addition to /usr/local for a chruby.sh

I considered using `brew --prefix` for this, but I'm not sure we can assume it's present at this point in the script


mcmire:
This makes sense, thanks for this!

PR #1495, opened 2022-07-12
update version into read me

gabrielbaldao:
just putting the version in the readme equal to current version


mcmire:
Hi @gabrielbaldao, are you not receiving the latest version when installing `shoulda-matchers`? `~> 5.0` should pick up the latest version that exists between 5.x and 6.x, so there's no real need to be explicit about the version here.

PR #1499, opened 2022-08-04
feat: implement allow_blank qualifier to validate_presence_of

### LINKED ISSUE #1407: Add `allow_blank` for `validate_presence_of`
Hi, sounds odd but all validators support `:allow_nil` and `:allow_blank` and there are cases where the column is not null but empty strings are fine. 

Also when one has not null `hstore` or `jsonb` column (PostgreSQL) with a default value of `'{}'::jsonb`, one must set `:allow_blank => true` otherwise these saving will fail since `{}.present?` is `false`. 

Thoughts?

andreLumor:
Resolves #1407 

allow_blank qualifier for validate_presence_of is now available.
Usage example:
```rb 
it { is_expected.not_to validate_presence_of(:attribute).allow_blank }
```

Tests were implemented.

### DIFF
--- a/lib/shoulda/matchers/active_model/qualifiers.rb
+++ b/lib/shoulda/matchers/active_model/qualifiers.rb
@@ -9,5 +9,6 @@ module Shoulda
 end
 
 require_relative 'qualifiers/allow_nil'
+require_relative 'qualifiers/allow_blank'
 require_relative 'qualifiers/ignore_interference_by_writer'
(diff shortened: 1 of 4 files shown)


mcmire:
This looks good to me, but waiting until CI passes.

VSPPedro (approved):

mcmire:
Thank you so much @andreLumor!

PR #1502, opened 2022-08-21
Update CHANGELOG

VSPPedro:
I'm thinking of releasing the v5.2.0 I'm taking a closer look at all open PRs.

Is there anything you @mcmire would like to see in this upcoming release besides support for Rails 7?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## Unreleased
+
+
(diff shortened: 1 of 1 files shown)


mcmire:
Hi @VSPPedro sorry for not replying sooner. I haven't been keeping up with recent changes I will admit 😳 but this set of changes is good - thanks for staying on top of this 🙂

PR #1504, opened 2022-08-23
Using local gems should be enough for testing

voxik:
I think that `--local` should be always preferred and it is already used on other places. Which also leads me to think that the code abuses `run_command_within_bundle!` for purposes it was not intended to be used and `install_gems` should be used instead. But that does not used the `--binstubs` option OTOH. But that might be the preferred solution anyway. Not sure. I could have try that :thinking:

### DIFF
--- a/spec/support/acceptance/helpers/step_helpers.rb
+++ b/spec/support/acceptance/helpers/step_helpers.rb
@@ -116,7 +116,7 @@ module AcceptanceTests
 
     def add_rspec_rails_to_project!
       add_gem 'rspec-rails', rspec_rails_version
-      run_command_within_bundle!('bundle install --binstubs') if rails_gt_6_0?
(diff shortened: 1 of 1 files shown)


voxik:
> I could have try that

I not such Rails expert to judge if it has to be just for `rails_gt_6_0?` or not, so I'll leave the experiment to others :blush:

VSPPedro (approved):

mcmire (approved):
Never thought about this. Seems fine to me.

voxik:
> Never thought about this. Seems fine to me.

To provide a bit of background, we need this flag to build Fedora shoulda-matchers package, where the build system has not access to the internet. But it could also improve the test performance IMO, because it reuses what is already installed without reaching somewhere.

VSPPedro:
Thank you for your contribution! I never thought about that. Good catch!

PR #1505, opened 2022-09-02
Add support for ActiveModel::Error in Rails 7

henriquesml:
#### Problem
The purpose of this PR is to enable the use of shoulda-matchers with Rails 7. There is a problem when `format_validation_errors` execute `errors.keys`. Because, `errors.keys` don't existis in Rails 7. We can check that [here](https://api.rubyonrails.org/classes/ActiveModel/Errors.html).

#### Soluction
I added `errors.messages` for return a hash with errors and their values ​​so we can use `.keys` in the hash

#### Error print
image

#### Versions
Ruby: 3.1.2
Rails: 7.0.3.1

### DIFF
--- a/lib/shoulda/matchers/active_model/helpers.rb
+++ b/lib/shoulda/matchers/active_model/helpers.rb
@@ -8,7 +8,7 @@ module Shoulda
         end
 
         def format_validation_errors(errors)
-          list_items = errors.to_hash.keys.map do |attribute|
+          list_items = errors.messages.keys.map do |attribute|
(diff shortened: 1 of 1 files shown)


mcmire:
Hi @henriquesml , it looks like you might be using version 4.5.1 of the gem, because the snippet of code in error you posted doesn't look the same as in the current version (your screenshot shows `errors.keys.map` but the latest version has `errors.to_hash.keys.map`). So this problem may already be fixed -- what happens if you upgrade to the latest version of this gem?

PR #1506, opened 2022-09-05
Add Rails 7

VSPPedro:
Continuing #1485


VSPPedro:
@mcmire, what do you think?

I'm afraid that squash and merging this PR would make no contribution to @technicalpickles, but I could be wrong. I'm thinking about this because I've never continued a PR from another user. Can you tell me if there is a better approach, please?

@technicalpickles, if you have a few minutes you could update your PR with the changes I added here. I think that would solve my question and I would close this PR.

VSPPedro:
@technicalpicles yes it did what I feared. I'm sorry for that. I really thought that by using the squash and merge button, the contribution would be yours too. Sorry about that.

mcmire:
@VSPPedro I meant to reply, sorry! In the future when you edit the commit message that Git creates for you when you press that squash and merge button, if you add a "Co-authored-by" line at the end it will allow you to attribute the commit to someone else.

VSPPedro:
No problems. Thanks for sharing this with me. It won't happen again.

mathieujobin:
FYI - @mcmire @VSPPedro , Actually Github automatically adds the `Co-Authored-by` near the end of the commit description. All that was needed to do is not clear the description.

VSPPedro:
I see, my mistake. I did not know that. I cleared the description without looking at it.

Issue #1493, opened 2022-05-14
Missing Support for the numericality `in:` Range Validation


nbibler:
There doesn't appear to be direct support for [ActiveRecord's "new" numericality `in:` range](https://guides.rubyonrails.org/active_record_validations.html#numericality) check:

```ruby
class MyRecord < ApplicationRecord
  validates :progress, numericality: { in: 0..100 }
end
```

Attempting to use shoulda-matcher's `validate_numericality_of` seems to only expose upper and lower bound checks:

```ruby
it {
  is_expected.to(
    validate_numericality_of(:progress)
      .is_greater_than_or_equal_to(0)
      .is_less_than_or_equal_to(100))
}
```

One might think doing something like the above would work, but it doesn't because the new failure message is `["must be in 0..100"]` and that doesn't match the built-in message expectations for the greater than and less than matchers.

There is a similarly unexpected failure if one uses the `validate_inclusion_of` matcher, instead:

```ruby
it { is_expected.to(validate_inclusion_of(:progress).in_range(0..100) }
```

Again, the new failure message (cited above) doesn't match the matcher's expected "is not included in the list" message.

Currently, the workaround seems to be to use the `validate_inclusion_of` to get Range support, but with a custom message matcher:

```ruby
it { is_expected.to validate_inclusion_of(:progress).in_range(0..100).with_message('must be in 0..100') }
```

mcmire:
Thanks for the report! Yes, we don't currently handle this right now. PR welcome :)

andreLumor:
Hi, I'm interested in start contributing to your project.
Would you know if someone is working on this issue?

nbibler:
As far as I'm aware, no one on our side (at @envylabs) is working on it, yet.

andreLumor:
Hi!
I noticed that the "numericality in: range" check was added on rails 7.
I don't know if resolving this issue is currently possible since shoulda-matchers still don't have rails 7 support.
Any thoughts on how to solve it?
Thanks.

VSPPedro:
Hi, @andreLumor, unfortunately, we are not supporting rails 7 at the moment. PR Welcome.

For now, I suggest creating a custom matcher and as soon as the support it's added you could create a PR with your custom matcher.

andreLumor:
Hi, @VSPPedro.
Thanks for the reply! I think that starting contributing to an issue that complex would be inefficient, but I'm interested in working on it sometime soon.
[...]

PR #1512, opened 2022-10-01
Add `in: range` matcher to validate_numericality_of

matsales28:
Closes: #1493

In Rails 7 was added a new option to validate numericality. You can use `in: range` to specify a range to validate an attribute.

```ruby
# Before Rails 7
class User < ApplicationRecord
  validates :age, numericality: { greater_than_or_equal_to: 18, less_than_or_equal_to: 65 }
end

# After Rails 7
class User < ApplicationRecord
  validates :age, numericality: { in: 18..65 }
end
```

In this commit we are adding the support matcher to this new functionality, while also making a refactor on the numericality matchers that use the concept of submatchers.

We've created a new class (`NumericalityMatchers::Submatcher`) that's been used by `NumericalityMatchers::RangeMatcher` and `NumericalityMatchers::ComparisonMatcher`, this new class wil handle shared logic regarding having submatchers that will check if the parent matcher is valid or not.

Our new class `Numericality::Matchers::RangeMatcher` is using as submatchers two `NumericalityMatchers::ComparisonMatcher` instances to avoid creating new logic to handle this new option and also to replicate what was being used before this option existed in Rails (see example above)

In this commit we are adding:

* NumericalityMatchers::RangeMatcher file to support the new `in: range` option.
* Specs on ValidateNumericalityOfMatcherSpec file for the new supported option, only running on rails_versions > 7.
* NumericalityMatchers::Submatchers file to handle having submatchers inside a matcher file.
* Refactors to NumericalityMatchers::ComparisonMatcher.

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
@@ -69,26 +69,34 @@ describe Shoulda::Matchers::ActiveModel::ValidateNumericalityOfMatcher, type: :m
           validation_name: :on,
           validation_value: :customizable,
         },
+        {
+          category: :range,
+          name: :is_in,
+          argument: 1..10,
+          validation_name: :in,
+          validation_value: 1..10,
+          rails_version: 7.0,
+        },
       ]
     end
 
     def qualifiers_under(category)
-      all_qualifiers.select do |qualifier|
+      all_available_qualifiers.select do |qualifier|
         qualifier[:category] == category
       end
     end
 
     def mutually_exclusive_qualifiers
-      qualifiers_under(:cardinality) + qualifiers_under(:comparison)
+      qualifiers_under(:cardinality) + qualifiers_under(:comparison) + qualifiers_under(:range)
     end
 
     def non_mutually_exclusive_qualifiers
-      all_qualifiers - mutually_exclusive_qualifiers
+      all_available_qualifiers - mutually_exclusive_qualifiers
     end
 
-    def validations_by_qualifier
-      all_qualifiers.each_with_object({}) do |qualifier, hash|
-        hash[qualifier[:name]] = qualifier[:validation_name]
+    def all_available_qualifiers
+      all_qualifiers.filter do |qualifier|
+        rails_version >= qualifier.fetch(:rails_version, rails_oldest_version_supported)
       end
     end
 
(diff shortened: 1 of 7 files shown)


spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -69,26 +69,34 @@ def all_qualifiers # rubocop:disable Metrics/MethodLength
...
     end
 
-    def validations_by_qualifier
-      all_qualifiers.each_with_object({}) do |qualifier, hash|
-        hash[qualifier[:name]] = qualifier[:validation_name]
+    def all_available_qualifiers
+      all_qualifiers.filter do |qualifier|
+        rails_version >= qualifier.fetch(:rails_version, rails_oldest_version_supported)
       end
     end
```
matsales28, line 101: This method was not being used at all, I've deleted it, but happy to revert if it's needed in some way or not in the scope of this PR to make this change.

spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb
```diff
@@ -2080,7 +2226,11 @@ def apply_qualifiers!(args)
 
     combination.each do |qualifier|
       args = self.class.default_qualifier_arguments.fetch(qualifier[:name])
-      matcher.__send__(qualifier[:name], *args)
+      if args
+        matcher.__send__(qualifier[:name], args)
+      else
+        matcher.__send__(qualifier[:name])
+      end
```
matsales28, line 2233: This method was causing an error because `args` when testing the `in` option was a range, and the splat operator was destructuring the range.

We were using the splat operator on `args` only to prevent passing it when it was nil, so I changed this method to reflect that better and not cause the error when using a range.

Happy to change to this other implementation if you think the use case of not destructuring is more clear when the argument is a range.
```suggestion
      if args.is_a? Range
        matcher.__send__(qualifier[:name], args)
      else
        matcher.__send__(qualifier[:name], *args)
      end
```

matsales28:
Happy to contribute to a gem that I used a lot. It was pretty challenging but a fun journey to make this PR because I didn't know the codebase at all.

I've tried to make a minor refactoring. If y'all think that's not in the scope of this PR, I'm happy to revert the changes and limit my PR only to add the new functionality, and any suggestions to make the code clearer / better are welcome! 😄

VSPPedro (approved):
Nice PR! I'm fine with all your changes, I haven't merged it yet because I'm wondering if @mcmire would like to take a look too.

Thanks!

mcmire:
Hey! I skimmed over the PR. My initial thought is that I feel like I've done something similar in years past, where I've created an abstraction around a _group_ of matchers rather than just one, but this particular implementation is really straightforward, so I like it. I'll give this a deeper dive today.

mcmire:
I'm on vacation right now but honestly I'm fine with merging this PR as well. If there's duplication we can always refactor it out later, it's no big deal.

rdsngit:
> I'm on vacation right now but honestly I'm fine with merging this PR as well. If there's duplication we can always refactor it out later, it's no big deal.

@mcmire thanks for the update, hope you enjoy the rest of your vacation :)

PR #1518, opened 2022-10-31
Support :uuid column type for validate_absence_of matcher

dlupu:
Hello, 

I run into a case where I needed to make sure that an `uuid` column was empty but could not use `validate_absence_of` matcher. 

It tried to set the column with "an arbitrary value" which is not a valid uuid value.

This Pull Requests attempts to fix the situation.

### DIFF
--- a/spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb
@@ -14,7 +14,7 @@ describe Shoulda::Matchers::ActiveModel::ValidateAbsenceOfMatcher, type: :model 
         :time,
         :date,
         :binary,
-        :uuid
+        :uuid,
       ]
     end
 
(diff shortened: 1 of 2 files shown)


VSPPedro (dismissed):

VSPPedro:
Hi, @dlupu, thank you very much for your contribution!

There is only one detail: rubocop detected trailing whitespace. Would be possible to fix that, please?

About the error with 3.1.2, rails_6_0 and PostgreSQL I'll take a closer look as soon as I have some time.

dlupu:
@VSPPedro sorry for the late reply.

I removed the rubocop issue. 

However, I'm surprised about the error that appears only on `Test / build (3.1.2, rails_6_0, postgresql)`. I've cloned the code locally and run the failing test on my machine using
* ruby 3.1.2
* appraisal rails_6_0
* postgresql db adapter

The tests passes ....

spec/unit/shoulda/matchers/active_model/validate_absence_of_matcher_spec.rb
```diff
@@ -14,6 +14,7 @@ def self.available_column_types
         :time,
         :date,
         :binary,
+        :uuid
```
dlupu, line 17:
```suggestion
        :uuid,
```

dlupu:
@VSPPedro  it seems that `Test / build (3.1.2, rails_6_0, postgresql)` is green not. I also fixed the last rubocop issue, please trigger the CI again.

VSPPedro:
Hi @dlupu, thanks for working on this. The failure we had earlier looks like an intermittent test. It was probably not caused by your changes. I'll keep an eye on that. Thank you again!

VSPPedro (approved):

PR #1519, opened 2022-11-09
fix example with wrong attribute name

franlocus:
Hello :)
I was reading the example and found an error with the names of the attributes being tested.

Greetings!


VSPPedro (approved):

VSPPedro:
Nice! Thank you for your contribution!

Issue #1433, opened 2021-03-24
Confusing error message from inclusion_of


runephilosof-abtion:
In the below case, the odd error message appears, when the subject is a new record with the id set to something.

```
class Low < ApplicationRecord
  has_many :issues
end
class Medium < ApplicationRecord
  has_many :issues
end
class High < ApplicationRecord
  has_many :issues
end
class Issue < ApplicationRecord
  belongs_to :severity, polymorphic: true

  validates_inclusion_of :severity_type,
    in: %w(Low Medium High)
end
RSpec.describe Issue, type: :model do
  subject { Issue.new(severity_id: 1) } # if this is commented out the test is succesful.
  it do
    should validate_inclusion_of(:severity).
      in_array(%w(Low Medium High)).
  end
end
```
Results in
```
     Failure/Error:
       should validate_inclusion_of(:severity_type).in_array(%w[Low Medium High])
     
     NameError:
       wrong constant name shoulda-matchers test string
```

mcmire:
Hi @runephilosof-abtion,

`severity_type` looks to be a "magic" polymorphic column. Therefore, I believe you are getting this error because ActiveRecord (and we) are expecting `severity_type` to be a reference to a model. What are the possible models that `severity` could refer to? Perhaps you want to use that set in the array rather than a set of severity levels?

runephilosof-abtion:
@mcmire  That is exactly what I meant. I have updated the first comment to reflect that.
I want to test that the type is only set to the class names in the validation.

AlfonsoUceda:
Hey folks, after updating to rails 7, this error happens in my test suite, is there a workaround to fix this?

VSPPedro:
Unfortunately, I'm working on adding support for Rails 7 at the moment. Thank you for sharing this problem with us, I'll take a look ASAP.

yboulkaid:
We're also seeing the same error in conjunction with the Rails 7 upgrade. I believe that this is because Rails is trying to `constantize` the passed string to check if it's a valid model, but is running into a invalid name for a constant here:
https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L272

Which explains the cryptic error message.

I'm not sure how to get around that, looks like it's a new behavior of rails to try to `constantize` this so early?

technicalpickles:
[...]

PR #1523, opened 2022-11-24
fix: confusing error message from validate_inclusion_of matcher

VSPPedro:
All the credits go to [christhomson](https://github.com/christhomson). I can't thank you enough for pointing me in the right direction.

Close https://github.com/thoughtbot/shoulda-matchers/issues/1433

What do you think, @mcmire?

About the issue: https://github.com/rails/rails/pull/42124. Rails now throws an error when the class doesn't exist. To fix this it was necessary to update 'shoulda-matchers test string' to a class name as you can see in this PR.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -3,6 +3,8 @@ require 'date'
 
 module Shoulda
   module Matchers
+    class ExampleClass; end
+
     module ActiveModel
       # The `validate_inclusion_of` matcher tests usage of the
       # `validates_inclusion_of` validation, asserting that an attribute can
@@ -269,7 +271,7 @@ module Shoulda
       # @private
       class ValidateInclusionOfMatcher < ValidationMatcher
         BLANK_VALUES = ['', ' ', "\n", "\r", "\t", "\f"].freeze
-        ARBITRARY_OUTSIDE_STRING = 'shoulda-matchers test string'.freeze
+        ARBITRARY_OUTSIDE_STRING = Shoulda::Matchers::ExampleClass.name
         ARBITRARY_OUTSIDE_INTEGER = 123456789
         ARBITRARY_OUTSIDE_DECIMAL = BigDecimal('0.123456789')
         ARBITRARY_OUTSIDE_DATE = Date.jd(9999999)
--- a/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
@@ -886,6 +886,41 @@ describe Shoulda::Matchers::ActiveModel::ValidateInclusionOfMatcher, type: :mode
     def validation_matcher_scenario_args
       super.deep_merge(model_creator: :active_record)
     end
+
+    context 'against a polymorphic association' do
+      it 'matches when the subject configures validate_inclusion_of' do
+        define_model(:issue, severity_id: :integer, severity_type: :string) do
+          belongs_to :severity, polymorphic: true
+          validates_inclusion_of :severity_type, in: %w(Low Medium High)
+        end
+        define_model(:high)
+        define_model(:medium)
+        define_model(:low)
+
+        expect { validate_inclusion_of(:severity_type).in_array(%w(Low Medium High)) }.to match_against(Issue.new)
+      end
+
+      it 'does not match when subject does not set validate_inclusion_of' do
+        define_model(:issue, severity_id: :integer, severity_type: :string) do
+          belongs_to :severity, polymorphic: true
+        end
+        define_model(:high)
+        define_model(:medium)
+        define_model(:low)
+
+        expected_message = <<-MESSAGE.strip
+Expected Issue to validate that :severity_type is either ‹"Low"›,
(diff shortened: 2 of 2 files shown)


spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```diff
@@ -886,6 +886,55 @@ def define_simple_model(
...
+    #   validates_inclusion_of :severity_type,
+    #     in: %w(Low Medium High)
+    # end
+    # RSpec.describe Issue, type: :model do
+    #   subject { Issue.new(severity_id: 1) } # if this is commented out the test is succesful.
+    #   it do
+    #     should validate_inclusion_of(:severity).
+    #       in_array(%w(Low Medium High)).
+    #   end
+    # end
```
VSPPedro, line 912: Oops, I'll remove this later

VSPPedro:
The failed test is intermittent. I'll work on it as soon as I finish this and release a new version.

spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```diff
@@ -886,6 +886,31 @@ def define_simple_model(
...
+
+      it 'does not match when subject does not set valid_inclusion_of' do
+        define_model(:issue, severity_id: :integer, severity_type: :string) do
+          belongs_to :severity, polymorphic: true
+        end
+        define_model(:high)
+        define_model(:medium)
+        define_model(:low)
+
+        expect { validate_inclusion_of(:severity_type).in_array(%w(Low Medium High)) }.not_to match_against(Issue.new)
```
VSPPedro, line 911: I intend to test the error message on this test.

VSPPedro: The error message at the moment: 
```
"After setting :severity_type to ‹\"Shoulda::Matchers::ActiveModel::ValidateInclusionOfMatcher::ArbitraryShouldaMatchersTestString\"›, the matcher expected the Issue to be invalid, but it was valid instead."
```

It seems too big for me. I'm working on it.

spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
```diff
@@ -886,6 +886,40 @@ def define_simple_model(
...
+          validates_inclusion_of :severity_type, in: %w(Low Medium High)
+        end
+        define_model(:high)
+        define_model(:medium)
+        define_model(:low)
+
+        expect { validate_inclusion_of(:severity_type).in_array(%w(Low Medium High)) }.to match_against(Issue.new)
+      end
+
+      it 'does not match when subject does not set valid_inclusion_of' do
```
christhomson, line 903:
```suggestion
      it 'does not match when subject does not set validate_inclusion_of' do
```

christhomson (commented):
Thanks for implementing the fix!

lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
```diff
@@ -269,7 +269,7 @@ def validate_inclusion_of(attr)
       # @private
       class ValidateInclusionOfMatcher < ValidationMatcher
         BLANK_VALUES = ['', ' ', "\n", "\r", "\t", "\f"].freeze
-        ARBITRARY_OUTSIDE_STRING = 'shoulda-matchers test string'.freeze
+        ARBITRARY_OUTSIDE_STRING = Shoulda::Matchers.name
```
mcmire, line 272: I am not sure whether we should use `Shoulda::Matchers.name` here. I feel like this would be confusing in the message that appears when the matcher fails. Also, I am not sure that we should use this for _all_ uses of the matcher.

What do we feel about making a new variable, `ARBITRARY_OUTSIDE_CLASS_NAME`, creating a new class that lives inside of the Shoulda::Matchers namespace (maybe call it `ExampleClass`?) and then using the name of that class for the value?

mcmire (commented):
Hey @VSPPedro sorry it took me so long to look at this, I just had one comment.

lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
VSPPedro: I think it's a great idea. I will work on it this weekend. I think I will have some free time to release the next version. Thanks for the review!

I tried to add some logic where the matcher would only use a class if the column was being used by a polymorphic association, but I ran into some issues. I decided to follow this path because it was much simpler. I can take another look later and try again. Maybe in another PR if you don't mind.

mcmire: Oh alright. No worries then, that sounds good.

spec/unit/shoulda/matchers/active_model/validate_inclusion_of_matcher_spec.rb
VSPPedro: Thank you!

lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
VSPPedro: Thank you!

PR #1542, opened 2023-03-05
Improve `have_db_index` to better handle columns with multiple indexes

abrom:
The `have_db_index` matcher currently matches on the first index that has the matching column names. But when used with the `unique` qualifier that *can* result in the "wrong" index being matched causing the test to fail. 

For example, if a column had multiple indexes.. one unique BTree index, and another non-unique partial or GIN index. Using the `unique` qualifier fail because the non-unique index was identified as the `matched_index`.

It would appear that the order returned by `actual_indexes` will be database specific (say by index name) or possibly pseudo-random??

This change applies a "pre sort" of the indexes prior to finding the matching one if the unique qualifier is specified to give the match the best chance of being the "correct" one.

Note that this does of course complicate things if additional qualifiers were to be added because then sort order precedence would need to be considered.

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
@@ -208,18 +208,17 @@ module Shoulda
         end
 
         def sorted_indexes
-          indexes = actual_indexes
-
           if qualifiers.include?(:unique)
-            indexes.sort_by! do |index|
+            # return indexes with unique matching the qualifier first
+            unsorted_indexes.sort_by do |index|
               index.unique == qualifiers[:unique] ? 0 : 1
             end
+          else
+            unsorted_indexes
           end
-
-          indexes
         end
 
-        def actual_indexes
+        def unsorted_indexes
           model.connection.indexes(table_name)
         end
 
--- a/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_index_matcher.rb
@@ -197,14 +197,26 @@ module Shoulda
         def matched_index
           @_matched_index ||=
             if expected_columns.one?
-              actual_indexes.detect do |index|
+              sorted_indexes.detect do |index|
                 Array.wrap(index.columns) == expected_columns
               end
             else
-              actual_indexes.detect do |index|
+              sorted_indexes.detect do |index|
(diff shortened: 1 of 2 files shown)


lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -197,16 +197,28 @@ def correct_unique?
...
+        def sorted_indexes
+          indexes = actual_indexes
+
+          if qualifiers.include?(:unique)
+            indexes.sort_by! do |index|
+              index.unique == qualifiers[:unique] ? 0 : 1
+            end
+          end
+
+          indexes
```
mcmire, line 219: The bang at the end of `sort_by` modifies the array in place. In-place mutation tends to be useful for performant or complex code, however, given how simple this method is, I wonder if we say things a bit more straightforwardly. What do you think about this?

``` suggestion
          if qualifiers.include?(:unique)
            actual_indexes.sort_by do |index|
              index.unique == qualifiers[:unique] ? 0 : 1
            end
          else
            actual_indexes
          end
```

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
```diff
@@ -197,16 +197,28 @@ def correct_unique?
...
+        end
+
         def actual_indexes
```
mcmire, line 222: Since we now have a new method called `sorted_indexes`, what do you think about renaming this to `unsorted_indexes`?

mcmire (commented):
Thanks for the PR! I like how simple it is. I just had two comments.

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
abrom: Sure. I'd done it using the `sort_by!` in case there were any new qualifiers added to support applying other sorting to the list. But it can of course be refactored if/when that's the case! Will update.

lib/shoulda/matchers/active_record/have_db_index_matcher.rb
abrom: yeah, sure. Sounds good to me

mcmire (approved):

mcmire:
Looks great! Thanks so much!

PR #1544, opened 2023-03-09
Add timestamptz to the dummy value case statement

callahat:
Ran into an exception when upgrading to Rails 7.0.4.2 and a column of type `timestamptz`.

```
     Failure/Error:
       is_expected.to validate_uniqueness_of(:amount).
         scoped_to(:column1, :column2, :column3, :date_time_column).
         with_message('must be unique')
     
     NoMethodError:
       undefined method `getutc' for nil:NilClass
```

After tracing, 'dummy value' was observed as being sent into this function, when then resulted in `time` being nil:
https://github.com/rails/rails/blob/v7.0.4.2/activerecord/lib/active_record/connection_adapters/postgresql/oid/timestamp_with_time_zone.rb#L122

### DIFF
--- a/lib/shoulda/matchers/util.rb
+++ b/lib/shoulda/matchers/util.rb
@@ -94,7 +94,7 @@ module Shoulda
             0
           when :date
             Date.new(2100, 1, 1)
-          when :datetime, :timestamp
+          when :datetime, :timestamp, :timestamptz
             DateTime.new(2100, 1, 1)
           when :time
             Time.new(2000, 1, 1)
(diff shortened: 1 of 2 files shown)


mcmire:
Hi @callahat, thanks for the PR. Would you mind adding tests for this so that the bug doesn't reappear later accidentally? You can see an example of existing tests here: https://github.com/thoughtbot/shoulda-matchers/blob/195153cf42239935062f10f29a148d9d40032ca2/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb#L832. Thanks!

callahat:
Tests added for `timestamptz` as well as `timestamp` (the later already existed, but didn't have a similar test to `datetime`, and all three are in the same `when` clause)

mcmire:
@callahat Great, thanks so much!

Issue #1302, opened 2020-05-15
should_not validate_presence_of can give a false positive


mcmire:
If you have a situation where you have a presence validation with a custom message and you say `it { should_not validate_presence_of(:attribute) }` when in fact the validation is being run, the test will pass. This is technically true, because `it { should_not validate_presence_of(:attribute) }` is the same thing as saying `it { should_not validate_presence_of(:attribute).with_message(:blank) }`. That may make sense for `should`, but for `should_not` this doesn't make any sense.

HeitorMC:
Hey @mcmire. I'm working on this one. But honestly, I'm getting lost searching for the responsible method of the false positive, in `lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb`.

Can you help me out? :smile:

PR #1545, opened 2023-03-10
Fix: false positive of validate_presence_of

HeitorMC:
Referent to #1302 

When your model has `validate_presence_of` with a custom message:
```rb
class Article < ApplicationRecord
  validates_presence_of :owner, message: 'this is wrong, it needs the owner'
end
```
it shouldn't pass, but it does

```rb
    it { should_not validate_presence_of(:owner) }
```

However, this spec `spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb` is failing in lines 554, 605, and 694. 

I was checking with `pry-rails` and I saw that the matcher is passing through:
https://github.com/thoughtbot/shoulda-matchers/blob/195153cf42239935062f10f29a148d9d40032ca2/lib/shoulda/matchers/active_model/allow_value_matcher.rb#L405

But it should pass through:
https://github.com/thoughtbot/shoulda-matchers/blob/195153cf42239935062f10f29a148d9d40032ca2/lib/shoulda/matchers/active_model/allow_value_matcher.rb#L432

Unfortunately, I have not found out why. @VSPPedro @mcmire do you have any idea what might be going on?

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb
@@ -245,7 +245,8 @@ validation for you? Instead of using `validate_presence_of`, try
         end
 
         def allows_original_or_typecast_value?(value)
-          allows_value_of(value, @expected_message)
+          @expected_message == :blank ? allows_value_of(value, nil) : allows_value_of(value, @expected_message)
+          # allows_value_of(value, @expected_message)
         end
 
         def disallows_and_double_checks_value_of!(value)


HeitorMC:
Hey, @mcmire @VSPPedro any updates on this case?

mcmire:
I have a long response that I tried to write here, but the short answer is that it's intentionally flipped. When used in the positive, the `validate_presence_of` matcher needs to test that the model does _not_ allow `nil`; and when used in the negative, it needs to test that it _does_ allow `nil`. Under the hood, the `validate_presence_of` matcher uses the `allow_value` matcher to do this, so this:

``` ruby
it { should_not validate_presence_of(:owner) }
```

is sort of like saying this:

``` ruby
it { should allow_value(nil).for(:owner) }
```

That's why it passes through `failure_message` and not `failure_message_when_negated`.

PR #1546, opened 2023-03-17
FIx: Indifference to I18n errors.format translation

### LINKED ISSUE #1492: Indifference to I18n "errors.format" translation
# Problem
When "errors.format" translation is defined differently than  "%{attribute} %{message}", a test using `validate_presence_of` with `strict` seems not to be proved.

## Example
If the translations are defined as;
```
  errors: 
    format: "%{attribute}%{message}"
    messages: 
      blank: を入力してください
```

And if you write a test as below;
```
class A
  incude ActiveModel::Model
  include ActiveModel::Validations

  attr_accessor :value
  
  validates :value, presence: true, strict: true
end

class MenuItemTest < ActiveSupport::TestCase
  subject { A.new }
  
  context "Validations" do
    should validate_presence_of(:value).strict
  end
end
```

The test above isn't proved with message like;
```
  After setting :value to ‹nil›, the matcher expected the
  A to be invalid and to raise a validation
  exception with message "Value を入力してください". The record was indeed
  invalid, but the exception message was "Valueを入力してください"
  instead.
```
The matcher seems to expect a space exists between attribute and message, but actually doesn't.

# Possible Solution
Taking "errors.format" definition into account might be a good solution.
Seeing how Ruby on Rails compose exception, it uses "errors.format" to compose attribute and message int a string.
https://github.com/rails/rails/blob/4bbd05c90e5a62c0316307b3ca7c2262ed98b567/activemodel/lib/active_model/errors.rb#L290-L294
```
      I18n.t(:"errors.format", {
        :default   => "%{attribute} %{message}",
        :attribute => attr_name,
        :message   => message
      })
```

But validator seems to check messages with fixed format.
https://github.com/thoughtbot/shoulda-matchers/blob/206d2dae4c3bad8b6450a1762d06d315c35801aa/lib/shoulda/matchers/active_model/allow_value_matcher.rb#L615
Changing this code into the same manner as in RoR may solve the problem.

HeitorMC:
Fixes #1492 

I didn't find anything related to internationalization tests :thinking:, also both `spec/unit/shoulda/matchers/active_model/allow_value_matcher_spec.rb` and `spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb` passed without errors.

@VSPPedro @mcmire do you have any suggestions on how to test this case, or is this enough?

### DIFF
--- a/lib/shoulda/matchers/active_model/allow_value_matcher.rb
+++ b/lib/shoulda/matchers/active_model/allow_value_matcher.rb
@@ -612,7 +612,7 @@ pass, or do something else entirely.
 
         def default_expected_message
           if expects_strict?
-            "#{human_attribute_name} #{default_attribute_message}"
+            I18n.t(:"errors.format", attribute: human_attribute_name, message: default_attribute_message)
           else
             default_attribute_message
           end


VSPPedro:
Hello @HeitorMC, thanks for your PR. I'll take a look at how to do a test for this ASAP. Probably tomorrow or next week.

HeitorMC:
Hey, @VSPPedro any updates on this case?

mcmire:
Hi @HeitorMC. We fake i18n in tests. You can see existing usage of this in the following `context` in the tests for `allow_value`:

https://github.com/thoughtbot/shoulda-matchers/blob/fababc398deaae12cc94cab6508877fc9f13a946/spec/unit/shoulda/matchers/active_model/allow_value_matcher_spec.rb#L417-L441

It's a bit hard to see because of the indirection, but these tests verify that if `with_message` is given a symbol, `shoulda-matchers` uses the same i18n key to resolve that symbol as Rails does. You can see I18n being stubbed here:

https://github.com/thoughtbot/shoulda-matchers/blob/fababc398deaae12cc94cab6508877fc9f13a946/spec/support/unit/record_builder_with_i18n_validation_message.rb#L38

So, I think in order to test your PR, you'd want to add a new `context` and a new test inside of that test that verifies that when `with_message` is _not_ specified, but `strict` is, and `errors.format` is stubbed to return a certain string that's different than what Rails uses, then the test still passes. You don't have to create a new class like we've done for RecordBuilderWithI18nValidationMessage; I think you could keep it simpler and do something like this:

https://github.com/thoughtbot/shoulda-matchers/blob/fababc398deaae12cc94cab6508877fc9f13a946/spec/unit/shoulda/matchers/active_model/allow_value_matcher_spec.rb#L70-L71

I can suggest something a little more specific later, but let me know if that helps in the meantime.

PR #1549, opened 2023-04-03
Update Ruby versions

VSPPedro:


VSPPedro:
The error in the CI is caused by a intermitent test.

After updating to current versions, IC started throwing errors like:

```
An error occurred while loading ./spec/unit/shoulda/matchers/action_controller/filter_param_matcher_spec.rb.
Failure/Error: require environment_file_path

NameError:
  undefined method `=~' for class `Pry::Code'

```

Fixed after changing versions of pry-byebyg/pry.

Issue #1514, opened 2022-10-04
Testing with validates_comparison_of or comparison


fabriciobonjorno:
Guys, how are you? Has anyone implemented testing using these new Rails 7 validates? If yes, could you give me a hint?

First validate
`validates :start_date, comparison: { greater_than_or_equal_to: Date.today }`

Second validate
`validates :end_date, comparison: { greater_than: :start_date }`

Tutorial link:
https://www.bigbinary.com/blog/rails-7-adds-comparison-validator-to-active-record

matsales28:
Hello, I'm fine, thanks for asking!

I'd say that currently, `shoulda-matchers` does not have an option to test this new feature of Rails 7.
While we don't implement this on `shoulda-matchers` you can test this validation using this logic (just an example, you don't need to follow this pattern)

```ruby
class Order
  validates :start_date, comparison: { greater_than_or_equal_to: Date.today }
  validates :end_date, comparison: { greater_than: :start_date }
end

RSpec.describe Order, type: :model do
  describe "validations" do
    context "when start date is less than the current day" do
      it "record is not valid and adds greater_than_or_equal_to error to start_date attribute" do
        reward = Order.new(start_date: Time.zone.tomorrow)

        expect(reward.valid?).to be false
        expect(reward.errors[:start_date]).not_to be_empty
      end
    end

    context "when start date is not less than the current day" do
      it "record is valid and doesn't add greater_than_or_equal_to error to start_date attribute" do
        reward = Order.new(start_date: Time.zone.tomorrow)

        expect(reward.valid?).to be true
        expect(reward.errors[:estimated_delivers_on]).to be_empty
      end
    end

    context "when end date is not greater than the start_date" do
      it "record is not valid and adds greater_than error to end_date attribute" do
        reward = Order.new(start_date: Time.zone.today, end_date: Time.zone.today)

        expect(reward.valid?).to be false
        expect(reward.errors[:start_date]).not_to be_empty
      end
    end

    context "when end date is greater than the start_date" do
      it "record is valid and doesn't add greater_than error to end_date attribute" do
        reward = Order.new(start_date: Time.zone.today, end_date: Time.zone.tomorrow)

        expect(reward.valid?).to be true
        expect(reward.errors[:start_date]).to be_empty
      end
    end
  end
end
```

I can probably work on implementing this in the next weeks, but we are happy to receive new contributions if you want to work on this 😄

JuzerShakir:
[...]

PR #1552, opened 2023-04-14
feat: Implement comparison of matcher

matsales28:
Closes: #1514 

## Why:

In Rails, 7 was added a new option to perform validations: the [ComparisonValidator](https://github.com/rails/rails/pull/40095) with it you make a comparison between any object.

Before Rails 7
```ruby
class Event < ApplicationRecord
  validates :start_date, presence: true
  validates :end_date, presence: true

  validate :end_date_is_after_start_date

  private

  def end_date_is_after_start_date
    if end_date < start_date
      errors.add(:end_date, 'cannot be before the start date')
    end
  end
end
```

After Rails 7
```ruby
class Event < ApplicationRecord
  validates :start_date, presence: true
  validates :end_date, presence: true
  
  validates :end_date, comparison: { greater_than: :start_date }
  # OR
  validates_comparison_of :end_date, greater_than: :start_date
end
```

Those examples were extracted from [this blogpost](https://blog.kiprosh.com/rails7-activerecord-comparison-validator/).

## How

On shoulda, we used to have a class called `lib/shoulda/matchers/active_model/comparison_matcher.rb`. That class was used to implement logic to make comparisons for the numericality matcher. In this PR we are making that class generic more generic to be used on both `validate_numericality_of_matcher.rb` and `validate_comparison_of_matcher.rb` files.

### Compatibility changes
The first two commits of this PR are making adjustments to that and other classes to make it easy to use on the new `ValidateComparisoOfMatcher` class.

#### Compatibility changes to Submatchers and ComparisonMatcher
We needed to make sure the ComparisonMatcher would be working for both
ValidateComparisonOfMatcher and ValidateNumericalityOfMatcher, some
changes were required to make it work properly.

We also changed the order of the submatchers, where the `DissalowMatcher`
of the `ValidateNumericalityOfMatcher` will always be evaluated first,
because of that change, it was required some minor changes in the specs.

#### fix: Get failure_message_when_negated from DisalllowMatcher
This commit intends to, instead of getting the first submatcher
failure_message_when_negated message to get from the DisallowMatcher
first, because the messages are easier to read and debug.

#### feat: Implement ValidateComparisonOfMatcher class
This commit implements the base of the validate_comparion_of_matcher.
Creating the `ValidateComparisonOfMatcher` class and specs for the
submatchers `is_greater_than`, `is_greater_than_or_equal_to` and
`is_less_than`. In the next commits we'll implement the specs for the
other submatchers and for combinations of submatchers.

We are also missing the specs and implementation for comparing with
other values than just pure integers. We need to implement an way to
compare with Procs and also symbols.

#### feat: Implement support for comparing between string values
In this commit we added the support for making validations using the
`validate_comparison_of` matcher for string values. Ruby compares
strings using their ASCII/Unicode value, so for example:

```ruby
'aaa' > 'bbb' # false
'bbb' > 'aaa' # true
```

In this example `'aaa'` gets converted to an array with the unicode
values for the letters
```ruby
'aaa' #=> [97, 97, 97]
```

Ruby compares two strings character by character based on their Unicode
code point values. When comparing two strings, it compares the
corresponding characters at each position until it finds a difference.

If the characters at the current position are the same, it moves on to
the next position and continues the comparison. If the characters at
the current position are different, it compares their Unicode code
point values to determine which character is greater. The string with
the greater character at the first differing position is considered
greater than the other string.

#### feat: Implement support for comparing between date values
In this commit we added the support for making validations using the
`validate_comparison_of` matcher for date values.

#### feat: Implement allow_nil, strict and with_message qualifiers
This commit inteds to implement the `allow_nil`, `strict` and
`with_message` qualifiers on top of ComparisonMatcher.

These qualifiers were copy/pasted from the NumericalityMatcher on
previous commits and some of them required some minor adjustments to
work properly.

#### Disclaimer

I'm sorry for the super long PR but I wanted to try to cover all the cases with specs.


lib/shoulda/matchers/active_model/comparison_matcher.rb
```diff
@@ -0,0 +1,158 @@
...
+                   else @value
+                   end
+        end
+
+        def diffs_to_compare
+          diff_to_compare = @matcher.diff_to_compare
+          values = case @value
+                   when String then diffs_when_string(diff_to_compare)
+                   else [-1, 0, 1].map { |sign| @value + (diff_to_compare * sign) }
+                   end
```
matsales28, line 137: This change was required to make a comparison between strings possible, as we can see on the specs for the comparison implementation in Rails.

https://github.com/ChaelCodes/rails/blob/2ed77a9577d75579eabda04f8abb3a18c71322ea/activemodel/test/cases/validations/comparison_validation_test.rb#L33

lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb
```diff
@@ -13,29 +13,46 @@ def matches?(subject)
...
+
+          def failing_submatcher
+            failing_submatchers.last
+          end
+
+          def not_failing_submatcher
+            not_failing_submatchers.detect { |submatcher|
+              submatcher.instance_of?(DisallowValueMatcher)
+            } || not_failing_submatchers.first
+          end
```
matsales28, line 55: That was something that bothered me for a while when creating the first specs for the `validate_comparison_of` matcher, sometimes the failure message was very hard to understand because it was using `AllowValueMatcher` failure messages, the `DisallowValueMatcher` ones are easy to understand when dealing with matcher that didn't fail but was supposed to.

lib/shoulda/matchers/active_model/validate_numericality_of_matcher.rb
```diff
@@ -377,6 +377,8 @@ def initialize(attribute)
           @allowed_type_name = 'number'
           @context = nil
           @expected_message = nil
+
+          add_disallow_non_numeric_value_matcher
```
matsales28, line 381: I changed that behavior to the initializer because by doing this, we can ensure that the first matcher to be checked is always the one to disallow non-numerical values. With that, we can display more intuitive failure messages.

spec/unit/shoulda/matchers/active_model/validate_comparison_of_matcher_spec.rb
matsales28, line 1: I know this is huge, and I'm sorry about that. But I created specs based on the `validate_numericality_of_matcher_spec` file, and added specs for comparing between dates, integers, and strings and also using the value to compare as procs or symbols.

matsales28:
@mcmire @VSPPedro Just wanted to check in on this PR - it's been sitting there for a little while now. I understand that you folks may be busy with other things, so there's no rush on my end. However, if you could take a look and let me know your thoughts when you get the chance, I'd really appreciate it. I'd love to hear if there's anything I can improve.

Thanks for all your hard work on this project!

VSPPedro (approved):
LGTM! Sorry for taking so long.

tabfugnic:
Nicely done! Would love to see this get merged in.

matsales28:
@VSPPedro do you think we can merge, or should we wait for another review? I would love another review, especially from @mcmire, as this PR contains new features and some refactoring.

VSPPedro:
I think both options are good, but it's always helpful to hear what @mcmire thinks because he really knows this project inside out and has way more experience.

mcmire:
I'll try to prioritize this tomorrow!

lib/shoulda/matchers/active_model/comparison_matcher.rb
```diff
@@ -0,0 +1,158 @@
...
+          "#{comparison_expectation} #{@value}"
+        end
+
+        def submatchers
+          @_submatchers ||= NumericalityMatchers::Submatchers.new(build_submatchers)
+        end
+
+        private
+
+        def build_submatchers
```
mcmire, line 98: Hmm, it looks like we have `@matcher` and we have `submatchers`. Should we rename these so that people won't get confused by them?

spec/support/unit/change_value.rb
```diff
@@ -107,5 +111,9 @@ def change_value(value, value_changer)
     def dummy_value_for_column
       Shoulda::Matchers::Util.dummy_value_for(column_type)
     end
+
+    def date_column
```
mcmire, line 115: Since this returns a boolean, should this be:

``` suggestion
    def date_column?
```

lib/shoulda/matchers/active_model/validate_comparison_of_matcher.rb
```diff
@@ -0,0 +1,587 @@
...
+      #     end
+      #
+      # @return [ValidateComparisonOfMatcher]
+      #
+      def validate_comparison_of(attr)
+        ValidateComparisonOfMatcher.new(attr)
+      end
+
+      # @private
+      class ValidateComparisonOfMatcher
```
mcmire, line 282: So, one quirk of ValidateNumericalityOfMatcher (which is where I presume you copied this code from) is that it doesn't inherit from ValidationMatcher, whereas the other matchers do. It's something that we've been wanting to fix for a long time but have just never gotten around to it. I'd like to not continue that pattern if we can 😂 That said, I know that there are a lot of extra things in here that aren't in ValidationMatcher. How complicated in your estimation is it to use ValidationMatcher instead? If it's too much trouble then perhaps we can refactor this in a future PR but just wanted to point that out.

lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb
```diff
@@ -13,29 +13,46 @@ def matches?(subject)
...
           private
 
           def failing_submatchers
             @_failing_submatchers ||= @submatchers.reject do |submatcher|
               submatcher.matches?(@subject)
             end
           end
+
+          def not_failing_submatchers
+            @_not_failing_submatchers ||= @submatchers.reject do |submatcher|
```
mcmire, line 42: Would this read better?

``` suggestion
          def non_failing_submatchers
            @_non_failing_submatchers ||= @submatchers.reject do |submatcher|

File: lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb (hunk 2 above)
Comment (mcmire) [reply lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb:55]: Hmm... I'm wondering if perhaps we should leave it to the developer to add matchers in the order that would lead to a better user experience instead of trying to "fix" it for them. I can see this behavior being somewhat confusing as well. Is there another approach we could take here or what do you think?

**Review by mcmire — COMMENTED** [review]
Hi @matsales28, I'm so sorry it took me a while to review this. Overall this looks like great work — I just had some comments.

File: lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb (hunk 2 above)
Comment (matsales28) [reply lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb:55]: This is tricky because the developer can't define the order of the submatchers. After all, in this specific case, when the user defines one comparison matcher, we define three submatchers for that matcher, and that order is defined by the `shoulda-matchers` code.

Generates
```ruby
validates_comparison_of :participant_count, greater_than_or_equal_to: 2
# Generates in shoulda-matchers
# ComparisonMatcher with the following submatchers
# DisallowValueMatcher with value 1
# AllowValueMatcher with value 2
# AllowValue matcher with value 2
```

I totally get what you're saying about this behavior being confusing. Honestly, I'm fresh out of ideas here. I'm also fine reverting that.

lib/shoulda/matchers/active_model/comparison_matcher.rb
```diff
@@ -0,0 +1,158 @@
...
+          ERROR_MESSAGES[@operator][:assertions]
+        end
+
+        def build_option_value
+          @value = case @value
+                   when Proc then @value.call(@subject)
+                   when Symbol then @subject.send(@value)
+                   else @value
+                   end
+        end
```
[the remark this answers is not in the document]
matsales28: Sure thing. That was necessary because we accept "callables" instead of just "values". I like your suggestion and it's also aligned with the naming Rails uses
https://github.com/rails/rails/blob/4e71edfca8525417597b079440d876310dd6290d/activemodel/lib/active_model/validations/comparison.rb#LL21C31-L21C31

I'll make this adjustment!

lib/shoulda/matchers/active_model/comparison_matcher.rb
matsales28: Would you happen to have any idea here? Maybe calling `submatchers` as `comparison_submatchers`?

spec/support/unit/change_value.rb
matsales28: Yep! Interesting that my linting setup missed that. Ty for the suggestion!

lib/shoulda/matchers/active_model/validate_comparison_of_matcher.rb
matsales28: It shouldn't be too complicated, but I'll have to look deeply to ensure. But I can and would be happy to work on a follow-up PR for refactoring that. That PR is already big and complex, and I would like to avoid adding more complexity to it if that's fine for you.

lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb
matsales28: I'll adjust that!

lib/shoulda/matchers/active_model/validate_comparison_of_matcher.rb
mcmire: Yup, no problem!

lib/shoulda/matchers/active_model/comparison_matcher.rb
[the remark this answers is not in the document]
matsales28: I've adjusted that on ed3d371.

lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb
matsales28: Fixed on 16df8f1a

spec/support/unit/change_value.rb
matsales28: Made this change on 16df8f1a

lib/shoulda/matchers/active_model/numericality_matchers/submatchers.rb
matsales28: I reverted that change on [5906112](https://github.com/thoughtbot/shoulda-matchers/pull/1552/commits/59061129aa5913caa1536f1d3f76e8b70473f089)

mcmire: Ah sorry for not replying, I'm slammed with work currently 😵‍💫 Thank you! If this indeed continues to be confusing, I'm all for revisiting this. It'll also be easier to see the consequences of this change separately.

mcmire (approved):
This looks good to me. Again sorry it took me a while to get to this!

michaelgpearce:
I tried this out with a database `time` column, and it looks to not be working. I am getting this message:

```
     Failure/Error: it { is_expected.to validate_comparison_of(:schedule_start_time).is_less_than(:schedule_end_time) }
     
       Expected Routing::ScheduleTemplate to validate that :schedule_start_time
       looks like a value less than schedule_end_time, but this could not be
       proved.
         After setting :schedule_start_time to ‹"2000-01-02 01:00:01 UTC"› --
         which was read back as ‹Sat, 01 Jan 2000 01:00:01.000000000 UTC
         +00:00› -- the matcher expected the Routing::ScheduleTemplate to be
         invalid, but it was valid instead.
     
         As indicated in the message above, :schedule_start_time seems to be
         changing certain values as they are set, and this could have something
         to do with why this test is failing. If you've overridden the writer
         method for this attribute, then you may need to change it to make this
         test pass, or do something else entirely.
```

It looks like 1 day is being added, but for a `time` column, Rails always has the day as 2000-01-01 (so there should never be a value of 2000-01-02).

matsales28:
Hey @michaelgpearce, I've thoroughly investigated the issue and discovered that the problem occurs not only with past dates but all dates when using a time column in Rails. The behaviour you're observing is caused by the way Rails' typecast of dates to time handles comparisons.

Rails convert any date assigned to a time column to the (default timezone)[https://guides.rubyonrails.org/configuring.html#config-active-record-default-timezone] and sets the date as 2000-01-01. This means that any validation comparison expecting a time value other than 2000-01-01 will fail because Rails compares the entire Date object, not only the hours, minutes and seconds.

For example:
```ruby
class Example
  validates :start_time, comparison: { other_than: Time.zone.local(2021, 1, 1) }
end

example = Example.new(start_time: Time.zone.local(2000, 1, 1)) # We are not specifying the time, so the time should be the same as declared in the comparison
example.start_time #=> "2000-01-01 00:00:00.000000000 +0000"
example.valid? #=> true

# Rails will make the following comparison
example.start_time != Time.zone.local(2000, 1, 1) #=> true

```

This example demonstrates that Rails performs the comparison using the Time object, which, in this case, will be different due to the year and not the time. It's important to note that this behaviour is not limited to past dates but affects all dates when using a time column.

Another issue arises when using `shoulda-matchers` to validate time boundaries, such as the start and end of a day. The gem adds one second to the comparison value, which can cause unexpected behaviour.

For example:

```ruby
class Example
  validates :start_time, comparison: { less_than: Time.zone.local(2020, 1, 1).end_of_day }
end

# We cannot ensure that this validation will work using `shoulda-matchers` because we cannot test that a value greater than the end of the day will be invalid.
# Adding one second to the comparison value causes it to fall into the next day (2020-01-02). However, when we assign that value to `start_time`, Rails typecasts it to the beginning of the day (2000-01-01).
```

Due to these behaviours, validating time boundaries precisely when using a time column and `shoulda-matchers` can be challenging when working with borderline time values. Alternative approaches or custom validations might be necessary to test such scenarios accurately.

Please let me know if you have further questions or need additional assistance!

@mcmire do you think that makes sense?

Some references:
Implementation of Time in Rails and how it's typecasted: https://github.dev/rails/rails/rails/activemodel/lib/active_model/type/time.rb
How the comparison matcher compares values (I was expecting both the value and the value to be compared to be typecasted, but that's not true):
https://github.dev/rails/rails/rails/activemodel/lib/active_model/validations/comparison.rb
https://stackoverflow.com/questions/34978905/rails-activerecord-postgres-time-format

jeffdill2:
@matsales28 @VSPPedro Hi! Awesome work on this PR. 👏 

Do you know when this will be released? Not sure if you're waiting for some other things to be wrapped up before releasing a 5.4.0.

matsales28:
> @matsales28 @VSPPedro Hi! Awesome work on this PR. 👏
> 
> Do you know when this will be released? Not sure if you're waiting for some other things to be wrapped up before releasing a 5.4.0.

Hey @jeffdill2, we're planning to cut a release soon. We'd like to include support for Ruby 3.3 on that release, so I'd say until the end of the year, we'll launch a new release. In the meantime, I've created a branch that includes this feature, you can use it if you want to.

```ruby
gem 'shoulda-matchers', github: 'thoughtbot/shoulda-matchers', branch: 'with-comparison-matcher'
```

jeffdill2:
> > @matsales28 @VSPPedro Hi! Awesome work on this PR. 👏
> > Do you know when this will be released? Not sure if you're waiting for some other things to be wrapped up before releasing a 5.4.0.
> 
> Hey @jeffdill2, we're planning to cut a release soon. We'd like to include support for Ruby 3.3 on that release, so I'd say until the end of the year, we'll launch a new release. In the meantime, I've created a branch that includes this feature, you can use it if you want to.
> 
> ```ruby
> gem 'shoulda-matchers', github: 'thoughtbot/shoulda-matchers', branch: 'with-comparison-matcher'
> ```

Awesome, thanks @matsales28! 😄

PR #1555, opened 2023-04-28
feat: Implement `of_sql_type` qualifier on `have_db_column` matcher

### LINKED ISSUE #1432: sql_type check in have_db_column
Would you guys be open to a PR to add a method into the `have_db_column` chain that checks the `sql_type` of the column? Currently there is only one for `type` which I think is insufficient when dealing with certain column types.

For instance, say you are storing guids in a `char(36)` column, rails will interpret that as `:string`. In this case it is quite important that the column is actually a mysql `char(36)` instead.

```ruby
irb(main):008:0> MyModel.columns.find { |c| c.name == 'some_guid_column' }.type
=> :string
irb(main):009:0> MyModel.columns.find { |c| c.name == 'some_guid_column' }.sql_type
=> "char(36)"
irb(main):010:0>
```

I was thinking to just add an additional method into `have_db_column` so you can specify something like:

```ruby
it { is_expected.to have_db_column(:some_guid_column).of_sql_type('char(36)') }
```

I'm happy to do the PR if you guys are open to it.

matsales28:
Closes #1432

This commit intends to implement the `of_sql_type` qualifier on the `have_db_column` matcher. The implementation is quite simple and follows the pattern of the `HaveDbColumnMatcher` class.

### DIFF
--- a/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
+++ b/lib/shoulda/matchers/active_record/have_db_column_matcher.rb
@@ -48,6 +48,30 @@ module Shoulda
       #       should have_db_column(:camera_aperture).of_type(:decimal)
       #     end
       #
+      # ##### of_sql_type
+      #
+      # Use `of_sql_type` to assert that a column is defined as a certain sql_type.
+      #
+      #     class CreatePhones < ActiveRecord::Migration
+      #       def change
+      #         create_table :phones do |t|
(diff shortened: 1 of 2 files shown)


VSPPedro (approved):

VSPPedro:
LGTM!

mcmire (approved):
LGTM as well.

PR #1557, opened 2023-05-17
fix: Remove non-inclusive terms

matsales28:
We have been using both the terms "whitelist" and "blacklist" to refer to a list of approved/denied values, but I wanted to point out that our dev community recommends the term "allowlist" and "blocklist" instead.

References:
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/w/whitelist
https://inclusivenaming.org/word-lists/tier-1/#whitelist-blacklist
https://github.com/thoughtbot/guides/tree/main/terminology


mcmire (approved):
Ah good call! Sorry I didn't realize these were in there. Makes sense.

VSPPedro (approved):
Nice!

PR #1558, opened 2023-05-22
Add `normalize` matcher

stephannv:
## Why

Rails 7.1 adds `normalizes` API.
- [ActiveRecord::Base::normalizes API.](https://github.com/rails/rails/pull/43945)
- [Docs](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Normalization/ClassMethods.html#method-i-normalizes)

I think would be nice to shoulda-matchers to have a matchers to easily test the new `normalizes` API, similar to `allow_value`/`allow_values` matchers. I'm using on my project and this:
```ruby
it "normalizes email" do
  user = User.new(email: " \n myEmail@here.com \n\n")
  user.normalize_attribute(:email)
  expect(user.email).to eq "myemail@here.com"
end
```

Will become this:
```ruby
it { is_expected.to normalize(:email).from(" \n myEmail@here.com \n\n").to("myemail@here.com") }
```

What do you think about? This PR makes sense? 

Some considerations:
- I'm not sure about the API `normalize(attr)`, `from(value)` and `to(value)`, I'm open to change.
- `normalizes` accepts `apply_to_nil` option, but I think this should be tested in its own spec, like
  ```ruby
  shoulda normalize(:name).from(nil).to("Untitled")
  ```
  Another option would be:
  ```ruby
  shoulda normalize(:name).from(" jane doe ").to("Jane Doe").normalizing_nil_to("Untitled")
  ```
- I had to add Rails 7.1 to Appraisals to be able to test. I'm not sure how to restrict this tests to run only in rails_7_1. Other questions: Do we need to flag this matcher to run only in Rails 7.1 projects? Do we need to add this info to the docs?
- I think this is my first PR to shoulda-matchers, so, feedback are welcome.
- This PR is a work in progress, I opened it to receive feedbacks, so some non-related tests can be broken, or rubocop can be failing.

### DIFF
--- a/lib/shoulda/matchers/active_record/normalize_matcher.rb
+++ b/lib/shoulda/matchers/active_record/normalize_matcher.rb
@@ -1,0 +1,151 @@
+module Shoulda
+  module Matchers
+    module ActiveRecord
+      # The `normalize` matcher is used to ensure attribute normalizations
+      # are transforming attribute values as expected.
+      #
+      # Take this model for example:
+      #
+      #     class User < ActiveRecord::Base
+      #       normalizes :email, with: -> email { email.strip.downcase }
+      #     end
+      #
+      # You can use `normalize` providing an input and defining the expected
+      # normalization output:
+      #
+      #     # RSpec
+      #     RSpec.describe User, type: :model do
+      #       it do
+      #         should normalize(:email).from(" ME@XYZ.COM\n").to("me@xyz.com")
+      #       end
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class User < ActiveSupport::TestCase
+      #       should normalize(:email).from(" ME@XYZ.COM\n").to("me@xyz.com")
+      #     end
+      #
+      # You can use `normalize` to test multiple attributes at once:
+      #
+      #     class User < ActiveRecord::Base
+      #       normalizes :email, :handle, with: -> value { value.strip.downcase }
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe User, type: :model do
+      #       it do
+      #         should normalize(:email, :handle).from(" Example\n").to("example")
+      #       end
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class User < ActiveSupport::TestCase
+      #       should normalize(:email, handle).from(" Example\n").to("example")
+      #     end
+      #
+      # If the normalization accepts nil values with the `apply_to_nil` option,
+      # you just need to use `.from(nil).to("Your expected value here")`.
+      #
+      #     class User < ActiveRecord::Base
+      #       normalizes :name, with: -> name { name&.titleize || 'Untitled' },
+      #         apply_to_nil: true
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe User, type: :model do
+      #       it { should normalize(:name).from("jane doe").to("Jane Doe") }
+      #       it { should normalize(:name).from(nil).to("Untitled") }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class User < ActiveSupport::TestCase
+      #       should normalize(:name).from("jane doe").to("Jane Doe")
+      #       should normalize(:name).from(nil).to("Untitled")
+      #     end
+      #
+      # @return [NormalizeMatcher]
+      #
+      def normalize(*attributes)
+        if attributes.empty?
+          raise ArgumentError, 'need at least one attribute'
+        else
+          NormalizeMatcher.new(*attributes)
+        end
+      end
+
+      # @private
+      class NormalizeMatcher
+        attr_reader :attributes, :from_value, :to_value, :failure_message,
+          :failure_message_when_negated
+
+        def initialize(*attributes)
+          @attributes = attributes
+        end
+
+        def description
+          %(
+            normalize #{attributes.to_sentence(last_word_connector: ' and ')} from
+            ‹#{from_value.inspect}› to ‹#{to_value.inspect}›
+          ).squish
+        end
+
+        def from(value)
+          @from_value = value
+
+          self
+        end
+
+        def to(value)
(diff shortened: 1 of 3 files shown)


mcmire (commented):
Hey @stephannv, I think this is a great idea! I would probably suggest that we extract adding Rails 7.1 to another PR as it may cause CI failures. As for whether we should handle the `apply_to_nil` option by adding another method or not, I think we should keep it simple and stick to `from` and `to` like you've done here. 

One thing about this is that we might want to implement a negative version of the matcher. Every matcher can be executed via `should_not`, and that may match or may not match, just as with `should`. We might want to have another failure message in that case. 

Other than that this looks great!

stephannv:
Done @mcmire . Can you take a look again?

I tried to create another PR focused on adding Rails 7.1 to pipeline but I faced some issues, I will give another try later.

Sub-Xaero:
Anything updates on this PR? 
Happy to help out if more needs doing to take it over the line.

stephannv:
> Any updates on this PR? Happy to help out if more needs doing to take it over the line.

@Sub-Xaero I was waiting 7.1 release to update this PR, but I think I don't need to wait.

stephannv:
@Sub-Xaero I remember now, this PR is waiting `shoulda-matchers` to support Rails 7.1. If you want to help, I think you can open a PR with necessary changes to prepare shoulda-matchers for Rails 7.1

stephannv:
Hi @matsales28 , are you the new code owner for this PR? I rebased it with main, I think it's ready for review now. Your feedback will be very much appreciated.

lib/shoulda/matchers/active_record/normalize_matcher.rb
```diff
@@ -0,0 +1,151 @@
...
+          @failure_message_when_negated = build_failure_message_when_negated(attribute)
+          false
+        end
+
+        def normalize_attribute?(subject, attribute)
+          subject.public_send("#{attribute}=", from_value)
+          subject.normalize_attribute(attribute)
+
+          subject.public_send(attribute) == to_value
+        end
```
matsales28, line 133: To avoid using `public_send` and assigning it to the instance and then performing the normalization, what do you think about using the `normalize_value_for` class method? My initial idea was also to make the subject a class and not an instance, but that would require changing the default subject of the specs, and I don't think that's ideal.

```suggestion
        def normalize_attribute?(subject, attribute)
          subject_class = subject.class
          subject_class.normalize_value_for(attribute, from_value) == to_value
        end
```

matsales28 (commented):
Overall it looks good to me. Left some small suggestions but once adjusted we're ready to merge. One thing that could make it even better is when we're dealing with several attributes, it would be awesome if we could check all of them and then provide an error message listing all the attributes that didn't pass the normalization. Right now, the way it's set up, only the first one that doesn't meet the criteria shows up in the error message. We can make this on an next PR.

lib/shoulda/matchers/active_record/normalize_matcher.rb
stephannv: I think it makes sense. I considered using `.normalize_value_for` at first, but I dropped because I wanted to act on the subject itself, but I think it doesn't matter in this case.

stephannv:
> Overall it looks good to me. Left some small suggestions but once adjusted we're ready to merge. One thing that could make it even better is when we're dealing with several attributes, it would be awesome if we could check all of them and then provide an error message listing all the attributes that didn't pass the normalization. Right now, the way it's set up, only the first one that doesn't meet the criteria shows up in the error message. We can make this on an next PR.

Yeah @matsales28 , I think it would be a great improvement. 

Thanks for the feedback, I think I addressed everything. Let me know if I'm missing something.

spec/unit/shoulda/matchers/active_record/normalize_matcher_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        record = model.new
+
+        expect(record).to normalize(:name).from(' Jane Doe ').to('Jane Doe')
+        expect(record).to normalize(:name).from(nil).to('Untitled')
+      end
+    end
+
+    context "when subject doesn't normalize attribute that it shouldn't normalize" do
+      it 'does not match' do
+        model = define_model_class(:User, email: :string)
```
matsales28, line 87:
```suggestion
        model = define_model(:User, email: :string)
```

spec/unit/shoulda/matchers/active_record/normalize_matcher_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+          but it was normalized to ‹"Xy Z@Example.Com\\n"›
+        ).squish
+
+        expect(&assertion).to fail_with_message(message)
+      end
+    end
+
+    context 'when subject normalizes just one attribute incorrectly among multiple attributes' do
+      it 'fails' do
+        model = define_model_class(:User, email: :string, name: :string) do
```
matsales28, line 54:
```suggestion
        model = define_model(:User, email: :string, name: :string) do
```

spec/unit/shoulda/matchers/active_record/normalize_matcher_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+          normalizes :email, :name, with: -> (email) { email.strip.downcase }
+        end
+
+        expect(model.new).to normalize(:email, :name).from(" XyZ\n").to('xyz')
+      end
+    end
+
+    context 'when subject normalizes single attribute incorrectly' do
+      it 'fails' do
+        model = define_model_class(:User, email: :string) do
```
matsales28, line 35:
```suggestion
        model = define_model(:User, email: :string) do
```

matsales28 (changes requested):

stephannv:
Sorry for that @matsales28 , I think it's working now.

matsales28 (approved):

Issue #1559, opened 2023-05-25
Extend ActiveModel matchers to support arrays attributes


jarenas9539:
Hi, I noticed that ActiveModel matcher doesn't support array attributes on validations. Here is a simple example:
```
require "rails_helper"

class DummyModel < ApplicationRecord
  attribute :my_array, :string, array: true

  validates :my_array, length: { minimum: 1 }, inclusion: { in: [1, 2, 3, 4] }
end

RSpec.describe DummyModel do
  describe "validations" do
    it { is_expected.to validate_inclusion_of(:my_array).in_array([1, 2, 3, 4]) }
    it { is_expected.to validate_length_of(:my_array).is_at_least(1) }
  end
end
```

And these are the outputs:
```
Expected DummyModel to validate that :my_array is either ‹1›, ‹2›, ‹3›,
       or ‹4›, but this could not be proved.
         After setting :my_array to ‹1› -- which was read back as ‹"1"› -- the
         matcher expected the DummyModel to be valid, but it was invalid
         instead, producing these validation errors:
     
         * my_array: ["is not included in the list"]
     
         As indicated in the message above, :my_array seems to be changing
         certain values as they are set, and this could have something to do
         with why this test is failing. If you've overridden the writer method
         for this attribute, then you may need to change it to make this test
         pass, or do something else entirely.
```
```
       Expected DummyModel to validate that the length of :my_array is at least
       1, but this could not be proved.
         After setting :my_array to ‹"x"› -- which was read back as ‹[]› -- the
         matcher expected the DummyModel to be valid, but it was invalid
         instead, producing these validation errors:
     
         * my_array: ["is too short (minimum is 1 character)"]
     
         As indicated in the message above, :my_array seems to be changing
         certain values as they are set, and this could have something to do
         with why this test is failing. If you've overridden the writer method
         for this attribute, then you may need to change it to make this test
         pass, or do something else entirely.
```

`my_array` is being treated as a simple string type. Maybe could be tried to recognise this types inside the matchers, or force array samples with an extra option 🤔. I'd like to help contributing, but first I'd like to know your point of views about this approaches. Thanks on advance! 👍

mcmire:
Adding support for array attributes makes sense. My guess is that it's something that was added in a more recent version of Rails and it just hasn't been added yet. So we'd love a contribution!

PR #1560, opened 2023-05-31
Extend AM validate_length_of matcher to support array attributes

jarenas9539:
First step to close: https://github.com/thoughtbot/shoulda-matchers/issues/1559

# Solution

Were added two mechanisms to treat `validate_length_of_matcher` related attribute as array. One was through `array_column?` helper(also used on `validate_absence_of_matcher`) based on column type if `array?` helper is defined on model's `columns_hash` and another one with an explicit chain method `as_array` that will force to validate column with arrays values.
Why this way? First one will cover pure `array: true` columns for postgresql, second one could be dynamically used to check, for example, json columns storing arrays and being treated like that. This second approach will leave more freedom when using the validator.

Main validate_length_of_matcher specs were duplicated for the two variants. After all review's iteration for this PR concluded and if it's finally approved and merged this approach, I'll continue extending the array support to other validators(as inclusion/exclusion).

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
@@ -143,6 +143,12 @@ module Shoulda
             @subject.class.reflect_on_association(@attribute)
         end
 
+        def array_column?
+          @subject.class.respond_to?(:columns_hash) &&
+            @subject.class.columns_hash[@attribute.to_s].respond_to?(:array) &&
+            @subject.class.columns_hash[@attribute.to_s].array
+        end
+
         def enum_column?
           @subject.class.respond_to?(:defined_enums) &&
             @subject.class.defined_enums.key?(@attribute.to_s)
--- a/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
@@ -143,12 +143,6 @@ module Shoulda
             @subject.class.reflect_on_association(@attribute)
         end
 
-        def array_column?
-          @subject.class.respond_to?(:columns_hash) &&
-            @subject.class.columns_hash[@attribute.to_s].respond_to?(:array) &&
-            @subject.class.columns_hash[@attribute.to_s].array
-        end
-
         def enum_column?
           @subject.class.respond_to?(:defined_enums) &&
             @subject.class.defined_enums.key?(@attribute.to_s)
--- a/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
@@ -468,10 +468,7 @@ module Shoulda
         end
 
         def array_column?
-          @options[:array] || (
-            @subject.class.respond_to?(:columns_hash) &&
-            @subject.class.columns_hash[@attribute.to_s].array?
-          )
+          @options[:array] || super
         end
 
         def translated_short_message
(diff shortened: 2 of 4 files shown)


lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
```diff
@@ -143,12 +143,6 @@ def reflection
             @subject.class.reflect_on_association(@attribute)
         end
 
-        def array_column?
```
mcmire, line 146: Since this PR is focused on updating the `validate_length_of` matcher, should we do this in a different PR? This will just override the method you've added to ValidationMatcher, but that's okay for now.

lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
```diff
@@ -451,15 +456,19 @@ def allow_nil_does_not_match?
...
-          allows_value_of(string_of_length(length), message)
+          allows_value_of(element_of_length(length), message)
         end
 
         def disallows_length_of?(length, message)
-          disallows_value_of(string_of_length(length), message)
+          disallows_value_of(element_of_length(length), message)
+        end
+
+        def element_of_length(length)
```
mcmire, line 466: "Element" in this context sounds like it would be an element in an array, which isn't quite accurate. What do you think about `value_of_length`?

spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
```diff
@@ -370,15 +370,205 @@ def configure_validation_matcher(matcher)
...
+        expect(validating_length(type: :jsonb, is: 4)).
+          not_to validate_length_of(:attr).as_array.is_equal_to(5).with_message(/.*/)
+      end
+
+      it 'does not override the default message with a blank' do
+        expect(validating_length(type: :jsonb, is: 4)).
+          to validate_length_of(:attr).as_array.is_equal_to(4).with_message(nil)
+      end
+    end
+  end
```
mcmire, line 555: We should also add tests against an ActiveModel attribute that is marked as an array. There doesn't appear to be any tests that use an ActiveModel model in this file, so you may have to define a new method that constructs such a model instead of an ActiveRecord model. Instead of the `define_model` helper, you should be able to use the `define_active_model_class` helper to do so. See: https://github.com/jarenas9539/shoulda-matchers/blob/feature/extending-active-model-validating-length-of-to-allow-arrays/spec/unit/shoulda/matchers/active_model/validate_presence_of_matcher_spec.rb#L1134.

mcmire (commented):
These changes make sense to me. Just had a few comments.

lib/shoulda/matchers/active_model/validate_absence_of_matcher.rb
jarenas9539: I decided to do it in this PR to avoid repetition of `array_column?` and to avoid forgetting to remove it later 😅, I'll roll it back and will do the change on another PR after this one gets merged 👍

jarenas9539:
@mcmire changes applied. This one is ready for another iteration. 👍

mcmire (commented):
Looking better! I just had two more suggestions related to documentation.

lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
```diff
@@ -260,6 +260,27 @@ module ActiveModel
       #       should validate_length_of(:bio).is_at_least(15).allow_blank
       #     end
       #
+      # ##### as_array
+      #
+      # Use `as_array` to force the attribute being validate with array values
+      #
+      #     class User
+      #       include ActiveModel::Model
+      #       attr_accessor :arr, array: true
```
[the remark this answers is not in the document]
jarenas9539: Good catches! Already updated. Thanks 👍.

mcmire (approved):
Looks good! Thanks so much for the PR!

PR #1566, opened 2023-07-07
refactor: Make `ValidateNumericalityOfMatcher` inherit from `ValidationMatcher`

matsales28:
This commit intends to refactor the `ValidateNumericalityOfMatcher` class in order to make it inherit from `ValidationMatcher` to continue on that pattern and make it consistent across all validation classes.

In the next PR/commit we'll refactor the ValidateComparisonOfMatcher also to follow that pattern.

The major change in that PR is the way we are creating the AllowValueMatcher when using the `allow_nil` qualifier.


VSPPedro (approved):
LGTM!

mcmire (approved):
Amazing!

PR #1567, opened 2023-07-14
feat: Make `ValidateComparisonOfMatcher` inherit from `ValidationMatcher`

matsales28:
This PR intends to refactor the `ValidateComparisonOfMatcher` class to make it inherit from `ValidationMatcher` to continue on that pattern and make it consistent across all validation classes.

We are also removing some existing methods in the `ValidationMatcher` class from the `ValidateNumericalityOfMatcher` class.


VSPPedro (approved):
LGTM!

mcmire (approved):
Huh... I didn't realize it was so easy. Makes me think we should have tried this years ago 😅 LGTM!

PR #1569, opened 2023-08-11
feat: Allow length validation on associations

### LINKED ISSUE #1007: validate_length_of does not work when used with associations

matsales28:
This commit allows the length matcher to be used on associations. It does this by checking if the attribute is an association, and if so, it uses the associations as the attribute to validate.

This commit also adds tests for the length matcher on associations (has_many and has_many through).

I want to credit @prashantjois for the initial work on this feature #1124. I took his work and expanded on it.

If/When this PR gets merged, it'll close #1007 and #1124.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
@@ -3,7 +3,7 @@ module Shoulda
     module ActiveModel
       # The `validate_length_of` matcher tests usage of the
       # `validates_length_of` matcher. Note that this matcher is intended to be
-      # used against string columns and not integer columns.
+      # used against string columns and associations and not integer columns.
       #
       # #### Qualifiers
       #
@@ -36,7 +36,8 @@ module Shoulda
       #
       # Use `is_at_least` to test usage of the `:minimum` option. This asserts
       # that the attribute can take a string which is equal to or longer than
-      # the given length and cannot take a string which is shorter.
+      # the given length and cannot take a string which is shorter. This qualifier
+      # also works for associations.
       #
       #     class User
       #       include ActiveModel::Model
@@ -61,7 +62,8 @@ module Shoulda
       #
       # Use `is_at_most` to test usage of the `:maximum` option. This asserts
       # that the attribute can take a string which is equal to or shorter than
-      # the given length and cannot take a string which is longer.
+      # the given length and cannot take a string which is longer. This qualifier
+      # also works for associations.
       #
       #     class User
       #       include ActiveModel::Model
@@ -84,7 +86,8 @@ module Shoulda
       #
       # Use `is_equal_to` to test usage of the `:is` option. This asserts that
       # the attribute can take a string which is exactly equal to the given
-      # length and cannot take a string which is shorter or longer.
+      # length and cannot take a string which is shorter or longer. This qualifier
+      # also works for associations.
       #
       #     class User
       #       include ActiveModel::Model
@@ -106,7 +109,7 @@ module Shoulda
       # ##### is_at_least + is_at_most
       #
       # Use `is_at_least` and `is_at_most` together to test usage of the `:in`
-      # option.
+      # option. This qualifies also works for associations.
       #
       #     class User
       #       include ActiveModel::Model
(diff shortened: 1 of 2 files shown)


lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
```diff
@@ -487,13 +490,29 @@ def disallows_length_of?(length, message)
...
 
         def value_of_length(length)
-          (array_column? ? ['x'] : 'x') * length
+          if array_column?
+            ['x'] * length
+          elsif association?
+            Array.new(length) { association.klass.new }
+          else
+            'x' * length
+          end
```
matsales28, line 499: My initial idea was to multiply by `length` the return of the condition, like the example bellow.
```ruby
if array_column?
  ['x']
elsif association?
  [association.klass.new]
else
  'x'
end * length
```

However, this causes the array of the associated object to have the same `object_id`, and Rails prevents the same object from being associated twice check on [collection_association](https://github.com/rails/rails/blob/48243d42aad9675e26408aaa561e677dac91db04/activerecord/lib/active_record/associations/collection_association.rb#L416-L426) class on active record gem for more context. That's why I needed to change this logic, so we can have different `object_ids` on the array of associations.

https://github.com/rails/rails/blob/48243d42aad9675e26408aaa561e677dac91db04/activerecord/lib/active_record/associations/collection_association.rb#L416-L426

lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
```diff
@@ -61,7 +62,8 @@ module ActiveModel
       #
       # Use `is_at_most` to test usage of the `:maximum` option. This asserts
       # that the attribute can take a string which is equal to or shorter than
-      # the given length and cannot take a string which is longer.
+      # the given length and cannot take a string which is longer. Thi qualifier
```
mcmire, line 65: Nit: Thi -> This

``` suggestion
      # the given length and cannot take a string which is longer. This qualifier
```

spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
```diff
@@ -370,6 +370,190 @@ def configure_validation_matcher(matcher)
...
+      it 'fails when used in the negative' do
+        assertion = lambda do
+          expect(validating_through_association_length(minimum: 4)).
+            not_to validate_length_of(:children).is_at_least(4)
+        end
+
+        message = <<-MESSAGE
+Expected Parent not to validate that the length of :children is at least
+4, but this could not be proved.
+  After setting :children to ‹[#<Child id: nil>, #<Child id: nil>,
```
mcmire, line 496: This is out of scope for this PR, but this error message seems strange to me. If a model doesn't validate that an association matches a certain length, then it should allow a value that would have been invalid with the validation in place. I don't see why having four children should be invalid in this case. If anything having _three_ children should be _valid_. Of course that counts as breaking behavior, and it also could be inconsistent with how "negative" versions of other matchers work. It would have to be changed in concert with other matchers, probably. So I'm not eager to change it now, but I thought I'd point it out.

mcmire (approved):
Nice! This looks great.

spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
matsales28: I share your sentiment. The "negative" message sounded very strange to me when I first started working on shoulda-matchers. However, I attributed this perception to my non-native English background. Reading it in Portuguese, my primary language, sounded weird. Should we open an issue to discuss this?

lib/shoulda/matchers/active_model/validate_length_of_matcher.rb
matsales28: Nice catch! Adjusted it.

spec/unit/shoulda/matchers/active_model/validate_length_of_matcher_spec.rb
mcmire: Yeah, that sounds like a good idea!

PR #1573, opened 2023-10-06
feat: Add support for Rails 7.1

matsales28:
Adding support for Rails 7.1 version, there are several commits each one fixing one problem in the specs. I'd recommend going through them one by one when reviewing, it'll be easier to review.


VSPPedro (approved):
LGTM!

I'm thinking of making a new version of shoulda-matchers once we incorporate support for Ruby 3.2. What are your thoughts on this idea?

matsales28:
> I'm thinking of making a new version of shoulda-matchers once we incorporate support for Ruby 3.2. What are your thoughts on this idea?

@VSPPedro I like the idea, but do you mean support for Ruby 3.3, right? If I'm not wrong we are already supporting Ruby 3.2. I have never released a new version of a gem. If we can do that together somehow I'd love to participate.

PR #1574, opened 2023-10-10
Bump `appraisal` gem

y-yagi:
I tried to run specs on my local, but it didn't work. It seems that ``appraisal` v2.4 has a problem with the latest bundler. So I would like to update to the V2.5.

Ref: https://github.com/thoughtbot/appraisal/issues/210#issuecomment-1562556347


VSPPedro (approved):

VSPPedro:
Thank you for your contribution!

Issue #1553, opened 2023-04-16
Can't be used without Rails and with ActiveModel but not ActiveRecord?


dgmstuart:
As I interpret the documentation, it seems like it should be possible to use shoulda-matchers without rails, with `ActiveModel` and not `ActiveRecord`?

I'm trying to use `validates_presence_of` in an isolated spec of an object which includes `ActiveModel`.

This fails with `uninitialized constant ActiveRecord`.

This seems to be because the matcher depends on `RailsShim`: [shoulda/matchers/active_model/validate_presence_of_matcher.rb#L385](https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb#L385)

...which does a type check of attributes against `::ActiveRecord::Type::Serialized`.
[shoulda/matchers/rails_shim.rb#L65](https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/rails_shim.rb#L65)

It seems like this might be fixable with a check like `Object.const_defined('::ActiveRecord::Type::Serialized')`, but I'm not sure if I understand enough what `RailsShim` is to understand if this is a good idea.

stonefield:
I tried to write a spec for @dgmstuart suggested implementation, without succeeding. 
I conclude that it is impossible to test for this as long as we do not want to destroy the internal constant mappings.
However, I think the suggested implementation should be done without a spec (since it is not possible).
It will at least allow us which is only using ActiveModel to use matchers.

So please replace the line [shoulda/matchers/rails_shim.rb#L65](https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/rails_shim.rb#L65) with:
```
              if Object.const_defined?('ActiveRecord::Type::Serialized') && attribute_type.is_a?(::ActiveRecord::Type::Serialized)

```

PR #1580, opened 2023-11-09
fix: Ensure that validation specs work for ActiveModel without ActiveRecord

stonefield:
This change addresses issue #1553
It is a simple one liner allowing ActiveModel related matchers to run when ActiveRecord is not used.

There are no specs for this change due the fact that stubbing of the constant namespace is neither recommended, nor possible.

### DIFF
--- a/lib/shoulda/matchers/rails_shim.rb
+++ b/lib/shoulda/matchers/rails_shim.rb
@@ -62,7 +62,7 @@ module Shoulda
         def serialized_attributes_for(model)
           attribute_types_for(model).
             inject({}) do |hash, (attribute_name, attribute_type)|
-              if attribute_type.is_a?(::ActiveRecord::Type::Serialized)
+              if Object.const_defined?('ActiveRecord::Type::Serialized') && attribute_type.is_a?(::ActiveRecord::Type::Serialized)
                 hash.merge(attribute_name => attribute_type.coder)
               else
                 hash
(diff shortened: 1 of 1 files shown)


lib/shoulda/matchers/rails_shim.rb
```diff
@@ -62,7 +62,7 @@ def generate_validation_message(
         def serialized_attributes_for(model)
           attribute_types_for(model).
             inject({}) do |hash, (attribute_name, attribute_type)|
-              if attribute_type.is_a?(::ActiveRecord::Type::Serialized)
+              if Object.const_defined?('ActiveRecord::Type::Serialized') && attribute_type.is_a?(::ActiveRecord::Type::Serialized)
```
matsales28, line 65: What about moving the `const_defined?` check to the beginning of the method? This way we avoid making this check for every attribute of the model and fix the Rubocop line length issue.

Something along those lines:
```ruby
type_serialized_defined = Object.const_defined?('ActiveRecord::Type::Serialized')
attribute_types_for(model).
            inject({}) do |hash, (attribute_name, attribute_type)|
              if type_serialized_defined && attribute_type.is_a?(::ActiveRecord::Type::Serialized)
```

matsales28 (approved):
It looks good for me. Left a small suggestion for a small improvement and to adjust the Rubocop warning.

Thanks for your contribution!

lib/shoulda/matchers/rails_shim.rb
stonefield: Thank you. I have changed the code accordingly.

If we want to support any serialized type we may change the code to the following:

```
          attribute_types_for(model).
            inject({}) do |hash, (attribute_name, attribute_type)|
              if attribute_type.respond_to?(:coder)
```
`ActiveRecord::Type::Serialized` is the only type that responds to `coder`.

PR #1581, opened 2023-12-04
add have_encrypt_attribute_matcher to test usage of the encrypts macro

theforestvn88:
Since Rails 7.0, [Active Record supports application-level encryption](https://guides.rubyonrails.org/v7.0/active_record_encryption.html).
So i think it would be nice to have a matcher that help to test  usage of the `encrypts` macro:

```ruby
RSpec.describe Survey, type: :model do
    it { should have_encrypt_attribute(:access_code).downcase(true)  }
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,8 @@ about any of them, make sure to [consult the documentation][rubydocs]!
   tests usage of `validates_uniqueness_of`.
 * **[normalize](lib/shoulda/matchers/active_record/normalize_matcher.rb)** tests
   usage of the `normalize` macro
+* **[encrypt](lib/shoulda/matchers/active_record/encrypt_matcher.rb)**
+  tests usage of the `encrypts` macro.
 
 ### ActionController matchers
 
--- a/lib/shoulda/matchers/active_record.rb
+++ b/lib/shoulda/matchers/active_record.rb
@@ -25,6 +25,7 @@ require 'shoulda/matchers/active_record/uniqueness'
 require 'shoulda/matchers/active_record/validate_uniqueness_of_matcher'
 require 'shoulda/matchers/active_record/have_attached_matcher'
 require 'shoulda/matchers/active_record/normalize_matcher'
+require 'shoulda/matchers/active_record/encrypt_matcher'
 
 module Shoulda
   module Matchers
--- a/lib/shoulda/matchers/active_record/encrypt_matcher.rb
+++ b/lib/shoulda/matchers/active_record/encrypt_matcher.rb
@@ -1,0 +1,174 @@
+module Shoulda
+  module Matchers
+    module ActiveRecord
+      # The `encrypt` matcher tests usage of the
+      # `encrypts` macro (Rails 7+ only).
+      #
+      #     class Survey < ActiveRecord::Base
+      #       encrypts :access_code
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Survey, type: :model do
+      #       it { should encrypt(:access_code) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class SurveyTest < ActiveSupport::TestCase
+      #       should encrypt(:access_code)
+      #     end
+      #
+      # #### Qualifiers
+      #
+      # ##### deterministic
+      #
+      #     class Survey < ActiveRecord::Base
+      #       encrypts :access_code, deterministic: true
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Survey, type: :model do
+      #       it { should encrypt(:access_code).deterministic(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class SurveyTest < ActiveSupport::TestCase
+      #       should encrypt(:access_code).deterministic(true)
+      #     end
+      #
+      # ##### downcase
+      #
+      #     class Survey < ActiveRecord::Base
+      #       encrypts :access_code, downcase: true
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Survey, type: :model do
+      #       it { should encrypt(:access_code).downcase(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class SurveyTest < ActiveSupport::TestCase
+      #       should encrypt(:access_code).downcase(true)
+      #     end
+      #
+      # ##### ignore_case
+      #
+      #     class Survey < ActiveRecord::Base
+      #       encrypts :access_code, deterministic: true, ignore_case: true
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Survey, type: :model do
+      #       it { should encrypt(:access_code).ignore_case(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class SurveyTest < ActiveSupport::TestCase
+      #       should encrypt(:access_code).ignore_case(true)
+      #     end
+      #
+      # @return [EncryptMatcher]
+      #
+      def encrypt(value)
+        EncryptMatcher.new(value)
+      end
+
+      # @private
+      class EncryptMatcher
+        def initialize(attribute)
+          @attribute = attribute.to_sym
+          @options = {}
+        end
+
+        attr_reader :failure_message, :failure_message_when_negated
+
+        def deterministic(deterministic)
+          with_option(:deterministic, deterministic)
+        end
+
+        def downcase(downcase)
+          with_option(:downcase, downcase)
+        end
+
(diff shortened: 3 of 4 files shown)


lib/shoulda/matchers/active_record/have_encrypt_attribute_matcher.rb
```diff
@@ -0,0 +1,160 @@
...
+                else
+                  @failure_message = "#{opt} option should be #{expected}, got #{actual}"
+                  return false
+                end
+              else
+                true
+              end
+            end
+            true
+          end
```
matsales28, line 143: I think we could simplify that method using guard clauses it feels a little confusing with all the `true`, and `false` values being returned. Another small improvement would be getting the value from the `encrypted_attribute_scheme` using the predicate public methods of that class. https://github.com/rails/rails/blob/900dc3280932df4a614cfa06b59959f249e65ce1/activerecord/lib/active_record/encryption/scheme.rb#L31-L40

On the other hand, I think we could also adjust the `failure_message` to be more explicit.

```suggestion
          def options_correct?(*opts)
            opts.all? do |opt|
              next true unless @options.key?(opt)

              expected = @options[opt]
              actual = encrypted_attribute_scheme.public_send("#{opt}?")

              next true if expected == actual

              @failure_message = "Expected to #{description} of #{class_name} using :#{opt} option as ‹#{expected}›, but got ‹#{actual}›"
              false
            end
          end
```

This change would require some changes in the specs, but I think it's a nice improvement, what do you think?

lib/shoulda/matchers/active_record/have_encrypt_attribute_matcher.rb
```diff
@@ -0,0 +1,160 @@
...
+            encrypted_attributes_included? &&
+              options_correct?(
+                :deterministic, 
+                :downcase, 
+                :ignore_case
+              )
+          end
+  
+          def description
+            "encrypt #{@attribute}"
```
matsales28, line 112: Small tweak 
```suggestion
            "encrypt :#{@attribute}"

```

lib/shoulda/matchers/active_record/have_encrypt_attribute_matcher.rb
```diff
@@ -0,0 +1,160 @@
...
+  
+          private
+
+          def encrypted_attributes_included?
+            if encrypted_attributes.include?(@attribute)
+              @failure_message_when_negated = "Did not expect #{@attribute} to be encrypted"
+              true
+            else
+              @failure_message = "Expect #{@attribute} should be encrypted, but not"
+              false
```
matsales28, line 123:
```suggestion
          def encrypted_attributes_included?
            if encrypted_attributes.include?(@attribute)
              @failure_message_when_negated = "Did not expect to #{description} of #{class_name}, but it did"
              true
            else
              @failure_message = "Expected to #{description} of #{class_name}, but it did not"
              false
          end
```

lib/shoulda/matchers/active_record/have_encrypt_attribute_matcher.rb
```diff
@@ -0,0 +1,160 @@
...
+            @attribute = attribute.to_sym
+            @options = {}
+          end
+  
+          attr_reader :failure_message, :failure_message_when_negated
+          
+          def deterministic(deterministic)
+            @options[:deterministic] = deterministic
+            self
+          end
```
matsales28, line 89: What do you think about extracting that to a method? The same pattern is used across the options.

Something like this:
```
def with_option(option_name, value)
  @options[option_name] = value
  self
end

def deterministic(deterministic)
  with_option(:deterministic, deterministic)
end
```

spec/unit/shoulda/matchers/active_record/have_encrypt_attributes_matcher_spec.rb
```diff
@@ -0,0 +1,139 @@
+require 'unit_spec_helper'
+
+describe Shoulda::Matchers::ActiveRecord::HaveEncryptAttributeMatcher, type: :model do
```
matsales28, line 3: Do you mind adding specs for the negated_matcher failure message?

spec/unit/shoulda/matchers/active_record/have_encrypt_attributes_matcher_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      it 'assigns a failure message' do
+        model = define_model(:example, attr: :string).new
+        matcher = have_encrypt_attribute(:attr)
+
+        matcher.matches?(model)
+
+        expect(matcher.failure_message).to match(/Expect attr should be encrypted, but not/)
+      end
+    end
```
matsales28, line 35: We could improve all the specs that are checking for the `failure_message` directly to actually check if the matcher is failing with the correct message. That's better than checking directly the `failure_message`.

Here's an example, but you can also check the [spec/unit/shoulda/matchers/active_model/validate_comparison_of_matcher_spec.rb](https://github.com/thoughtbot/shoulda-matchers/blob/main/spec/unit/shoulda/matchers/active_model/validate_comparison_of_matcher_spec.rb) file for more examples.
```ruby
        context 'when the column is an string column' do
          it 'rejects and raise an error' do
            record = build_record_validating_comparison(column_type: :string, equal_to: 10)

            assertion = lambda do
              expect(record).to validate_comparison_of(:attr).is_equal_to(10)
            end

            expect(&assertion).to fail_with_message(<<~MESSAGE)
Expected Example to validate that :attr looks like a value equal to 10,
but this could not be proved.
  After setting :attr to ‹"10"›, the matcher expected the Example to be
  valid, but it was invalid instead, producing these validation errors:

  * attr: ["must be equal to 10"]
            MESSAGE
          end
        end
```

matsales28 (commented):
Thanks a lot for your contribution! I left some comments, please let me know if I can help you with anything to move this PR forward. There are also some linting issues. You can check them [here](https://github.com/thoughtbot/shoulda-matchers/actions/runs/7085054572/job/19460712944?pr=1581).

Once again, thank you for contributing.

lib/shoulda/matchers/active_record/have_encrypt_attribute_matcher.rb
theforestvn88: it's better!

spec/unit/shoulda/matchers/active_record/have_encrypt_attributes_matcher_spec.rb
theforestvn88: what you mean ? could you give me some examples like other reviews ?

matsales28: What I meant was to use the matcher in the negated version and assert the failure message.
A quick can be found on : https://github.com/thoughtbot/shoulda-matchers/blob/baabf89127d712855684b351073764026a6c1df1/spec/unit/shoulda/matchers/active_model/validate_comparison_of_matcher_spec.rb#L164-L182

danielolivaresd:
This would be a great addition. May I suggest a name change? ~`have_encrypt_attribute`~ `have_encrypted_attribute`

matsales28:
> This would be a great addition. May I suggest a name change? ~`have_encrypt_attribute`~ `have_encrypted_attribute`

After a second look, this new method/matcher is quite similar to the `serialize` one, and our matcher is just `serialize` in that case. I'd say to keep a pattern across the matches to change to `encrypts(:attr)` instead of `have_encrypt_attribute`. I'd love to hear your thoughts on this @VSPPedro.

matsales28:
@theforestvn88 I think it would be better to name this method as `encrypts(:attr)`. Do you mind updating your PR with this change? Also, there's a small conflict on the `README.md` file, can you resolve it?

VSPPedro:
Sorry for taking so long to respond. I agree. I think your suggestion it's a good idea.

theforestvn88:
When the first time i use `shoulda`, i surprised that it uses `should have_many()` meanwhile Rails use `has_many()` because i thought that it provide some thing like `should <whatever rails do>` to test `<whatever rails do>`. Turn out, you guys try to correct the  `grammar` here.

Anyway, i just want to confirm that in this case what willwe should  use: `should encrypts(:attr)` or   `should encrypt(:attr)` without `s` ?

VSPPedro:
Good question! `should encrypt(:attr)`, without `s`, has my vote.

matsales28:
I'm fine either way. Let's use `encrypt(:attr)` then.

matsales28 (approved):

VSPPedro (approved):
LGTM!

PR #1584, opened 2023-12-11
chore: Remove references to unsupported Rails versions

matsales28:
This commit removes references to Rails versions that are no longer supported by `shoulda-matchers`. This includes:

* Rails 4.2
* Rails 5.0
* Rails 5.1

We also simplified the logic on some places to remove the need for checking the Rails version.


VSPPedro (approved):
LGTM!

PR #1590, opened 2023-12-19
fix: Adjust flaky spec on `ValidateAbsenceOfMatcher`

matsales28:
I'm not entirely sure that this will fix this flaky spec, but my bet is that sometimes an invalid `uuid` was generated using the `next` method, this commit avoids that randomness by always generating a new and valid `uuid`.


VSPPedro (approved):
It makes perfect sense to me! I recall when the ValidateAbsenceOfMatcher intermittent test first started failing. Nice catch!

PR #1591, opened 2023-12-19
chore: Update Rails versions

matsales28:
This commit updates the Rails versions used in the Gemfiles.


spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
```diff
@@ -1335,48 +1335,6 @@ def having_many_non_existent_class(model_name, assoc_name, options = {})
         and_fail_with(expected_message)
     end
 
-    it 'rejects an association with a valid :class_name and a bad :foreign_key option (CPK)' do
```
matsales28, line 1338: As I mentioned in the commit message, those specs were testing passing an array to the `foreign_key` qualifier, which is an invalid assignment, so I decided to remove them.
> [fix: Remove composite foreign_key specs](https://github.com/thoughtbot/shoulda-matchers/pull/1591/commits/38be129a2ecd0fe6c36d2971a4da22508479fee0) 
> This commit removes the specs for composite foreign keys. This is
because Rails never supported passing arrays to `foreign_key` modifier
and it will raise an error on Rails 7.1.2+.
>Check the following commit for more information:
>https://github.com/rails/rails/commit/529d1f55a8138798d92a6f2aba01681315958ab3

I'll also open an issue to create the new `query_constraints` qualifier that was introduced in Rails 7.1 to serve this purpose.

https://edgeguides.rubyonrails.org/association_basics.html#associations-between-models-with-composite-primary-keys

VSPPedro: LGTM! You read my mind, I was planning on updating Rails. Thank you for working on this!

VSPPedro (approved):

PR #1600, opened 2023-12-26
chore: bump rubocop to 1.59.0

VSPPedro:
- chore: bump rubocop to 1.59.0
- style: fix rubocop offense


VSPPedro:
Ruby 3.3.0 just came out. When I tried running it with shoulda-matchers, but there is on problem, as you can see in this [report](https://github.com/thoughtbot/shoulda-matchers/actions/runs/7331570636/job/19964413985).

To sort this out properly, we need to update RuboCop.

What do you think?

matsales28 (approved):
Yeah, it looks good to me! Only a minor question on the change of the `platform` in the Gemfile.

Gemfile.lock
```diff
@@ -72,14 +76,14 @@ GEM
...
-    unicode-display_width (2.4.2)
+    unicode-display_width (2.5.0)
     warnings_logger (0.1.1)
     yard (0.9.25)
     zeus (0.15.14)
       method_source (>= 0.6.7)
 
 PLATFORMS
-  ruby
+  x86_64-linux
```
matsales28, line 86: Was that change intended?

VSPPedro: No, nice catch!

PR #1606, opened 2024-01-19
feat: Add `have_delegated_type` matcher

matsales28:
On this PR, we're adding the matcher for the `delegated_type` macro, which is basically a `belongs_to` polymorphic association where you define the possible classes the polymorphic relation can act as.

Currently, it is hard to check those possible classes as Rails does not expose them in any way, so we need to check if the methods for those classes are defined, which is not ideal.

```ruby
class Vehicle < ActiveRecord::Base
  delegated_type :drivable, types: %w(Car Truck)
end

# Usage 
RSpec.describe Vehicle, type: :model do
 it { should have_delegated_type(:drivable) }
end
```


lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -1414,6 +1730,30 @@ def touch_correct?
...
+            types = options[:types]
+
+            correct = types.all? do |type|
+              scope_name = type.tableize.tr('/', '_')
+              singular   = scope_name.singularize
+              query      = "#{singular}?"
+
+              Object.const_defined?(type) && @subject.respond_to?(query) &&
+                @subject.respond_to?(singular)
+            end
```
matsales28, line 1744: This is not great, but currently, there's no way to check the types besides doing that. A PR was merged recently, adding a `types` method to the object, which will greatly simplify this check. I will update this method once it is released in Rails.

https://github.com/rails/rails/pull/50662

lib/shoulda/matchers/active_record/association_matchers/model_reflection.rb
```diff
@@ -13,7 +13,11 @@ def initialize(reflection)
           end
 
           def associated_class
-            reflection.klass
+            if polymorphic?
+              subject
+            else
+              reflection.klass
+            end
```
matsales28, line 20: I'll wait to see what happens when running all the specs, but when the relation is polymorphic,  the `reflection.klass` method raises an error. I found this while trying to create a spec to replicate a case where you have a polymorphic association or delegated_type one, but you also pass the `scopes` qualifier to that association. When trying to make a spec for that cause, I noticed this problem.

 My solution was to return a valid ActiveRecord object(which in that case, I returned the subject itself) because this will be used only to grab the value of the `where_values_hash` in the `ModelReflector#extract_relation_clause_from`, which I think doesn't matter which model you're using to. But we can think on another solution here.

VSPPedro (approved):
LGTM!

PR #1607, opened 2024-01-26
Support association strict_loading option

rhannequin:
`strict_loading` was added to [Rails 6.1] to prevent lazy loading of associations.

As adding it to an association declaration can have a massive impact on the way the record and its association is treated, it can be useful to ensure in a test suite the presence of this option.

This adds support for adding the `strict_loading` option to an association declaration.

[Rails 6.1]: https://github.com/rails/rails/pull/37400

Co-authored-by: Jose Blanco @laicuRoot <jose.blanco@thoughtbot.com>

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -235,6 +235,18 @@ module Shoulda
       #       should have_many(:people).strict_loading(true)
       #     end
       #
+      # Default value is true when no argument is specified
+      #
+      #     # RSpec
+      #     RSpec.describe Organization, type: :model do
+      #       it { should have_many(:people).strict_loading }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class OrganizationTest < ActiveSupport::TestCase
+      #       should have_many(:people).strict_loading
+      #     end
+      #
       # ##### autosave
       #
       # Use `autosave` to assert that the `:autosave` option was specified.
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -217,6 +217,24 @@ module Shoulda
       #       should belong_to(:organization).touch(true)
       #     end
       #
+      # ##### strict_loading
+      #
+      # Use `strict_loading` to assert that the `:strict_loading` option was specified.
+      #
+      #     class Organization < ActiveRecord::Base
+      #       has_many :people, strict_loading: true
+      #     end
+      #
+      #     # RSpec
+      #     RSpec.describe Organization, type: :model do
+      #       it { should have_many(:people).strict_loading(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class OrganizationTest < ActiveSupport::TestCase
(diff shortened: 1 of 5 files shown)


matsales28:
That's a great contribution! Thanks for working on this.

The `strict_loading` option can also be defined at the model and application levels.
```
# In the model level
class Example < ApplicationRecord
  self.strict_loading_by_default = true
end

# Globally
config.active_record.strict_loading_by_default = true
```

We should consider that when making the check in the `AssociationMatcher`. We should also create test cases for the possible combinations of those assignment levels. Enabling it in the application, and disabling it in the model, or the association, for example. There's an example of setting up the configuration on the specs using the `ApplicationConfigurationHelpers#with_belongs_to_required_by_default` method.

lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -217,6 +217,24 @@ module ActiveRecord
...
+      #     # RSpec
+      #     RSpec.describe Organization, type: :model do
+      #       it { should have_many(:people).strict_loading(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class OrganizationTest < ActiveSupport::TestCase
+      #       should have_many(:people).strict_loading(true)
+      #     end
+      #
```
matsales28, line 237: We could add to the documentation that we can use `strict_loading` without specifying the argument and it'll default to `true`.

rhannequin:
Thank you for your review, @matsales28 

It got into our mind to deal with more possible situations with this option, as you mentioned.
We encountered an issue with `strict_loading: false` explained in this issue: #1608 
With that in mind, we started with one simple use-case. I agree that this would be better to support every single combination with strict loading.

We might be missing something, so please forgive us if the issue is written as "bug report" if it's not a bug.

I'm happy to discuss the situation, and make the necessary changes to this PR based on the outcome of our discussion.

rhannequin:
We have started working on covering all possible combinations but it's going to take a while, so I'm moving the PR to draft again.

laicuRoot:
I'm reopening the PR as we have a working solution for all the cases. 

We have added the logic in the `associated_class` method of the `ModelReflection` class, as this was the most straightforward working approach. @matsales28, please let us know if you feel this logic should live elsewhere. We are open to suggestions.

spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
```diff
@@ -1199,6 +1199,409 @@ def belonging_to_non_existent_class(model_name, assoc_name, options = {})
       expect(Parent.new).to have_many(:children)
     end
 
+    context 'when the application is configured with strict_loading disabled by default' do
```
matsales28, line 1202: What do you think about nesting all those specs under the same context? When trying this branch, I wrapped everything in a `context 'strict_loading'` to run the specs in isolation, it also helps to identify between all the other features presents in the `association_matcher_spec`

matsales28 (approved):
I left two comments after adjusting those comments, it's ready for merge to me!

rhannequin:
@matsales28 Thanks a lot for helping us with a suggested change. We just applied it and also nested all specs inside a single `describe` block.

VSPPedro (approved):

PR #1611, opened 2024-02-02
chore: Update dependencies

matsales28:
In this PR we are updating the base dependencies and the Rails 7.1.x version to Rails 7.1.3.


VSPPedro (approved):
LGTM!

PR #1613, opened 2024-02-16
Lazy load `ActionController` and `Routing` matchers for `ActionController::TestCase`

ilianah:
This PR sets up two `on_load` hooks to include the `ActionController` and `Routing` matchers modules into `ActionController::TestCase`. They run only when `ActionController::TestCase` is loaded rather than immediately whenever the `integrate_with` method is called.

This change should improve app booting performance by delaying the inclusion of the matchers until they are actually needed.


matsales28 (approved):
LGTM!

VSPPedro (approved):

PR #1615, opened 2024-02-20
Allow beginless and endless ranges in validates_inclusion_of

### LINKED ISSUE #1614: RangeError: cannot get the first element of beginless range
### Description

Currently, beginless and endless ranges are not correctly handled by `.in_range` on top of `validates_inclusion_of`, even though ActiveRecord properly handles beginless and endless ranges.

### Reproduction Steps

Given a sample model: 
```ruby
# app/models/past_post.rb
class PastPost < ApplicationRecord
  validates :created_on, inclusion: { in: -> { (..Date.yesterday) } }
end
```

And the respective spec:
```ruby
# spec/models/past_post_spec.rb

RSpec.describe PastPost do
  # ...

  describe 'Validations' do
    it { is_expected.to validate_inclusion_of(:created_on).in_range(..Date.yesterday) }
  end
end
```

An error is raised instead of correctly asserting the validation:
```
     Failure/Error: it { is_expected.to validate_inclusion_of(:created_on).in_range(..Date.yesterday) }

     RangeError:
       cannot get the first element of beginless range
```

A similar issue is raised for endless ranges:
```
     RangeError:
       cannot get the maximum of endless range
```

### System configuration
**shoulda_matchers version**: 6.1.0
**rails version**: 7.1.3
**ruby version**: 3.2.2

pjpires:
### Problem
Currently, the min and max range values are checked with `range.first` and `range.max`, that will fail with `cannot get the first element of beginless range (RangeError)` and `cannot get the maximum of endless range (RangeError)` respectively for beginless/endless ranges.

### Proposed solution
We can rely on `range.begin` and `range.end` (by correctly checking if `end` should be considered by looking at `exclude_end?`) instead, that will both return `nil` instead of raising for beginless/endless ranges.

Also added the required early returns on minimum/maximum value checks when the begin/end of range is absent and updated the specs with two scenarios for beginless/endless ranges.

Fixes #1614.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -309,8 +309,8 @@ EOT
 
         def in_range(range)
           @range = range
-          @minimum = range.first
-          @maximum = range.max
+          @minimum = minimum_range_value
+          @maximum = maximum_range_value
           self
         end
 
@@ -400,6 +400,18 @@ EOT
 
         private
 
+        def minimum_range_value
+          @range.begin
+        end
+
+        def maximum_range_value
+          if @range.exclude_end?
+            @range.end ? (@range.end - 1) : nil
+          else
+            @range.end
+          end
+        end
+
         def matches_for_range?
           disallows_lower_value &&
             allows_minimum_value &&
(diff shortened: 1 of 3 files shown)


lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
```diff
@@ -441,26 +453,38 @@ def disallows_lower_value
         end
 
         def allows_minimum_value
+          return true unless @minimum
+
```
matsales28, line 457: I'd say we could change this by removing the guard clause to stay consistent with the other methods in this class.

```suggestion
          @minimum.nil? || allows_value_of(@minimum, @low_message)
```

matsales28 (approved):
LGTM! I left a single comment about changing the way we handle the guard clause otherwise, everything looks good to me. Thanks for fixing this bug!

VSPPedro (approved):
LGTM! Thank you for working on this!

lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
pjpires: That makes a lot of sense, thanks! I've updated the nil checks. 👍

PR #1620, opened 2024-03-15
chore: Update dependencies of rails-7.0 gemfile

matsales28:
On this PR we're updating the dependencies for the rails-7_0 Gemfile. I'll do the same thing for the other Rails versions Gemfiles and the base Gemfile as well.


VSPPedro (approved):

gemfiles/rails_7_0.gemfile
```diff
@@ -20,7 +20,7 @@ gem "yard"
 gem "spring"
 gem "spring-watcher-listen", "~> 2.0.0"
 gem "rails-controller-testing", ">= 1.0.1"
-gem "rails", "7.0.8"
+gem "rails", "7.0.8.1"
```
VSPPedro, line 23: I think it's a good idea to update the Appraisal file too. That way, when we regenerate the gemfiles, they won't have old versions of Rails.

matsales28: That's true. I'll update it on the other PR's as well

VSPPedro (approved):

PR #1623, opened 2024-03-22
fix: with_high_message expects a custom validation message too

freesteph:
The only impact being in the generated documentation (via description_clauses_for_qualifiers):

before:

> is expected to validate that Y lies inside the range R

after

> is expected to validate that X lies inside the range R producing a
custom validation error on failure

As with_low_message already does.

### DIFF
--- a/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
+++ b/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb
@@ -344,6 +344,7 @@ EOT
 
         def with_high_message(message)
           if message
+            @expects_custom_validation_message = true
             @high_message = message
           end
 


matsales28 (commented):
Do you mind adding a spec to cover this case?

I think you can do something similar to the one on `spec/unit/shoulda/matchers/active_model/validate_numericaly_of_spec.rb`

https://github.com/thoughtbot/shoulda-matchers/blob/33bec1c26e62437aa7f623c8fcdd49eb34e71076/spec/unit/shoulda/matchers/active_model/validate_numericality_of_matcher_spec.rb#L2111-L2116

matsales28:
I'm closing this one in favour of https://github.com/thoughtbot/shoulda-matchers/pull/1718.

Thanks for contributing and for raising this error.

PR #1624, opened 2024-03-22
fix: Polymorphic + STI uniqueness check

### LINKED ISSUE #1622: polymorph uniq check fails with 6.2
### Description

After the update to 6.2 one of my uniqueness-specs fails

### Reproduction Steps

Spec:

    is_expected.to validate_uniqueness_of(:select_option).scoped_to([:checklistable_type, :checklistable_id])

class:

    #  checklistable_type :string           not null
    #  checklistable_id   :bigint           not null
    #  select_option_id   :bigint           not null

    belongs_to :select_option
    belongs_to :checklistable,
      polymorphic: true,
      inverse_of: :checked_elements

    validates :select_option, uniqueness: {scope: [:checklistable_type, :checklistable_id]}

### Expected behavior

The spec should return true, did this until 6.2

### Actual behavior

fails with

> Expected SelectOption::CheckedElement to validate that :select_option is
> case-sensitively unique within the scope of :checklistable_type und
> :checklistable_id, but this could not be proved.
> After taking the given SelectOption::CheckedElement, whose
> :select_option is ‹#<SelectOption id: 2, account_id: 3, name:
> "option_5", description: "MyText", data: {}, type:
> "SelectOption::InterruptionReason", start_date: "2022-01-01",
> end_date: nil, visible_service_types: [], created_at: "2024-03-18
09:07:42.564821000 +0100", updated_at: "2024-03-18 09:07:42.564821000
+0100", short_form: "kurz_5", invoiceable: true, sort_number: nil,
classic_id: nil, group_type: nil, client_access: false>›, and saving
it as the existing record, then making a new
SelectOption::CheckedElement and setting its :select_option to
‹#<SelectOption id: 2, account_id: 3, name: "option_5", description:
"MyText", data: {}, type: "SelectOption::InterruptionReason",
start_date: "2022-01-01", end_date: nil, visible_service_types: [],
created_at: "2024-03-18 09:07:42.564821000 +0100", updated_at:
"2024-03-18 09:07:42.564821000 +0100", short_form: "kurz_5",
invoiceable: true, sort_number: nil, classic_id: nil, group_type: nil,
client_access: false>› as well, its :checklistable_type to a different
value, ‹"Client"› und its :checklistable_id to a different value, ‹2›,
the matcher expected the new SelectOption::CheckedElement to be valid,
but it was invalid instead, producing these validation errors:
     
> * select_option: ["ist bereits vergeben"]

### System configuration
**shoulda_matchers version**: 6.2.0
**rails version**: 7.1.3.2
**ruby version**: 3.2.3

matsales28:
Instead of just initializing a new class that inherits from the parent class, we are initializing a new class but also overriding some method definitions. That is necessary because the Rails guides recommend that whenever you have a polymorphic association and STI you should override the `..._type` column to use the type name of the base STI class.

https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#module-ActiveRecord::Associations::ClassMethods-label-Polymorphic+Associations

This should fix #1622

### DIFF
--- a/lib/shoulda/matchers/active_record/uniqueness/model.rb
+++ b/lib/shoulda/matchers/active_record/uniqueness/model.rb
@@ -29,7 +29,19 @@ module Shoulda
           end
 
           def symlink_to(parent)
-            namespace.set(name, Class.new(parent))
+            table_name = parent.table_name
+
+            new_class = Class.new(parent) do
+              define_singleton_method :table_name do
+                table_name
+              end
+
+              define_singleton_method :base_class do
+                self
+              end
+            end
+
+            namespace.set(name, new_class)
           end
 
           def to_s
--- a/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/validate_uniqueness_of_matcher_spec.rb
@@ -1274,6 +1274,36 @@ long as it is not blank, but this could not be proved.
         scoped_to(:favoriteable_type)
     end
 
+    context 'when polymorphic model is used on a STI model' do
+      it 'still works' do
+        columns = {
+          attachable_id: { type: :integer, options: { null: false } },
+          attachable_type: { type: :string, options: { null: false } },
+          title: { type: :string, options: { null: false } },
+        }
+
+        asset_model = define_model 'Asset', columns do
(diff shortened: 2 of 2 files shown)


VSPPedro (approved):
LGTM! 🚀

PR #1627, opened 2024-04-12
feat: Add `default` qualifier to `define_enum_for` matcher

matsales28:
On this commit we're adding a new qualifier to the `define_enum_for` matcher called `with_default`. This qualifier is used to test that the enum is defined with a default value. A proc can also be passed, and will be called once each time a new value is needed.

It's nice to note that using Time or Date as the return of Procs as default value can lead to flaky tests, so it's recommended to freeze time or date to avoid this.


VSPPedro (approved):
Nice work!

PR #1637, opened 2024-06-13
Support deprecated status code symbols.

ioquatix:


simi:
Just to add a little background, some code names were changed and related constant should not be reached directly to get proper deprecation warning.

matsales28:
> Just to add a little background, some code names were changed and related constant should not be reached directly to get proper deprecation warning.

It makes sense to me; I triggered the CI, and once it pass, I'm good merging. Thanks for adding that!

matsales28 (approved):

ioquatix:
Thanks team!

PR #1657, opened 2024-12-05
Add Ruby 3.4 support

voxik:
* Use a single quote instead of a backtick as an opening quote. [[Feature #16495](https://bugs.ruby-lang.org/]issues/16495)]
* Hash#inspect rendering have been changed. [[Bug #20433](https://bugs.ruby-lang.org/issues/20433)]

One thing to consider is if the single quote / backtick support should not be guarded by some Ruby version check. That would make it more explicit and later allowed easier removal, when Ruby <= 3.3 support is going to be removed.

The same applies to the Hashes, although this is likely good change on its own, since the Hash string representation is considered internal detail AFAIK.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
+          - head
           - 3.3.0
           - 3.2.2
           - 3.1.4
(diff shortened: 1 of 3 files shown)


voxik:
Hm, there seems to be errors such as:

~~~
LoadError:
  cannot load such file -- mutex_m
~~~

I thought it was [addressed for Rails 7.2](https://github.com/rails/rails/pull/49674). Anyway, updating the Rails 7.2 Gemfile.lock could help IMHO, but that should be separate ticket. I leave that to somebody else.

voxik:
And Rubocop 🙈 The change was supposed to be minimalist ...

PR #1667, opened 2025-05-08
feat: Support encrypted ActionText attributes

spdawson:
## Synopsis

When rich text attributes are declared by passing `encrypted: true` to `has_rich_text`, the attribute class will be
`ActionText::EncryptedRichText` rather than the `ActionText::RichText` base class.

Extend the existing `have_rich_text` matcher such that it will continue to match, even when rich text attributes are specified with the `encrypted: true` option.

Extends work done in #1263.


matsales28 (commented):
There's a failure in the spec. The `encrpted_rich_text_is_defined?` method is not available inside of the define_model block. @spdawson

Issue #1684, opened 2025-10-31
shoulda-matchers 7.0.0 has undeclared and unused mutex_m dependency


tindron:
### Description

After upgrading to shoulda-matchers 7.0.0 trying to run specs on ruby 3.4 results in an error requiring `mutex_m` because it is not a declared dependency in the gemspec.  There doesn't appear to be any reference to the `Mutex_m` module that `mutex_m` defines so removing the require should fix the issue.

```
➜  logged-data git:(rubocop-1.81) ✗ rspec spec/lib/field_finder_spec.rb      
/Users/randy.stoller/.rvm/gems/ruby-3.4.5/gems/shoulda-matchers-7.0.0/lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb:1: warning: mutex_m was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add mutex_m to your Gemfile or gemspec to silence this warning.

An error occurred while loading ./spec/lib/field_finder_spec.rb.
Failure/Error: require File.expand_path('../config/environment', __dir__)

Bundler::GemRequireError:
  There was an error while trying to load the gem 'shoulda-matchers'.
  Gem Load Error is: cannot load such file -- mutex_m
  Backtrace for gem load error is:
  /Users/randy.stoller/.rvm/rubies/ruby-3.4.5/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
  /Users/randy.stoller/.rvm/rubies/ruby-3.4.5/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
  /Users/randy.stoller/.rvm/gems/ruby-3.4.5/gems/zeitwerk-2.7.3/lib/zeitwerk/core_ext/kernel.rb:34:in 'Kernel#require'
  /Users/randy.stoller/.rvm/gems/ruby-3.4.5/gems/shoulda-matchers-7.0.0/lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb:1:in '<top (required)>'
  /Users/randy.stoller/.rvm/rubies/ruby-3.4.5/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
  /Users/randy.stoller/.rvm/rubies/ruby-3.4.5/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
[...]

Issue #1685, opened 2025-10-31
Add mutex_m dependency


garrettgregor:
### Description

When running test suite, either in CI or Locally, shoulda-matchers fails because mutex_m is no longer included in the standard library for Ruby

### Reproduction Steps

You can simply run your test suite - anything that relies on shoulda matchers will reproduce this.

### Expected behavior

Tests should be able to run.

### Actual behavior

```sh
.rbenv/versions/3.4.2/lib/ruby/gems/3.4.0/gems/shoulda-matchers-7.0.0/lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb:1: warning: mutex_m was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add mutex_m to your Gemfile or gemspec to silence this warning.

An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require "shoulda/matchers"

LoadError:
  cannot load such file -- mutex_m
# ./spec/spec_helper.rb:1:in '<top (required)>'
No examples found.


Finished in 0.00003 seconds (files took 0.14107 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

or in CI:

```log
/home/runner/work/<PROJECT>/<PROJECT>/vendor/bundle/ruby/3.4.0/gems/shoulda-matchers-7.0.0/lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb:1: warning: mutex_m was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add mutex_m to your Gemfile or gemspec to silence this warning.
Also please contact the author of shoulda-matchers-7.0.0 to request adding mutex_m into its gemspec.
bin/rails aborted!
Bundler::GemRequireError: There was an error while trying to load the gem 'shoulda-matchers'. (Bundler::GemRequireError)
Gem Load Error is: cannot load such file -- mutex_m
Backtrace for gem load error is:
/opt/hostedtoolcache/Ruby/3.4.2/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
/opt/hostedtoolcache/Ruby/3.4.2/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
[...]

PR #1687, opened 2025-10-31
hotfix: Add mutex_m gem dependency

matsales28:
Closes: #1684 #1685


lib/shoulda/matchers/active_record/uniqueness/test_model_creator.rb
garrettgregor: Does this not need mutex anymore?

shoulda-matchers.gemspec
```diff
@@ -38,4 +38,5 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 3.2'
   s.add_dependency 'activesupport', '>= 7.1'
+  s.add_development_dependency 'mutex_m'
```
n-rodriguez: This is useless since it's already defined in Gemfile

PR #1690, opened 2025-11-18
Introduce support for deprecated associations API

### LINKED ISSUE #1681: Introduce support for deprecated associations API
### Background

  Rails 8.1 introduced the ability to mark Active Record
  associations as deprecated. This allows
  developers to mark associations for deprecation while maintaining
   backward compatibility.

  ### Current Rails Behavior

  Associations can now be marked as deprecated:

  ```ruby
    has_many :posts, deprecated: true
    has_one :profile, deprecated: true
    belongs_to :publisher, deprecated: true
  end
```

  When a deprecated association is used, Rails will report the
  usage according to the configured reporting mode. The reporting
  behavior (mode and backtrace options) is configured globally or
  per-model, not on the association declaration itself.

  Reporting modes (configured separately):
  - :warn - Issues a deprecation warning (default)
  - :raise - Raises an exception
  - :notify - Sends a notification via ActiveSupport::Notifications

  Proposed Feature

  Add matcher support to test that associations are properly marked
   as deprecated:

  # Assert association is deprecated
  it { should have_many(:posts).deprecated }
  it { should have_one(:profile).deprecated }
  it { should belong_to(:author).deprecated }

  # Assert association is NOT deprecated
  it { should have_many(:posts).deprecated(false) }
  # or
  it { should_not have_many(:posts).deprecated }

  Implementation Considerations

  1. The feature should work with all association types:
    - has_many
    - has_one
    - belongs_to
    - has_and_belongs_to_many
    - has_many :through
  2. The matcher should only verify the deprecated: true attribute:
    - Should check if the association has deprecated: true set
    - Should NOT test reporting mode or backtrace settings (those
  are configuration concerns)
  3. Rails version compatibility:
    - This feature requires Rails 8.1+
    - Should gracefully skip or provide helpful error on older
  Rails versions

  Example Test Cases
```ruby
  RSpec.describe 'deprecated association matcher' do
    context 'with deprecated association' do
      it 'matches when association is deprecated' do
        define_model :Author do
          has_many :posts, deprecated: true
        end

        expect(Author.new).to have_many(:posts).deprecated
      end

      it 'works with different association types' do
        define_model :Author do
          has_one :profile, deprecated: true
          belongs_to :publisher, deprecated: true
        end

        expect(Author.new).to have_one(:profile).deprecated
        expect(Author.new).to belong_to(:publisher).deprecated
      end
    end

    context 'without deprecated association' do
      it 'does not match when association is not deprecated' do
        define_model :Author do
          has_many :posts
        end

        expect(Author.new).not_to have_many(:posts).deprecated
      end

      it 'can explicitly assert not deprecated' do
        define_model :Author do
          has_many :posts
        end

        expect(Author.new).to have_many(:posts).deprecated(false)
      end
    end
  end
```

  Usage Example

```ruby
  # In your model
    has_many :legacy_posts, deprecated: true
    has_many :posts
  end

  # In your spec
  RSpec.describe Author do
    it { should have_many(:legacy_posts).deprecated }
    it { should have_many(:posts).deprecated(false) }
    # or
    it { should_not have_many(:posts).deprecated }
  end
```

stefannibrasil:
Closes #1681 

Rails 8.1 introduced the ability to mark [Active Record associations as deprecated](https://github.com/rails/rails/pull/55285). This allows developers to mark associations for deprecation while maintaining backward compatibility.

## Example test cases

Assert association is deprecated:

```ruby
it { should have_many(:posts).deprecated }
it { should have_one(:profile).deprecated }
it { should belong_to(:author).deprecated }
```

Assert association is NOT deprecated:

```ruby
it { should have_many(:posts).deprecated(false) }
it { should_not have_many(:posts).deprecated }
```

## Notes and Questions

- Added docs and coverage for all associations, let me know if I missed anything!
- What type of exception should I raise when the Active Record version is < 8.1? I didn't find an existing case to follow. `NotImplementedError` makes sense to me, although I debated on returning `ArgumentError`, since that's what ActiveRecord returns when trying to use it on versions < 8.1. Open to suggestions to provide more helpful error on older
AR versions ✅ 
- I copied the rails_version check in the tests. Do you usually just check for rails versions or should I create a helper for checking active record instead? I am checking for ActiveRecord's version in the association class, since `shoulda-matchers` can be used in a non-Rails app ✅ 

## Tests with a real app

I tested the code changes with these scripts: https://gist.github.com/stefannibrasil/02809758148838bf3950d34a64c37106 (Rails 8.0 and 8.1.1). Both provide lots of examples of how to use the new matcher.

### DIFF
--- a/lib/shoulda/matchers/active_record/association_matcher.rb
+++ b/lib/shoulda/matchers/active_record/association_matcher.rb
@@ -702,7 +702,7 @@ module Shoulda
       #
       # ##### deprecated
       #
-      # Use `deprecated` to assert that the association is not allowed to be nil.
+      # Use `deprecated` to assert that the `:deprecated` option was specified.
       # (Enabled by default in Rails 8.1+).
       #
       #     class Vehicle < ActiveRecord::Base
--- a/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
+++ b/spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
@@ -532,7 +532,7 @@ describe Shoulda::Matchers::ActiveRecord::AssociationMatcher, type: :model do
       end
 
       context 'when qualified with required(false)' do
-        it 'passes' do
+        it 'fails with an appropriate message' do
           assertion = lambda do
             expect(belonging_to_parent(required: true)).
               to belong_to(:parent).required(false)
@@ -762,7 +762,7 @@ describe Shoulda::Matchers::ActiveRecord::AssociationMatcher, type: :model do
     end
 
     if active_record_gte_8_1?
-      context 'an association with a :touch option' do
+      context 'an association with a :deprecated option' do
         [false, true].each do |deprecated_value|
           context "when the model has deprecated: #{deprecated_value}" do
             it 'accepts a matching deprecated option' do
@@ -804,6 +804,18 @@ describe Shoulda::Matchers::ActiveRecord::AssociationMatcher, type: :model do
         it 'rejects deprecated()' do
           expect(belonging_to_parent).not_to belong_to(:parent).deprecated
         end
+      end
+
+      it 'rejects an association with a non-matching :deprecated option with the correct message' do
+        define_model :parent, adopter: :boolean
+        define_model :child, parent_id: :integer do
+          belongs_to :parent, deprecated: false
+        end
+
+        message = 'Expected Child to have a belongs_to association called parent (parent should have deprecated set to true)'
+        expect do
+          expect(Child.new).to belong_to(:parent).deprecated(true)
+        end.to fail_with_message(message)
       end
     end
 
@@ -3138,14 +3150,6 @@ Expected Person to have a has_and_belongs_to_many association called relatives (
             end
           end
         end
-
-        it 'rejects an association with a non-matching :deprecated option with the correct message' do
-          message = 'Expected Vehicle to have a belongs_to association called drivable (drivable should have deprecated set to true)'
-
-          expect do
-            expect(delegating_type_to_drivable(deprecated: false)).to have_delegated_type(:drivable).deprecated(true)
-          end.to fail_with_message(message)
-        end
       end
 
       context 'an association without a :deprecated option' do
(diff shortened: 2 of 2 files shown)


lib/shoulda/matchers/active_record/association_matcher.rb
```diff
@@ -1586,7 +1692,8 @@ def matches?(subject)
             touch_correct? &&
             types_correct? &&
             strict_loading_correct? &&
-            submatchers_match?
+            submatchers_match? &&
+            deprecated_correct?
```
matsales28: Let's leave the `submatchers_match?` as the last check, as they're usually a slower check.

spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
matsales28: We need tests that assert that the correct message output, e.g.

```ruby
 it 'rejects an association with a non-matching :autosave option with the correct message' do
      message = 'Expected Vehicle to have a belongs_to association called drivable (drivable should have autosave set to true)'

      expect {
        expect(delegating_type_to_drivable(autosave: false)).to have_delegated_type(:drivable).autosave(true)
      }.to fail_with_message(message)
    end
````

File: lib/shoulda/matchers/active_record/association_matcher.rb [hunk 2]
```diff
@@ -374,6 +374,25 @@ module ActiveRecord
...
+      #     # RSpec
+      #     RSpec.describe Account, type: :model do
+      #       it { should belong_to(:bank).deprecated(true) }
+      #     end
+      #
+      #     # Minitest (Shoulda)
+      #     class AccountTest < ActiveSupport::TestCase
+      #       should belong_to(:bank).deprecated(true)
+      #     end
+      #
```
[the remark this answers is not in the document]
stefannibrasil: Good catch, thanks! Done in [2239770](https://github.com/thoughtbot/shoulda-matchers/pull/1690/commits/223977003d4ab2ef6065cc79e983f6cd1c3abfaa)

spec/unit/shoulda/matchers/active_record/association_matcher_spec.rb
stefannibrasil: Another great catch. I missed this test 🙈 Added these in [4ea3d9c](https://github.com/thoughtbot/shoulda-matchers/pull/1690/commits/4ea3d9c50cbe652a549ee0524c6e1bf203586ceb)

matsales28 (approved):
LGTM!

